PR #18, opened 2017-02-06
allow detecting eigen on easybuild; fixes #17

### LINKED ISSUE #17: Library dependencies detection fails on EasyBuild

bast:
@stigrj @ilfreddy With this change I was able to configure under EasyBuild:
```
ml notur
source mod_setup.sh
module load CMake
module load Eigen
export EIGEN3_ROOT=/global/hds/software/cpu/eb2/Eigen/3.2.9-foss-2016b
module load Boost
./setup --cc=icc --cxx=icpc
```

### DIFF
--- a/cmake/custom/FindEigen3.cmake
+++ b/cmake/custom/FindEigen3.cmake
@@ -61,7 +61,7 @@ if (EIGEN3_INCLUDE_DIR)
 else (EIGEN3_INCLUDE_DIR)
   find_path(EIGEN3_INCLUDE_DIR NAMES signature_of_eigen3_matrix_library
       PATHS ${EIGEN3_ROOT} ENV EIGEN3_ROOT
-      PATH_SUFFIXES include/eigen3 include/eigen
+      PATH_SUFFIXES include/eigen3 include/eigen include
       NO_DEFAULT_PATHS
     )
 


gitpeterwind:
The line 
export EIGEN3_ROOT=/global/hds/software/cpu/eb2/Eigen/3.2.9-foss-2016b
Is certainly not something, for instance a master student, would guess!
I thought the main point of module/setup/cmake/makefile was to avoid such things.
Isn't some layer not doing its work here?

bast:
Yup. [This file](https://github.com/MRChemSoft/mrchem/blob/master/cmake/custom/FindEigen3.cmake#L63) uses `$EIGEN3_ROOT` to find the library and header files. It is of course possible to make the discovery independent of `$EIGEN3_ROOT` but then you take away control from the user to select which one you want if there are several in the environment. You could also ask why is `$EIGEN3_ROOT` not set. This is a question outside MRChem.

gitpeterwind:
I have sent the question to uit-support :)
and the answwer was

"Vad jag kan se ligger inte
problemet på våran sida. MrChem borde uppdatera sin findeigen rutin."

That is the main question: which layer is not doing its job as it should?

bast:
I really don't think this is up to the UiT HPC support to decide. It is up to MRChem to decide which environment variables you expect from the system. The question is, do you want to rely on `$EIGEN3_ROOT` being set? If yes, then you can approach UiT HPC and try to convince them to make that happen.

gitpeterwind:
I would prefer not to rely on anything being set.

But I do not know what are "normal" expectations about what should be 
available.

I personnaly can easily live with an additional "export ...", but I 
understand that MRChem should aim at following some standards.

What do you think should be the best way to solve this problem?

bast:
I think we need to open a new issue thread for this: https://github.com/MRChemSoft/mrchem/issues/19. This is a closed pull request.

PR #35, opened 2017-07-21
Exchange and Shared memory

gitpeterwind:
There are several issues we should discuss. Before accepting the pull request, you should at least check these points, as they concern the non-MPI section too:

1) I changed the "screening" precision in internal Exchange: it was  prec = getScaledPrecision(i, j) modified to  prec = std::max(getScaledPrecision(i, j), getScaledPrecision(j, i)).
The former did use the "i" precision also for the "j" orbital. Meaning that a different result can be obtained if the order of the orbitals is modified.
There are other ways of defining this too.

2)  In the testPreComputed I modified   (ex_i != 0) to (ex_i->getNNodes() != 0) as the former is never true (ex_i is defined, even if empty) . This condition was not used anyway I think.

3) I didn't modify this, by it seems to me that the adjoint of Vij must be used for calculating ex_j. As long as we have only real orbitals it does not matter.

The parallel exchange does give the same result within the precision, but not behind. This is possibly due to non-symmetry in parts of the code, i.e. for example that doing <i|O j> or <j|O i> can be different even if the operator formally is symmetrical. We should discuss this.

### DIFF
--- a/src/mrchem/MREnv.cpp
+++ b/src/mrchem/MREnv.cpp
@@ -10,7 +10,6 @@ MultiResolutionAnalysis<3> *MRA;
 
 void MREnv::initializeMRCPP(int argc, char **argv) {
 #ifdef HAVE_MPI
-    MPI_Init(NULL, NULL);
     MPI_Initializations();
 #endif
     int nThreads = omp_get_max_threads();
@@ -59,11 +58,11 @@ void MREnv::initializeMRCPP(int argc, char **argv) {
     println(0, "BLAS was NOT found, Eigen will be used instead!" << endl);
 #endif
 
-    if (MPI_size > 1 or nThreads > 1) {
+    if (MPI_Orb_size > 1 or nThreads > 1) {
         println(0,"+++ Parallel execution: ");
-        println(0,"  MPI hosts available     : " << MPI_size);
+        println(0,"  MPI hosts available     : " << MPI_Orb_size);
         println(0,"  Threads/host            : " << nThreads);
-        println(0,"  Total used CPUs         : " << MPI_size*nThreads);
+        println(0,"  Total used CPUs         : " << MPI_Orb_size*nThreads);
         println(0,"");
     } else {
         println(0,"+++ Serial execution" << endl);
--- a/src/mrchem/MREnv.cpp
+++ b/src/mrchem/MREnv.cpp
@@ -58,11 +58,11 @@ void MREnv::initializeMRCPP(int argc, char **argv) {
     println(0, "BLAS was NOT found, Eigen will be used instead!" << endl);
 #endif
 
-    if (MPI_Orb_size > 1 or nThreads > 1) {
+    if (mpiOrbSize > 1 or nThreads > 1) {
         println(0,"+++ Parallel execution: ");
-        println(0,"  MPI hosts available     : " << MPI_Orb_size);
+        println(0,"  MPI hosts available     : " << mpiOrbSize);
         println(0,"  Threads/host            : " << nThreads);
-        println(0,"  Total used CPUs         : " << MPI_Orb_size*nThreads);
+        println(0,"  Total used CPUs         : " << mpiOrbSize*nThreads);
         println(0,"");
     } else {
         println(0,"+++ Serial execution" << endl);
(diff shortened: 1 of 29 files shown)


stigrj:
1. Shouldn't it be std::min instead of std::max? If the two precs differ, we should use the tighter one, right?

2. You are right, this was written at a time when ex_i actually could be NULL. Now ex_i is always defined, but its real and imaginary parts can be NULL (if the operator is not set up before use). In fact, I would now use a reference instead of pointer:
```
Orbital &ex_i = this->exchange.getOrbital(i);
```

3. Yes, since we allow for complex-valued orbitals we should use the adjoint, even if we know that the ground state is real. The mult.adjoint() takes the adjoint of the first argument, so you can replace
```
mult(*phi_jij, fac_jij, phi_j, *V_ij);
mult(*phi_iij, fac_iij, phi_i, *V_ij);
```
with
```
mult(*phi_jij, fac_jij, *V_ij, phi_j);
mult.adjoint(*phi_iij, fac_iij, *V_ij, phi_i);
```
(change the order of multiplication in both for OCD reasons).

gitpeterwind:
1) I do not really have an overview how the screening is used. There are 
other forms that seems as natural for me; for instance something like:
min(tot_norms(i)/part_norms(j,i) , tot_norms(j)/part_norms(i,j))

What is the meaning of the "max(part_norms(i,j), part_norms(j,i))" in 
getScaledPrecision? (why not take only one of them ?)

(But, in doubt I can take the most conservative prec)


and 4)
How do we proceed practically with the git system? Do I do these changes 
in my next commit to my clone?

stigrj:
1. It is bit involved, as it includes both additions and multiplications/convolutions. The idea is to get the proper _relative_ precision in the result, which is computed as the sum of products. Then each of the products must be computed to the same _absolute_ precision, based on the norm of the result (tot_norms(i) on the code). This means that small contributions can be computed with less precision, and if the size of the contribution (part_norms(i,j) in the code) is smaller than the requested precision it can be omitted (getScaledPrecision() > 1.0 in the code). I guess this should be scaled by the number of contributions, which I don't think is currently done... Finally, given the _absolute_ requirement of the product we get the _relative_ requirements of each of the factors (orbital and convolution), but since the same convolution is used both for (i,j) and (j,i), we use the most conservative precision of the two. This is a poor mans (myself) attempt at linear scaling using localized orbitals, since more and more contributions will drop below the threshold. There are better (much more complicated) ways of doing this, as described by Harrison in their Hartree-Fock paper.

4. If you push a new commit to the same branch on your fork it will automatically be included in the pull request (as long as it has not been accepted already, of course). This is why it is a good idea to have your pull requests on separate branches: if you have everything on master you have to wait for the pull request to be accepted before you can continue working.

stigrj:
Some comments on coding style as well :)

In decreasing order of importance:

1. We use 4 space indentation (not 0, 1, 2 or 3)

2. The name of variables, objects and functions should start with a lower case letter (only classes start with upper case). I have made a few exceptions for single-letter names like:
```
MatrixXd M;
PoissonOperator P;
```

3. Functions should preferably look like this:
```
void myFunctionName();
```

4. Variables/object one of the two (not both combined):
```
double myVariableName;
FunctionTree<3> my_variable_name;
```
Lately, I tend to use the latter for variables/objects and the former only for functions, but this is not  consistent in the code.

5. Both Luca and I like spaces, and we think
 ```
for (int a = 0; a < 10; a++) {
    if (a == b) {
 ```
looks better than
 ```
 for(int a=0;a<10;a++){
     if(a==b){
 ```

gitpeterwind:
Shared memory capabilities under MPI are now included in the code and used for Coulomb Density.
Can still be developed.

New group definitions to handle the shared memory communication.

The style issues have been addressed (therefore a lot of file changed!).
Also the "ScaledPrecision" issue has been taken into account.

The code can now run on several nodes (tested for H2O with HF: same result until the second last digit for 1 and 5 MPI processes. 5MPI*6OMP is now 35% faster on two compute-nodes compared to 1MPI*20OMP on one compute-node)

stigrj:
Sounds great! I guess the MPI shared memory is just for reducing the memory footprint? The work load of computing the density is not distributed in any way, right?

robertodr:
As for code format conventions, I use [clang-format](https://clang.llvm.org/docs/ClangFormat.html)
It can:
- be integrated with both emacs and vim, so that open files can be formatted according to the specified rules.
- be run as a Git pre-commit hook, so that a `git commit` command will be stopped if any of the modified files does not comply with the formatting rules.

[This configurator](https://zed0.co.uk/clang-format-configurator/) can help you define the `.clang-format` file to include at the root of the repository.

PR #40, opened 2017-07-28
Compile issues

stigrj:
The code did not compile without OMP due to improper "unparallelization" of Eigen. Also replaced the redundant #define OPENMP with HAVE_OPENMP


gitpeterwind:
Looks fine to me! (but I am not allowed to merge)

PR #41, opened 2017-08-03
Compile issues

stigrj:
Small changes to make the code compile _without_ MPI, as well as run _with_ MPI


stigrj:
For the Orbital we have three levels of copy:
1. The copy constructor takes only the parameters of the class (occupancy/spin), but leaves the actual function undefined (real=imag=0)
```
Orbital newOrb(oldOrb);
```
2. The shallow copy takes only the real/imag pointers (occupancy/spin is unchanged) 
```
newOrb.shallowCopy(oldOrb);
```
3. The deep copy creates new real/imag FunctionTrees (occupancy/spin is unchanged)
```
newOrb.deepCopy(oldOrb);
```

PR #45, opened 2017-08-07
bug in KAIN parallelization corrected

gitpeterwind:
The updates where incorrectly sent to the other processes (to few elements sent in mpi_reduce)


stigrj:
Also, have a look at Accelerator::verifyOverlap(), which needs to be synchronized in some way. Here we test the overlap of each orbital with the corresponding one in the previous iteration. If the overlap is _not_ close to one it might indicate that two orbitals have swapped position in the OrbitalVector, and in this case we clear the KAIN history and start fresh.

gitpeterwind:
Accelerator::verifyOverlap has been updated now

PR #46, opened 2017-08-09
Cleanups

stigrj:
Cleanups in Helmholtz operator application: There was a growing number of getHelmholtzArgument() and related routines, now there's only one each for GroundStateSolver and LinearResponseSolver (not implemented yet).

New IdentityOperator: It's effect on an orbital is a deep copy, and its expectation value is the orbital overlap. Overlap routines moved from OrbitalVector. Usage:
```
IdentityOperator I;
I.setup(prec);
Orbital *new_orb = I(old_orb); // Deep copy
double S_ij = I(orb_i, orb_j); // Orbital overlap
MatrixXd S = I(orb_vec_1, orb_vec_2); // Overlap matrix
I.clear();
```
Reason: more consistent, all matrices computed as operator expectation values.


gitpeterwind:
You have tested that this give the same results also for a MPI run?

stigrj:
Yes, H20 energy differs only in last two digits between OMP=15 (compiled without MPI), MPI=5/OMP=3 and MPI=3/OMP=5.

PR #50, opened 2017-08-16
[WIP] HFCC's with regularized molecular potential

stigrj:
Description
---------------
Implement SCF using NEMO orbitals and a regularized molecular potential following F.A. Bischoff (2014). By removing the singularities in the nuclear potential we achieve faster point-wise convergence of the wave function at the nuclear positions, which is crucial for numerically accurate hyperfine coupling constants. The idea is to separate an analytical cusp from the numerical orbital using a nuclear correlation factor R: ``` |phi_i> = R|F_i>```.

TODO
--------
- [x] Implement nuclear correlation factor
- [x] Implement regularized potential
- [x] Adapt QMOperator expectation values
- [x] Adapt two-electron operators
- [x] Adapt SCF algorithm
- [x] Extend to several nuclei
- [x] Implement HFCC property
- [ ] Parallelize correlation factor (MPI shared memory)

Status
--------
- [ ] Ready to go


gitpeterwind:
I tried to merge with my latest master. 
There seems to be a compilation problem (seems not to be related to the merge directly):

```
/home/mifapw/mrchem/src/mrchem/SCFDriver.cpp(389): error: no instance of constructor "XCPotential::XCPotential" matches the argument list
            argument types are: (XCFunctional, OrbitalVector, ABGVOperator<3> *)
          XC = new XCPotential(*xcfun, *phi, ABGV_00);
```
XCPotential has the two last arguments with a default, but the third is not an ABGVOperator.

PR #54, opened 2017-10-10
Reduce scaling of kinetic matrix from N^2 to 2N

stigrj:
In principle it's still N^2 in the scalar product, but
in practice the time is spent in d/dx phi_i, which is
order N.


gitpeterwind:
This is not optimal in a MPI run: each MPI process will still need to compute the derivatives for all orbitals.
A better approach would be that each MPI computes its own derivatives, and then only the derivatives are sent. No need to send orbitals, since these are only used to make the derivatives.
This does not fit in the way the routine is called now, so a new routine would be required.
Since your solution works fine for serial runs, and will not break the parallel runs, we can still have it temporarily.

PR #58, opened 2017-11-21
Doxygen

ilfreddy:
Doxygen/Sphinx infrastructure for documenting the code is now up and running.
Most of it is (almost) verbatim copied form PCMSolver (thanks @robertodr!!!).
Now the "real" work can start.


cmake/custom/documentation.cmake
```diff
@@ -0,0 +1,34 @@
+# This script checks the availability of the tools
+# needed to build the HTML documentation pages:
+# - Doxygen
+# - Sphinx
+# - Perl (to count the lines of code)
+# - PyYAML module (to postprocess the lines of code counting)
+# - Matplotlib module (to generate the bar charts)
+# - Breathe module (to bridge Doxygen and Sphinx)
+
+option(ENABLE_DOCS "Enable building of documentation" ON)
```
robertodr: As @bast suggested in PCMSolver/pcmsolver#102, it might be a better idea to separate configuration and documentation building. `conf.py` can handle configuration of the `.in` files necessary for building the documentation, and `make doc` is currently just a simple alias for `sphinx-build`.

doc/conf.py
```diff
@@ -1,7 +1,9 @@
+#!/usr/bin/env python
```
robertodr: You might probably want to run `2to3` to get the script working with Python3. @arnfinn and @bast have fixed it in PCMSolver/pcmsolver#102 already

ilfreddy: @arnfinn did it this morning.

cmake/custom/documentation.cmake
ilfreddy: So you would not have anything about the documentation in the CMake structure?

bast: The advantage of decoupling is that the documentation can be built even if the CMake code is broken. For instance one might want to build documentation to figure out how to compile the code and for this it is nice if I don't have to compile the code to figure out how to compile the code. You might say: just look at public doc but one might want to build documentation for a code in the past.

ilfreddy: I see the point. I will try to do that and ask for help if need be :-)

ilfreddy: @robertodr: concerning the .in files, it seems to me the only one used in building the documentation is Doxyfile.in. Anything else?

ilfreddy:
@bast I have added a oneliner to build the doc without conf (makedoc.sh)
@robertodr Besides Doxygen.in are there other files that need need parsing to build the documentation of the code?

CMakeLists.txt
```diff
@@ -37,10 +37,11 @@ include(tests)
 include(python)
 include(git)
 include(initial_guess)
-include(doxygen)
+#include(doxygen)
```
robertodr: This line can be removed now.

CMakeLists.txt
```diff
@@ -37,10 +37,11 @@ include(tests)
...
 include(mw_filters)
 include(main)
 include(feature_summary)
 include(autocmake_save_flags)
+include(documentation)
```
robertodr: And this one can go too.

cmake/custom/FindSphinx.cmake
```diff
@@ -0,0 +1,38 @@
+# - This module looks for Sphinx
```
robertodr: This file can be removed from the repo

cmake/custom/documentation.cmake
```diff
@@ -0,0 +1,34 @@
+# This script checks the availability of the tools
```
robertodr: Same here, this file can now be removed.

cmake/custom/find_python_module.cmake
```diff
@@ -0,0 +1,24 @@
+# Find if a Python module is installed
```
robertodr: This file can be removed from the repo

makedoc.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+pip install -r doc/requirements.txt
```
robertodr: Can you add two lines in `README.md` with the commands to be run to build docs?

bast:
Alternative to running pip install would be to check whether dependencies are there and if not, give a nice error message pointing to how to install the dependencies. People might have Anaconda for instance and then they don't want to run pip install.

ilfreddy:
Please have a look. I have now cleaned up the CMake stuff and fixed the things tou were asking for. The only thing missing is the `2to3`

doc/conf.py
```diff
@@ -229,8 +253,8 @@
 # (source start file, target name, title,
 #  author, documentclass [howto, manual, or own class]).
 latex_documents = [
-  (master_doc, 'MRChem.tex', u'MRChem Documentation',
-   u'Luca Frediani, Stig Rune Jensen', 'manual'),
+  (master_doc, 'MRChem.tex', 'MRChem Documentation',
+   'Roberto Di Remigio, Luca Frediani, Krzysztof Mozgawa', 'manual'),
```
robertodr: Authors are the same as PCMSolver :sweat_smile:

doc/programmers/coding-standards.rst
```diff
@@ -0,0 +1,149 @@
+Coding standards
```
robertodr: I don't know if the contents of this file apply to MRChem too @stigrj

doc/programmers/general-structure.rst
```diff
@@ -0,0 +1,42 @@
+General Structure
```
robertodr: Same here, probably contents of this aren't relevant for MRChem @stigrj

doc/conf.py
ilfreddy: Ops...

ilfreddy:
@bast I have cleand up the documentation stuff. The only point which I am not so happy about is how to check for the required python packages. There is a rudimentary shell script, but @robertodr suggested to use a `try: catch:` directly in conf.py. What would you do?

stigrj:
Yes, the programmer's manual need some revision. That is, I comply with what's in there, but some of it isn't relevant and I think some things should be added. I will get to it next week. I suggest to keep the source (it's a good starting point) but to not publish it on readthedocs yet.

ilfreddy:
The programmer reference manual is now hidden. @stigrj can you merge this now? Then I (and everybody else :-)  ) can start documenting the code. I know you are all looking forward to it :-)

PR #59, opened 2017-12-08
CMake patch for mac to fix the librt problem

ilfreddy:
Changes to the CMake infrastructure that fix the librt compilation problem on a mac.
The libmrcpp library required librt, which does not exist on a mac. The librt at linking stage is now only present for `UNIX AND NOT MAC`.

I have also added a `FindLibrt.cmake` which checks for librt. The absence of librt is detected, but I have not yet used this information to conditionally remove librt.

Only cmake files have been modified.


stigrj:
After doing some git archeology I found that the ``-lrt`` option was introduced as a dependency (I think) along with the ``timer.h`` replacement for the ``boost::timer`` about a year ago. This timer was later replaced with the ``chrono`` timer which is the one currently in use, but the ``-lrt`` option was never removed. I think it is safe to remove librt altogether, at least it compiles and runs on both my laptop and on Stallo without it.

ilfreddy:
I removed everything that had to do with librt now (i hope...)
The only remaining fix for mac is the inclusion of `libstdc++` when linking `mrchem.x`

stigrj:
Did we get to the bottom of this? Was libstdc++ really necessary?

ilfreddy:
According to @robertodr, it should not be necessary if one sets up the compiler properly. We also noticed (mentioning it so it is on the records) that basically all compiler flags are absent for mac.
Another thing @robertodr mentioned is that we should be able to simplify our CMake infrastructure even further (compared e.g. to pcmsolver) because we have only C++

stigrj:
Ok, so should I merge or should we do it correctly first?

ilfreddy:
Changed to WIP. Still some stuff is not configured properly on a mac (CXX flags should be fixed)

ilfreddy:
I guess it can be merged. For some reason the GCC/G++ under mac are recognized as clang.

src/mrchem/CMakeLists.txt
```diff
@@ -65,7 +65,12 @@ target_link_libraries(mrchem.x
     ${BLAS_LIBRARIES}
 )
 
-option (MRCHEM_DRIVER_DEBUG "Debug the driver (extmod)" OFF)
+
+#if (APPLE)
+#    target_link_libraries(mrchem.x stdc++)
```
stigrj: Could you remove this if it is not needed anymore?

PR #60, opened 2017-12-10
[WIP] Xcfun update

ilfreddy:
The xcfun library has been revised significantly and the new interface is no longer compatible with MRChem.


src/mrchem/qmfunctions/OrbitalAdder.cpp
```diff
@@ -80,14 +81,14 @@ void OrbitalAdder::operator()(Orbital &out,
 
     // set output spin
     for (int i = 0; i < inp.size(); i++) {
-        if (fabs(coefs[i]) < MachineZero) continue;
+        if (abs(coefs[i]) < MachineZero) continue;
```
robertodr: `std::abs` to make 100% sure that the correct overload is selected.

CMakeLists.txt
```diff
@@ -41,5 +52,6 @@ include(mw_filters)
 include(main)
 include(feature_summary)
 include(autocmake_save_flags)
+include(FindLibrt)
```
stigrj: Why is librt still here? Didn't we find that it was unnecessary?

src/mrchem/qmoperators/CMakeLists.txt
```diff
@@ -1,3 +1,6 @@
+set (DEBUG_qmoperators TRUE)
```
stigrj: What is this debug stuff? Should it be here in the end or is it some experimentation to get gdb to work?

src/mrchem/qmoperators/XCFunctional.cpp
```diff
@@ -5,151 +5,50 @@
...
+void XCFunctional::evalSetup(const int order)
+{
+    unsigned int func_type = this->isGGA(); //only LDA and GGA supported for now
+    unsigned int dens_type = 1 + this->spin; // only n (dens_type = 1) or alpha & beta (denst_type = 2) supported now.
+    unsigned int mode_type = 1; //HACK HARD CODED
+    unsigned int laplacian = 0; // no laplacian
+    unsigned int kinetic = 0; // no kinetic energy density
+    unsigned int current = 0; // no current density
+    unsigned int explicit_derivatives = 0; // only gamma-type derivatives now (soon to be changed!)
+    std::cout << "In evalSetup. Order  " << order << " Dens " << dens_type << " Func " << func_type << std::endl;
```
stigrj: You can use the (slightly) simpler `printout`and `println`macros: 
```
println(0, "In evalSetup. Order " << order << ...);
```
First argument is print level, `println`includes a trailing `std::endl`. At least any "official" printing should be done in this way.

src/mrchem/qmoperators/XCOperator.cpp
```diff
@@ -197,7 +241,7 @@ void XCOperator::clearXCInput() {
     if (spin) this->xcInput[1] = 0;
 
     // the rest should be deleted
-    this->xcInput = deletePtrArray<FunctionTree<3> >(nInp, &this->xcInput);
+    this->xcInput = deletePtrArray<FunctionTree<3> >(nInp, &this->xcInput);  //LUCA: is this enough? Are we not leaving garbage around?
```
stigrj: Yes, this properly deallocates all `FunctionTree` pointers that are not NULL, as well as `xcInput` itself (which is a pointer to pointers)

src/mrchem/qmoperators/XCPotential.h
```diff
@@ -14,21 +14,58 @@
...
+class XCPotential {
 public:
-    XCPotential(XCFunctional &F, OrbitalVector &phi, DerivativeOperator<3> *D = 0)
-        : XCOperator(1, F, phi, D) { }
-    virtual ~XCPotential() { } 
-
-    virtual void setup(double prec);
-    virtual void clear();
+    XCPotential(int dO = 1, int dI = 0);
+    virtual ~XCPotential() {
```
stigrj: I would make this a one-liner
```
~XCPotential() { clear(); }
```
and remove the `virtual` so that no unnecessary `vtable` is made for this class.

src/mrcpp/CMakeLists.txt
```diff
@@ -28,10 +28,14 @@ target_link_libraries(mrcpp
     mwtrees
     mwoperators
     mwbuilders
-    rt
 )
 
-install(TARGETS mrcpp
+if (UNIX AND NOT APPLE)
+    target_link_libraries(mrcpp rt)
```
stigrj: Again, I think the librt is not necessary.

src/mrchem/scf_solver/OrbitalOptimizer.cpp
```diff
@@ -67,7 +67,7 @@ bool OrbitalOptimizer::optimize() {
         printCycle(nIter);
         adjustPrecision(err_o);
         orb_prec = getOrbitalPrecision();
-
+
```
stigrj: Please remove white spaces in this routine

src/mrchem/qmoperators/CMakeLists.txt
ilfreddy: It will go. Just some testing to activate debug selectively for each folder. I think it is a good feature to have, although this realization might not be the best.

src/mrchem/qmoperators/XCFunctional.cpp
ilfreddy: Thanks!
This is anyhow debugging "garbage". I will remove all the `std::cout` stuff once I am ready to merge with the main branch

src/mrcpp/CMakeLists.txt
ilfreddy: You are right. I guess I was doing that in another branch that has not been merged.

stigrj:
Covered by #90

PR #68, opened 2018-02-27
CMake self update

stigrj:


CMakeLists.txt
```diff
@@ -1,18 +1,18 @@
 # This file is autogenerated by Autocmake v1.0.0-alpha-x http://autocmake.org
-# Copyright (c) 2015-2017 by Radovan Bast, Jonas Juselius, and contributors.
+# Copyright (c) 2015-2018 by Radovan Bast, Roberto Di Remigio, Jonas Juselius, and contributors.
 
 # set minimum cmake version
 cmake_minimum_required(VERSION 2.8 FATAL_ERROR)
```
robertodr: Any reason not to do the jump to CMake 3.3? Note that the documentation page you land on when googling it's the one for CMake v3.

CMakeLists.txt
```diff
@@ -1,18 +1,18 @@
...
 
 # project name
-project(MRChem)
+project(MRChem C CXX)
```
robertodr: Is there actually any plain C in MRChem?

cmake/autocmake.yml
```diff
@@ -1,6 +1,10 @@
 name: MRChem
-
 min_cmake_version: 2.8
```
robertodr: Ditto about `min_cmake_version`

cmake/autocmake.yml
```diff
@@ -1,6 +1,10 @@
...
+setup_script: setup
+default_build_type: release
+language:
+  - C
```
robertodr: Ditto about C

cmake/autocmake.yml
```diff
@@ -25,11 +29,11 @@ modules:
 - libs:
   - source:
     - 'custom/eigen.cmake'
+    - 'custom/catch.cmake'
 - tests:
```
robertodr: You can fold the `docopt` and `define` into the `custom/tests.cmake` module:
```
# autocmake.yml configuration::
#
#   docopt: "--enable-tests Enable tests [default: False]."
#   define: "'-DENABLE_TESTS={0}'.format(arguments['--enable-tests'])"
```

CMakeLists.txt
stigrj: I don't know, is there any reason pcmsolver is not using cmake 3.3?  :stuck_out_tongue: I will change this.

CMakeLists.txt
stigrj: No, but xcfun requires C compiler options, and this was the only way could get this info passed down to xcfun...

CMakeLists.txt
robertodr: Not breaking compatibility with downstream programs that are extremely conservative with respect to their dependencies. And I think you know which ones :smile:

CMakeLists.txt
robertodr: You can do `enable_language(C)` right before setting up XCFun as external submodule to get those.

stigrj: Thanks!

stigrj:
Thanks for the help, @robertodr ! What should I use instead of ``include_directories`` etc?

robertodr:
`target_include_directories`, `target_link_libraries`, `target_compile_options` and `target_compile_definitions`. These commands give you fine-grained control over which targets link/include etc etc what other stuff.

PR #69, opened 2018-02-27
Replace mrcpp

stigrj:
I removed the old src/mrcpp/ directory and replaced it with the new submodule. The pilot code successfully compiles and links with the new mrcpp, but the full mrchem code will not compile and is currently ignored by cmake. In the following I will add mrchem piece by piece by moving code one level up from src/mrchem/ and make it compile with the new mrcpp. At the same time I will add a corresponding test case and publish the changes in small PRs to this branch.


src/CMakeLists.txt
```diff
@@ -1,7 +1,7 @@
-include(DefaultPathsMRChem)
+#include(DefaultPathsMRChem)
```
robertodr: I think you forgot to remove this file.

external/CMakeLists.txt
```diff
@@ -7,6 +7,24 @@ add_custom_target(git-submodule-init
     WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}
     )
 
+ExternalProject_Add(submodule-mrcpp
+    DOWNLOAD_COMMAND git submodule update
```
robertodr: Are you sure that doing the update this way won't check out any changes made into the `mrcpp` submodule folder?

stigrj: Not at all, I just copied the submodule blocks we already have for getkw and xcfun

robertodr: It used to be that way in DALTON/LSDALTON/DIRAC, which caused tons of surprises.

stigrj:
I guess this is a good time to re-think the whole cmake and external projects stuff. Would it be possible to find some time for us (@robertodr , @ilfreddy ) to set everything up properly from scratch before @robertodr leaves? It shouldn't be much work at this stage as most of the code is excluded from compilation, we just need to add the dependencies (mrcpp, eigen, xcfun, getkw, catch) in a sensible way and write a couple of short CMakeLists.

robertodr:
No problem for me. I'm leaving on Monday

PR #70, opened 2018-03-01
Cleanup cmake

stigrj:
Please read these ~20 000 line changes carefully :stuck_out_tongue:
@robertodr, could you please look over the following files and identify any stupidity:

- CMakeLists.txt
- src/CMakeLists.txt
- src/qmfunctions/CMakeLists.txt
- src/qmoperators/CMakeLists.txt
- tests/CMakeList.txt
- external/CMakeLists.txt

.robertoignore:
- src/mrchem/*
- pilot/*


stigrj:
Not sure if `make install` is working as it should. The external modules are *not* installed along with the mrchem library and executable, but this may be as it should?

robertodr:
That's certainly wrong. I'll take a look.

cmake/custom/tests.cmake
```diff
@@ -7,6 +7,24 @@
 
 option(ENABLE_TESTS "Enable test suite" ON)
 
+macro(add_Catch_test _name _labels)
```
robertodr: I've rewritten this macro so that named arguments can be used. See PCMSolver/pcmsolver#130

external/CMakeLists.txt
```diff
@@ -8,55 +8,55 @@ add_custom_target(git-submodule-init
     )
```
robertodr: If you want to install the submodules, you need to add some `install` commands. The reason is that now we build the submodules and _stage_ their installation to `${CMAKE_BINARY_DIR}/external` (for the reasons we discussed together the other day). It should be enough to add the following at the end of this file:
```
install(
   DIRECTORY 
       ${CMAKE_BINARY_DIR}/stage
   DESTINATION
       ${CMAKE_INSTALL_PREFIX}
    USE_SOURCE_PERMISSIONS
    )
```

pilot/CMakeLists.txt
```diff
@@ -1,7 +1,31 @@
-include(DefaultPathsMRChem)
+link_directories(${CMAKE_BINARY_DIR}/stage/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})
```
robertodr: I would put this in the `target_link_libraries` command below.

pilot/CMakeLists.txt
```diff
@@ -1,7 +1,31 @@
...
+        mrchem
+        mrcpp
+        getkw
+        )
+    target_include_directories(mrchem-pilot.x PRIVATE
+        ${CMAKE_SOURCE_DIR}/src
+        ${CMAKE_SOURCE_DIR}/src/qmfunctions
+        ${CMAKE_SOURCE_DIR}/src/qmoperators
+        ${CMAKE_SOURCE_DIR}/external/mrcpp/${CMAKE_INSTALL_INCLUDEDIR}
+        ${CMAKE_SOURCE_DIR}/external/libgetkw/C++
```
robertodr: This should point to the staged installation of Getkw: `${CMAKE_BINARY_DIR}/external/libgetkw`, most likely.

pilot/CMakeLists.txt
```diff
@@ -1,7 +1,31 @@
...
+        )
+    configure_file(${CMAKE_SOURCE_DIR}/src/mrchem.in ${CMAKE_BINARY_DIR}/pilot/mrchem)
+    execute_process(COMMAND chmod 755 ${CMAKE_BINARY_DIR}/pilot/mrchem OUTPUT_QUIET)
```
robertodr: You can copy and adjust permissions of this file with (check that relative paths work)
```
file(
  COPY 
     src/mrchem.in
   DESTINATION
     pilot/mrchem
   FILE_PERMISSIONS
      OWNER_READ OWNER_WRITE OWNER_EXECUTE 
      GROUP_READ GROUP_EXECUTE
      WORLD_READ WORLD_EXECUTE
    )
```

src/CMakeLists.txt
```diff
@@ -1,7 +1,46 @@
-#include(DefaultPathsMRChem)
 
-#include_directories(mrcpp)
-#include_directories(mrchem)
+link_directories(${CMAKE_BINARY_DIR}/stage/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})
```
robertodr: Same comment here. I think this belongs into a `target_link_libraries`

src/CMakeLists.txt
```diff
@@ -1,7 +1,46 @@
...
+
+add_dependencies(mrchem external-modules)
+target_link_libraries(mrchem
+    mrcpp
+    )
+target_include_directories(mrchem PRIVATE
+    ${CMAKE_SOURCE_DIR}/src/qmfunctions
+    ${CMAKE_SOURCE_DIR}/src/qmoperators
+    ${CMAKE_SOURCE_DIR}/external/mrcpp/${CMAKE_INSTALL_INCLUDEDIR}
+    ${CMAKE_SOURCE_DIR}/external/libgetkw/C++
```
robertodr: Same here about using the staged installation.

src/CMakeLists.txt
```diff
@@ -1,7 +1,46 @@
...
+target_link_libraries(mrchem.x
+    mrchem
+    mrcpp
+    getkw
+    )
+target_include_directories(mrchem.x PRIVATE
+    ${CMAKE_SOURCE_DIR}/src/qmfunctions
+    ${CMAKE_SOURCE_DIR}/src/qmoperators
+    ${CMAKE_SOURCE_DIR}/external/mrcpp/${CMAKE_INSTALL_INCLUDEDIR}
+    ${CMAKE_SOURCE_DIR}/external/libgetkw/C++
```
robertodr: Ditto about using the staged installation.

src/CMakeLists.txt
```diff
@@ -1,7 +1,46 @@
...
+    )
+set (MRCHEM_EXECUTABLE ${CMAKE_INSTALL_PREFIX}/bin/mrchem.x)
```
robertodr: You should be using `${CMAKE_INSTALL_BINDIR}` as defined in `GNUInstallDirs.cmake` for consistency.

tests/CMakeLists.txt
```diff
@@ -1,6 +1,22 @@
-include(DefaultPathsMRChem)
-include_directories(${PROJECT_SOURCE_DIR}/src)
-include_directories(${PROJECT_SOURCE_DIR}/external/catch)
+include_directories(SYSTEM ${PROJECT_SOURCE_DIR}/external/catch)
+include_directories(${CMAKE_BINARY_DIR})
 
-link_directories(${CMAKE_BINARY_DIR}/external/lib)
+link_directories(${CMAKE_BINARY_DIR}/stage/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR})
```
robertodr: Same here, I think it can be moved to `target_link_libraries`

pilot/CMakeLists.txt
stigrj: Then the link statement needs to be this one?
```
target_link_libraries(mrchem-pilot.x
    mrchem
    ${CMAKE_BINARY_DIR}/stage/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/libmrcpp.a
    ${CMAKE_BINARY_DIR}/stage/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}/libgetkw.a
    )
```

pilot/CMakeLists.txt
stigrj: This command apparently correspond to
```
    $ cp src/mrchem.in pilot/mrchem/mrchem.in
```
e.i. not including a name change. Couldn't find any direct way to copy with name change

```
    $ cp src/mrchem.in pilot/mrchem
```

robertodr: Oh gee, yes, you're right! Shouldn't look at code before 8:00 AM! I would still suggest to do `configure_file` to configure the file (passing `@ONLY` to have CMake to only run substitutions on strings starting and ending with `@`) and `file(COPY`to change permissions.

pilot/CMakeLists.txt
robertodr: This should do:
```
target_link_libraries(mrchem-pilot.x
    mrchem
    ${CMAKE_BINARY_DIR}/stage/${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_LIBDIR}
    mrcpp
    getkw
    )
```

robertodr:
Consider adding this Autocmake module: https://github.com/coderefinery/autocmake/blob/master/modules/default_build_paths.cmake
It will make sure that binaries are stored under `<build-dir>/bin` and libraries under `<build-dir>/lib`.

stigrj:
The `default_build_paths.cmake` worked great, but it also moved `unit-tests.x` to `build/bin`...

robertodr:
Just redefine the `add_Catch_test` macro to pick up the executable from there?

PR #72, opened 2018-03-07
OrbitalVector

stigrj:
Reintroduce `OrbitalVector` using the new mrcpp, MPI routines not included.

Some changes along the way that I hope will make it more dynamic and user friendly. The idea is to properly implement assignment operators and copy constructors for `Orbital` and `OrbitalVector`, both as shallow copies, e.i. complete copy of meta data but shallow copy of `FunctionTree` pointers. The point of this is to avoid the use of pointers and references for `Orbital` and `OrbitalVector` and use "real" objects instead, with the benefit that the `FunctionTree` pointer ownership can be easily transferred to another `Orbital` while the old `Orbital` goes out of scope and dies (but *you* need to keep track of this ownership and free the memory when appropriate). This change was already introduced for the `Orbital` somewhat under the radar in the previous PR, but should be considered again here.

Now `OrbitalVector` contains a `std::vector<Orbital>` instead of a vector of pointers, and the `QMOperator` takes `Orbital` copies (instead of reference) as arguments and returns `Orbital` objects instead of pointers.

Please have a look at the unit tests to see how it works.


src/qmfunctions/OrbitalVector.h
```diff
@@ -0,0 +1,69 @@
+#pragma once
+
+#include <vector>
+using std::vector;
```
robertodr: This will import `std::vector` in all the classes that `include` this file. Maybe tuck it behind the `mrchem` namespace?

stigrj:
Following @robertodr 's suggestion, I greatly simplified the OrbitalVector which is now nothing more than a `typedef std::vector<Orbital>`. I *think* this still meets our needs, although I'm not able to see all ends at the moment.

robertodr:
You can blame me if it's not completely up to snuff :rofl:

PR #74, opened 2018-03-13
Testing and coverage

stigrj:
Set up Travis CI and codecov


.travis.yml
```diff
@@ -0,0 +1,50 @@
...
+matrix:
+  include:
+    - os: linux
+      addons:
+        apt:
+          packages:
+            - cmake3
+            - cmake3-data
+            - g++-4.8
+            - gfortran-4.8
```
robertodr: Why is the dreaded Fortran needed?

.travis.yml
```diff
@@ -0,0 +1,50 @@
...
+            - libeigen3-dev
+            - libboost-math-dev
+      env:
+        - BUILD_TYPE='release'
+        - CXX_COMPILER='g++-4.8'
+        - TESTS='--enable-tests'
```
robertodr: I think you'll always want tests to run on travis. I'd hardwire this flag into the setup line.

.travis.yml
```diff
@@ -0,0 +1,50 @@
...
+  - cmake --version
+  - ${CXX_COMPILER} --version
+
+script:
+  - export OMP_NUM_THREADS=2
+  - ./setup --type=${BUILD_TYPE} --cxx=${CXX_COMPILER} ${TESTS} ${COVERAGE} ${OPENMP}
+  - cd build
+  - cmake --build . -- VERBOSE=1
+  - |
+    if [[ "${TESTS}" = "--enable-tests" ]]; then
```
robertodr: If you always enable tests this is redundant.

.travis.yml
stigrj: Just trying to get a successful build, xcfun asks for a fortran compiler: https://travis-ci.org/stigrj/mrchem/builds/352748045

stigrj:
@robertodr do you think we can use the same encrypted slack notification token for mrchem as we already have for mrcpp?

robertodr:
I would think not.

PR #78, opened 2018-04-08
Use standard FindBLAS, just need to specify cblas header file.

susilehtola:
This pull request changes the way the mrchem can be built with BLAS support.

FindBLAS has been a standard thing in CMake for a long time, so the script Jonas wrote ages ago is no longer necessary.

However, FindBLAS doesn't have CBLAS support, so the header file needs to be specified separately.

I would recommend switching to standard BLAS though, because it's much more available than CBLAS. E.g. Armadillo has a really nice C++ wrapper to BLAS, so you don't have to worry about the C <=> Fortran interfacing.


stigrj:
Thanks! Jonas' old script doesn't even work anymore, so this change was long overdue.

Please note that there are some big changes under way in the code structure. The `mrcpp` source directory has been extracted to a separate repo (https://github.com/MRChemSoft/mrcpp) and included as an external project to MRChem. This is done in the `new-mrcpp` branch of MRChem, and the current `master` will be gone as soon as the new version catches up (should be a matter of weeks).

robertodr:
i second the suggestion of having your own C wrapper to BLAS/LAPACK. The mangling can be taken care quite easily using the `FortranCInterface` module of CMake. @stigrj if merging this one will conflict with `new-mrcpp`, open issues and assign them to a milestone so they are not forgotten.

susilehtola:
So I did what you asked. I'm still not sure why you want to use scripts around CMake, which is simple enough to write by itself :stuck_out_tongue:

PR #81, opened 2018-04-09
Use normal BLAS instead of CBLAS.

susilehtola:
This pull request implements a wrapper for ddot and dgemv so that BLAS detection works out of the box.

If you need 64-bit interface support, that can be done just by customizing the headers in blas.h.


susilehtola:
No longer needed due to #82 .

PR #82, opened 2018-04-10
Remove cblas calls, use EIGEN_USE_BLAS instead.

susilehtola:
As requested in #81, this is a pull request that eliminates CBLAS in favor of BLAS support inside Eigen3.

Note that to actually compile the code against the present version of Eigen3, you need to address a bug in Eigen3 by adding a missing include statement
```
$ hg diff
diff -r 3a08a093adac Eigen/src/Core/products/TriangularMatrixMatrix_BLAS.h
--- a/Eigen/src/Core/products/TriangularMatrixMatrix_BLAS.h     Mon Apr 09 13:29:26 2018 +0200
+++ b/Eigen/src/Core/products/TriangularMatrixMatrix_BLAS.h     Tue Apr 10 15:51:12 2018 +0300
@@ -33,6 +33,8 @@
 #ifndef EIGEN_TRIANGULAR_MATRIX_MATRIX_BLAS_H
 #define EIGEN_TRIANGULAR_MATRIX_MATRIX_BLAS_H

+#include "../../misc/blas.h"
+
 namespace Eigen {

 namespace internal {

```

### DIFF
--- a/cmake/custom/blas.cmake
+++ b/cmake/custom/blas.cmake
@@ -1,7 +1,8 @@
-option(BLAS_H "CBLAS header file" "")
-if(BLAS_H)
+option(ENABLE_BLAS "Enable use of BLAS? (Requires a Fortran compiler)" OFF)
+if(ENABLE_BLAS)
   include(FindBLAS)
   if(BLAS_FOUND)
-    set(HAVE_BLAS 1)
+    # Enable BLAS in Eigen
+    add_definitions(-DEIGEN_USE_BLAS)
   endif(BLAS_FOUND)
-endif(BLAS_H)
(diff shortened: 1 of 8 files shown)


robertodr:
Which version of Eigen is bugged? I can link against BLAS with Eigen 3.3.3 without problems. @stigrj maybe it's time to update Eigen?

robertodr:
Weird... I'm in favor of pulling this in as-is and then figuring out the issue. @stigrj?

P.S.: The example I was compiling is [this one](https://github.com/bast/cmake-recipes/blob/master/chapter-03/recipe-07/cxx-example-3.5/CMakeLists.txt)

PR #90, opened 2018-04-17
Xcfun update2

ilfreddy:
PR for the NEW xcfun in the NEW mrchem. 
Now we can use xcfun both with closed and open-shell (as before) and we can also choose whether to use explicit derivatives (wrt to density and gradient) or gamma-type ones (wrt to density and gamma). The former seems to be more stable (less affected by numerical instabilities in some functionals, such as f.ex. BLYP), and it should be easier to extend to higher order derivatives (hessian and further...)

The code has been tested on both closed and open shell cases..

### DIFF
--- a/src/qmoperators/two_electron/XCFunctional.cpp
+++ b/src/qmoperators/two_electron/XCFunctional.cpp
@@ -1,15 +1,14 @@
-#include "GridGenerator.h"
-#include "MWAdder.h"
-#include "MWMultiplier.h"
-#include "MWDerivative.h"
+#include "mrchem.h"
 #include "XCFunctional.h"
-#include "TelePrompter.h"
 #include "constants.h"
 
+using namespace mrcpp;
 using namespace std;
 using namespace Eigen;
 
 extern MultiResolutionAnalysis<3> *MRA;
+
+namespace mrchem {
 
 /** @brief constructor
  *
@@ -831,3 +830,4 @@ FunctionTree<3>* XCOperator::calcDotProduct(FunctionTreeVector<3> &vec_a,
     return out;
 }
 
+} //namespace mrchem
--- a/src/qmoperators/two_electron/XCFunctional.cpp
+++ b/src/qmoperators/two_electron/XCFunctional.cpp
@@ -1,0 +1,833 @@
+#include "GridGenerator.h"
+#include "MWAdder.h"
+#include "MWMultiplier.h"
+#include "MWDerivative.h"
+#include "XCFunctional.h"
+#include "TelePrompter.h"
+#include "constants.h"
+
+using namespace std;
+using namespace Eigen;
+
+extern MultiResolutionAnalysis<3> *MRA;
+
+/** @brief constructor
+ *
+ * Initializes the new functional
+ * 
+ * @param[in] s True for spin-separated calculations
+ * @param[in] thrs Threshold for func calculation
+ *
+ */
+XCFunctional::XCFunctional(bool s, bool e, double thrs)
+        : spin(s), cutoff(thrs) {
+    this->functional = xc_new_functional();
+    if(e) {
+        this->expDerivatives = 1;
+    }
+    else {
+        this->expDerivatives = 0;
+    }        
+}
+
+/** @brief destructor
+ *
+ */
+XCFunctional::~XCFunctional() {
+    xc_free_functional(this->functional);
+}
+
+/** @brief functional setup
+ *
+ * @usage For each functional part in calculation a corresponding token is created in xcfun
+ *
+ * @param[in] name The name of the chosen functional
+ * @param[in] coef The amount of the chosen functional
+ */
+void XCFunctional::setFunctional(const string &name, double coef) {
+    xc_set(this->functional, name.c_str(), coef);
+}
+
+/** @brief User-friendly setup of the xcfun calculation
+ *
+ * Setup the XC functional for evaluation. In MRChem we use only a subset of the alternatives offered by xcfun. 
+ * More functinality might be enabled at a later stage.
+ *
+ * @param[in] order Order of the requested operator (1 for potential, 2 for hessian, ...)
+ * 
+ */
+void XCFunctional::evalSetup(const int order)
+{
+    unsigned int func_type = this->isGGA();  //!< only LDA and GGA supported for now
+    unsigned int dens_type = 1 + this->spin; //!< only n (dens_type = 1) or alpha & beta (denst_type = 2) supported now.
+    unsigned int mode_type = 1; //!< only derivatives (neither potential nor contracted)
+    unsigned int laplacian = 0; //!< no laplacian
+    unsigned int kinetic = 0;   //!< no kinetic energy density
+    unsigned int current = 0;   //!< no current density
+    if(this->isLDA()) { // Fall back to gamma-type derivatives if LDA (bad hack: no der are actually needed here!)
+        this->expDerivatives = 0;
+    }
+    xc_user_eval_setup(this->functional, order, func_type, dens_type, mode_type, laplacian, kinetic, current, this->expDerivatives);
+}
+    
+
+/** \breif Evaluates XC functional and derivatives
+ *
+ * Computes the alpha and beta exchange-correlation functionals and
+ * their derivatives.  The electronic density (total/alpha/beta) and their gradients are
+ * given as input. Results are then stored in the xcfun output
+ * functions. Higher order derivatives can be computed changing the parameter k. 
+ *
+ * XCFunctional output (with k=1 and explicit derivatives):
+ *
+ * LDA: \f$ \left(F_{xc}, \frac{\partial F_{xc}}{\partial \rho}\right) \f$
+ *
+ * GGA: \f$ \left(F_{xc},
+ *  \frac{\partial F_{xc}}{\partial \rho},
+ *  \frac{\partial F_{xc}}{\partial \rho_x},
+ *  \frac{\partial F_{xc}}{\partial \rho_y},
+ *  \frac{\partial F_{xc}}{\partial \rho_z}\right) \f$
+ *
+ * Spin LDA: \f$ \left(F_{xc}, \frac{\partial F_{xc}}{\partial \rho^\alpha},
+ *  \frac{\partial F_{xc}}{\partial \rho^\beta}\right) \f$
+ *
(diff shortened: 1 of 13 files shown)


src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,157 @@
+#pragma once
+
+#pragma GCC system_header
```
stigrj: You don't need this `system_header` line anymore

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,157 @@
...
+    int getOutputLength() const { return xc_output_length(this->functional); }
+
+    bool isLDA() const { return (!(this->isGGA() || this->isMetaGGA())); }
+    bool isGGA() const { return (xc_is_gga(this->functional)); }
+    bool isMetaGGA() const { return (xc_is_metagga(this->functional)); }
+    
+    bool isSpinSeparated() const { return this->spin; }
+    bool needsGamma() const { return (expDerivatives == 0);};
+
+    void evaluate(int k, Eigen::MatrixXd &inp, Eigen::MatrixXd &out) const;
```
stigrj: I've made `typedefs` of the `Eigen` names (not sure @robertodr would like this, though :stuck_out_tongue:) This would be a `DoubleMatrix`. More can be found in `src/mrchem.h`

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,157 @@
...
+/** 
+ *  \class XCFunctional
+ *  \brief Compute XC functional with XCFun
+ *
+ *  Interface class for the XCFun library
+ *
+ *  \author Stig Rune Jensen
+ *  \date 2015
+ *  
+ */
```
stigrj: There should be a `namespace mrchem` enclosing all this

src/qmoperators/two_electron/XCOperator.h
```diff
@@ -0,0 +1,29 @@
...
+ * \class XCOperator
+ * \brief Exchange and Correlation operator
+ *
+ * Notes for myself: I need a way to "select" the correct potential to apply
+ */
+namespace mrchem {
+    
+class XCOperator final : public RankZeroTensorOperator {
+ public:
+ XCOperator(mrcpp::XCFunctional &F, OrbitalVector &Phi, mrcpp::DerivativeOperator D, int k)
```
stigrj: The `XCFunctional` is not a `mrcpp` feature

src/qmoperators/two_electron/XCPotential.h
```diff
@@ -0,0 +1,52 @@
...
+    ~XCPotential() { }
+
+    void setup(double prec);
+    void clear();
+
+    double getEnergy() const { return this->energy; }
+    void setup(double prec);
+    void clear();
+
+    Orbital apply (Orbital phi);
```
stigrj: This should be `protected`

src/qmoperators/two_electron/XCFunctional.h
robertodr: It's not that I don't like it (I don't, by the way) it's more that then you don't see where this type comes from or where it is defined. You could as well use `auto` at that point (not that it would work, I think)

ilfreddy: Changed it. Just to upset @robertodr a little

robertodr: :angry: if `inp` is an input matrix, you can pass by `const` reference to make sure that the compiler will complain if you modify it internally.

src/qmoperators/two_electron/XCFunctional.cpp
```diff
@@ -808,18 +805,14 @@ FunctionTree<3>* XCOperator::calcDotProduct(FunctionTreeVector<3> &vec_a,
...
-
     FunctionTreeVector<3> out_vec;
     for (int d = 0; d < vec_a.size(); d++) {
         FunctionTree<3> &tree_a = vec_a.getFunc(d);
         FunctionTree<3> &tree_b = vec_b.getFunc(d);
         FunctionTree<3> *out_d = new FunctionTree<3>(*MRA);
-        grid(*out_d, tree_a);
-        grid(*out_d, tree_b);
-        mult(*out_d, 1.0, tree_a, tree_b, 0);
+        //ASK STIG: which grid shall I copy here?
```
stigrj: I guess it doesn't really matter here since all grids are equal anyway, but you can copy several grids:
```
copy_grid(*out_d, tree_a);
copy_grid(*out_d, tree_b);
```
which will build the union of the two grids.

stigrj:
You should do `git rebase upstream/new-mrcpp`, resolve any conflicts and `git push -f origin/xcfun_update2`. This will remove the first 5 commits, as they are already incorporated into `new-mrcpp` (but with different hashes). Also, I think you are not using the latest `mrcpp/master` since you are able to `#include "FunctionNode.h"`. This should now be `#include "MRCPP/mwtrees/FunctionNode.h"`

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,160 @@
+#pragma once
+
+#include <Eigen/Core>
+
+#include "FunctionTree.h"
+#include "FunctionTreeVector.h"
+#include "xcfun.h"
+#include "Density.h"
+#include "MRCPP/Printer"
```
stigrj: Try to move this to the source file.

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,160 @@
+#pragma once
+
+#include <Eigen/Core>
+
+#include "FunctionTree.h"
+#include "FunctionTreeVector.h"
```
stigrj: The `FunctionTree` headers should be replaced by `MRCPP/MWFunctions`.

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,160 @@
...
+    mrcpp::FunctionTree<3> * addPotentialContributions(mrcpp::FunctionTreeVector<3> & contributions,
+                                                int maxScale);
+    mrcpp::FunctionTree<3> * calcDivergence(mrcpp::FunctionTreeVector<3> &inp,
+                                    mrcpp::DerivativeOperator<3> *derivative,
+                                    int maxScale);
+    mrcpp::FunctionTree<3> * calcGradDotPotDensVec(mrcpp::FunctionTree<3> &V,
+                                           mrcpp::FunctionTreeVector<3> &rho,
+                                           mrcpp::DerivativeOperator<3> *derivative,
+                                           int maxScale);
+    mrcpp::DerivativeOperator<3> *derivative;  ///< External derivative operator
```
stigrj: Could you move all the `protected` functions below the `protected` data members?

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,160 @@
...
+        return 0;
+    }
+
+private:
+    bool spin;                  ///< Spin polarization
+    unsigned int expDerivatives;///< whether gamma-type or explicit derivatives are used
+    double cutoff;              ///< Below the cutoff value, the density will be considered zero
+    xc_functional functional;   ///< The functional in the XCFun library (struct from xcfun library)
+    int max_scale; //HACK: the old XCOperator used to inherit this. Wher do I get it from now?
+    OrbitalVector orbitals;     ///< Set of orbitals used to compute the density defining the functional
```
stigrj: The `OrbitalVector` should be a pointer. It needs to reference the actual vector that is used in the SCF since the orbitals are expected to change during the optimization.

src/qmoperators/two_electron/XCFunctional.h
```diff
@@ -0,0 +1,160 @@
...
+        }
+        return 0;
+    }
+
+private:
+    bool spin;                  ///< Spin polarization
+    unsigned int expDerivatives;///< whether gamma-type or explicit derivatives are used
+    double cutoff;              ///< Below the cutoff value, the density will be considered zero
+    xc_functional functional;   ///< The functional in the XCFun library (struct from xcfun library)
+    int max_scale; //HACK: the old XCOperator used to inherit this. Wher do I get it from now?
```
stigrj: The `max_scale` parameter should be removed. It is now included in the `MRA`, and I can't see that you actually use this parameter anyway.

ilfreddy:
Indeed I am not using the `master` branh for mrcpp but `mrchem-version`. Should I switch to master now?

src/qmoperators/two_electron/XCFunctional.h
ilfreddy: It is there.

src/qmoperators/two_electron/XCFunctional.h
ilfreddy: Done. But I believe a good deal of cleanup here is needed soon.

src/qmoperators/two_electron/XCFunctional.h
stigrj: The density needs to be rebuilt every time `setup(prec)` is run on the `XCOperator` at least. But I think the `XCFunctional` should be constructed only once in the beginning, so I think you should move density building out of the constructor.

src/qmoperators/two_electron/XCFunctional.h
ilfreddy: Tried. But there are a is a template function with `MSG_ERROR` which need Print. Unless there is a nicer solution which as usual I am not aware of.

stigrj:
Please consider my [PR](https://github.com/ilfreddy/mrchem/pull/3) on your fork. I will wait until your changes have been incorporated into XCFun before I merge this, so that we can point our submodule to a proper version.

ilfreddy:
Done! My xcfun stuff is also ready to be merged. Hopefully the formatting nazi (aka @robertodr ) is happy about it now. Once he approves it, we are good to go. Feels good cleaning up this submodule hell finally!

ilfreddy:
Good to go 👍

stigrj:
Why does Travis build an old commit?

stigrj:
I made Travis build the right commit on my fork, and the same missing Fortran compiler error is still there.

stigrj:
Ok, great work @ilfreddy, finally full DFT support!

PR #94, opened 2018-05-03
[WIP] Update libgetkw

stigrj:
Move to a more recent libgetkw with improved CMake setup. This removes some annoying compiler warnings on my laptop.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,6 @@
 [submodule "external/libgetkw"]
 	path = external/libgetkw
-	url = https://github.com/juselius/libgetkw.git
+	url = https://github.com/dev-cafe/libgetkw.git
 [submodule "external/xcfun"]
 	path = external/xcfun
 	url = https://github.com/dftlibs/xcfun.git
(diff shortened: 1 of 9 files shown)


stigrj:
Travis complains on the `Boost::boost` target. Any ideas?

robertodr:
Boost is found but it seems the imported target `Boost::boost` isn't propagated from the `getkw` target to dependents (in this case the `cxx_test`) Can you try a serial build?

robertodr:
I can't see anything that is obviously wrong. The set up is the exact same as used when compiling Getkw on its own (apart from `-DCMAKE_C_COMPILER` which is no longer needed).

stigrj:
I get the same error on stallo, but not on my laptop

robertodr:
I think the problem is that Getkw is a different project, hence its dependencies are not exported and transitively applied to MRChem. This will need work a little work Getkw-side. In the meantime you can search for Boost on your side, pass it down to Getkw (so the same version is used) and then list `Boost::boost` in `target_link_libraries` when you build the executable.

robertodr:
Closing as it has been superseded by #119

PR #96, opened 2018-05-06
[WIP] Build MRChem as a superbuild

robertodr:
This is motivated by the recent changes in MRCPP, XCFun and Getkw which can now all be consumed as CMake imported targets.
it is obviously not ready for public consumption, hence the WIP status.

### TODO
- [x] Switch origin for MRCPP submodule back to the canonical org.
- [x] Update XCFun to its latest and greatest.
- [x] Update Getkw to its latest and greatest. Will subsume #94 but needs work Getkw-side, ongoing.
- [x] Switch to Eigen 3.3 and use imported target thereof. I'll probably drop this point, it will come _via_ MRChemSoft/mrcpp#27, most likely.
- [ ] Recast `examples` as integrations tests for `mrchem.x`
- [ ] Build `libmrchem` and test it separately from `mrchem.x`

### Questions
- How fond are you of the `git-submodule-init` custom target? I think it's a nuisance because it changes sources in the source tree at build time. But I understand that you might not be used to not do `git clone --recursive` and/or `git submodule sync && git submodule update --init --recursive` and you'd like to keep it.
- The main `CMakeLists.txt` is generated by Autocmake, but probably the one that should be autogenerated is the one inside `src`.


robertodr:
Travis says: 
```
error: Server does not allow request for unadvertised object 295a3f955776a0b0077b9deead257d42fc909b17
Fetched in submodule path 'external/mrcpp', but it did not contain 295a3f955776a0b0077b9deead257d42fc909b17. Direct fetching of that commit failed.
```
Is it because MRCPP is pointing to `robertodr/mrcpp`? Does anybody know how to get around this?

stigrj:
Which commit is it supposed to point to? It could be rebase related, that your history does not coincide with `upstream/mrcpp`.

robertodr:
OK, it's running now. Time for a break. Next up is Getkw.

[I had a branch named `includes` that was in a weird state, removing it and checking it out again got me to the right state of `robertodr/mrcpp`]

robertodr:
I have to pause on this until some other time this week...

stigrj:
Any chance of getting this before Berlin July 23?

robertodr:
This needed work on Getkw and I got a bit CMake-ed out... Apologies for the delay. I should be able to finish up within the deadline though :smile_cat:

stigrj:
Ok, so the build process looks smooth and everything seems to compile properly. However, the resulting executables are not working. The `mrchem` python script turns out `zsh: exec format error:`, and the full executable `mrchem.x` crashes immediately. The test suite does not compile out of the box, but I managed to tweak it, and at least most of the test cases finished successfully, although some failed.

robertodr:
I knew this was far from perfect, but what you report is really a carnage :smile: 
1. I managed to get all tests compiled and running, but since there is no test for the actual `mrchem.x` executable I have no idea if it's valid or not. What is the crash you get?
2. The Python script might have its paths screwed up because variables in a `configure_file` might not have been set correctly.
3. `Boost::boost` what version of CMake and what version of Boost on stallo? The two things are sadly not independent.

Some more general points. As you have noticed, the `src/CMakeLists.txt` has become a project in itself. The flow within CMake is now: top `CMakeLists.txt` (generated by Autocmake) --> `main.cmake` --> `external/CMakeLists.txt` (this search for the dependencies and builds them if not found) --> `src/CMakeLists.txt` (as an `ExternalProject`, so that we can pass full-formed targets for the dependencies in)
The motivation is to not carry around `*_INCLUDE_DIRS` and `*_LIBRARIES`, since these can become quite brittle and break things easily down the line. Now, there are clearly problems to this approach. I am quite unsure how to proceed further, because any option I've come up with boils down to significant reorganization of directories. This might be too disruptive and I need you to weigh in. 
**Problem** What to do with the unit tests? They are in a separate subdirectory, which knows nothing about the MRChem project in `src/CMakeLists.txt`. As soon as we're out of that project, it's a clean slate: `libmrchem.a`, `mrchem`, and `mrchem.x` are unkwown in the `tests` subdirectory.
   + **Solution a** Move the tests within `src/CMakeLists.txt` so that everything belongs together into the `MRChem_core` project. This is relatively straightforward to do. I'd suggest renaming `src` to `mrchem` to make the geography a bit cleaner:
    ```
    mrchem
    |-- external
    |-- mrchem
         |-- cmake
         |-- src
         |-- tests
    ```
   In addition, it would make sense to move the Autocmake setup _down_ one level, whereas the "steering" `CMakeLists.txt` at the top-level would be managed by hand.
   + **Solution b** Make tests yet another external project, which accepts the location of the MRChem freshly installed target. We already have the target (though probably not well-configured) I have experimented with this, but it is unclear if we can then run `ctest` and have meaningful reporting without hopping around in subdirectories of `build`.
    + **Solution c** The one I have currently adopted and just pushed. Within `src/CMakeLists.txt` do `add_subdirectory(../tests)` This works, because if mimicks solution a and I can get tests to run. If I hop into the build directory for the `MRChem_core` project.

These are the solutions I can think of and the drawbacks of each. It seems to me it's a choice of the lesser evil, and probably that's staying put with what is currently in place and is known to work. I have no problems in nixing this PR if that's the conclusion we reach :+1: 

P.S.: Why is Travis not running anymore for this PR?

stigrj:
1. You can test the main executable on one of the example inputs, e.g. `examples/he/mrchem.inp`
    ```
   $ ./mrchem mrchem.inp
   ```
3. Must have been an old CMake version on Stallo, I tried again with v3.9.5 and it works. Here also `mrchem.x` seems to work, but still not the `mrchem` script.

stigrj:
Travis complains: "could not parse .travis.yml"

stigrj:
Regarding the proposed solutions, do you know of any other codes with a similar setup?

PS: `--enable-tests` is no longer a valid option to setup (not sure if this was intentional?)

robertodr:
Yes, `--enable-tests` removal was intentional and _momentary_. Obvious inspiration is Psi4, but there the difference is that the tests are integration tests, rather than unit tests so they only need to know the location of the final executable. I tried to set up a "mock" target for MRChem, but then the dependencies on MRCPP, Getkw, Eigen, and so forth are not propagated (which makes the whole PR kind of pointless)

stigrj:
Hmm, the `libmrchem` target requires unit tests, but we should have integration tests on top as well that runs executables. Perhaps separate `libmrchem` w/ unit tests from the executable and have only integration tests on the main project?

robertodr:
Hmmmm. OK, let me see I understand correctly:
```
mrchem
|-- CMakeLists.txt --> This builds mrchem.x and does the integration testing
|-- external
|-- mrchem
    CMakeLists.txt --> This builds libmrchem.a (or .so) and does the unit testing
    |-- cmake
    |-- src
    |-- tests
```

stigrj:
Something like that 🙂

robertodr:
OK. Let me see what I can do. I'll turn the `examples` into integration tests. If at any point you judge these changes as too much, know that it's totally fine for me if you reject them and scrap this PR. We can always revisit/reopen in the future.

stigrj:
No, this is definitely moving in the right direction, if we can just manage to wrap it up nicely. I anyway wanted to separate a middle layer between MRCPP and the top level MRChem in order to fully encapsulate the MPI stuff. Also, @ilfreddy has been asking for integration tests for some time now :slightly_smiling_face:

robertodr:
Building getkw as submodule finds Boost, but can't link to it. Very funny, since it succeeds flawlessly when it's compiled on its own... I can't see any outstanding differences between the two `.travis.yml` files :crying_cat_face:

robertodr:
It actually looks like getkw gets built with CMake 3.9.2 when on its own, even though I am installing 3.5.1 from APT :thinking:

robertodr:
So, **not** installing CMake in the APT section of `.travis.yml` forces the use of CMake 3.9.2 which is magically already installed on the Travis worker. I'll need to investigate in getkw whether the construct `Boost::boost` is OK for CMake 3.3

The unit tests now run, but give me:
```
*** Error in `/home/travis/build/MRChemSoft/mrchem/build/MRChem_core-prefix/src/MRChem_core-build/tests/mrchem-tests': double free or corruption (out): 0x0000000001d9b4e0 ***
```

robertodr:
This PR graduated to irrelevance...

Issue #71, opened 2018-03-06
DFT open shell input


ilfreddy:
The MRChem input requires an explicit variable to use a spin-sparated functional. For open-shell systems, a separated functional should be the default, whereas the spin-restricted one might be activated in special circumstances (e.g. test against NIST results).

Default values in input handling shall be changed.

susilehtola:
So how is one supposed to run spin-polarized DFT?

The manual at http://mrchem.readthedocs.io/en/latest/mrchem_manual.html is outdated since it appears the proper keyword is not ```spin_polarized``` but ```spin```. However, running with ```spin = true``` results in an error saying that spin DFT is not implemented...

stigrj:
Unfortunately, spin-polarized DFT is not available in the current version, but will be very soon in the upcoming version.

susilehtola:
So is the current DFT spin-restricted?

stigrj:
Yes, only spin-restricted DFT, but both RHF and UHF. Also, note that the initial guess is not very good in the current version, and only `initial_guess = none` is possible unless you have precomputed MO coefficients and a corresponding LSDalton basis set file. This means that there is a fair chance that your computed energy is *not* going to be the global minimum. This will be much improved in the upcoming version :)

stigrj:
Well, you will find a solution, but it might not be the global minimum :)

No, I have to apologize for the current condition of the code. We had a very old version that was able to do all this properly (including linear response), but we have rewritten the code in several stages since then and not yet caught up with ourselves. Also, it was dependent on a locally hacked version of LSDalton to provide a proper initial guess, and this was never intended for distribution. We are closing in on a proper v1.0 now, though, that should be able to do unrestricted ground state calculations on arbitrary systems. This should be available within a few weeks :crossed_fingers:

stigrj:
Why are the issues not automagically closed through the commits?

PR #102, opened 2018-05-12
DFT input

stigrj:
* Automatically set the ``spin`` keyword of the ``DFT`` input section based on the value of ``restricted`` in the ``WaveFunction`` section. This default is overwritten if ``spin`` is set explicitly.
* Allow open-shell calculations using a spin-restricted functional.
* Update the keyword list of the ``DFT`` input section in the MRChem manual.

Spin-restricted open-shell has been tested against NIST for the lithium atom.

This patch closes issue #71.


ilfreddy:
Looks good. Once the check passes it can be merged&closed.

PR #103, opened 2018-05-22
Reimplementing external field

ilfreddy:
The external potential is reintroduced as a simple additional operator in the Fock Operator.

- [x] New input section
- [x] Add external potential to `FockOperator`
- [x] Test

### DIFF
--- a/src/SCFDriver.cpp
+++ b/src/SCFDriver.cpp
@@ -381,7 +381,7 @@ void SCFDriver::setup() {
     fock = new FockOperator(T, V);
     // For Hartree, HF and DFT we need the coulomb part
     if (wf_method == "Hartree" or wf_method == "HF" or wf_method == "DFT") {
-        J = new CoulombOperator(*P, *phi);
+        J = new CoulombOperator(P, phi);
         fock->setCoulombOperator(J);
     }
     // For HF we need the full HF exchange
@@ -397,7 +397,7 @@ void SCFDriver::setup() {
     //For DFT we need the XC operator
     if (wf_method == "DFT") {
         mrcpp::DerivativeOperator<3> * der_dft = 0;
-        xcfun = new XCFunctional(dft_spin, dft_explicit_der, dft_cutoff, *phi, useDerivative(diff_dft));
+        xcfun = new mrdft::XCFunctional(*MRA, dft_spin); //, dft_use_gamma, dft_cutoff, *phi, useDerivative(diff_dft));
         for (int i = 0; i < dft_func_names.size(); i++) {
             xcfun->setFunctional(dft_func_names[i], dft_func_coefs[i]);
         }
--- a/src/SCFDriver.cpp
+++ b/src/SCFDriver.cpp
@@ -396,12 +396,7 @@ void SCFDriver::setup() {
     }
     //For DFT we need the XC operator
     if (wf_method == "DFT") {
-        mrcpp::DerivativeOperator<3> * der_dft = 0;
-        xcfun = new mrdft::XCFunctional(*MRA, dft_spin); //, dft_use_gamma, dft_cutoff, *phi, useDerivative(diff_dft));
-        for (int i = 0; i < dft_func_names.size(); i++) {
-            xcfun->setFunctional(dft_func_names[i], dft_func_coefs[i]);
-        }
-        std::cout << "and here 0" << std::endl;
+        xcfun = setupFunctional(1);
         XC = new XCOperator(xcfun, phi);
         fock->setXCOperator(XC);
     }
@@ -915,4 +910,23 @@ void SCFDriver::setupInitialGrid(mrdft::XCFunctional &func, const Molecule &mol)
     Printer::printFooter(0, timer, 2);
 }
 
+    /** @brief helper routine to set up the correct parameters in the functional before using it
+     *
+     * param[in] order the requested order of the derivative (order=1 for SCF)
+     *
+     */
+
+mrdft::XCFunctional* SCFDriver::setupFunctional(int order) {
+    mrdft::XCFunctional* fun = new mrdft::XCFunctional(*MRA, dft_spin);
+    for (int i = 0; i < dft_func_names.size(); i++) {
+        fun->setFunctional(dft_func_names[i], dft_func_coefs[i]);
+    }
+    fun->setUseGamma(dft_use_gamma);
+    fun->setDensityCutoff(dft_cutoff);
+    fun->evalSetup(order);
+    setupInitialGrid(*fun, *molecule);
+    return fun;
+}
+    
 } //namespace mrchem
+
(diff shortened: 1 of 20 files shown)


src/qmoperators/one_electron/ElectricFieldOperator.h
```diff
@@ -0,0 +1,36 @@
...
+ * moment. The operator is simply implemented as a scalar product of
+ * the dipole moment operator with the electric field vector.
+ *
+ */
+
+namespace mrchem {
+
+class ElectricFieldOperator final : public RankZeroTensorOperator {
+public:
+ ElectricFieldOperator(Eigen::Vector3D f)
```
stigrj: This argument should be a `(const Eigen::Vector3d &f)`. We should probably make such a typedef for a 3D coordinate to use in generar, currently we are using `const double *r`, which is a bit primitive...

src/qmoperators/two_electron/FockOperator.cpp
```diff
@@ -139,13 +157,15 @@ SCFEnergy FockOperator::trace(OrbitalVector &Phi, const ComplexMatrix &F) {
...
     if (this->xc   != 0) E_xc  =  this->xc->getEnergy();
     if (this->xc   != 0) E_xc2 =  this->xc->trace(Phi).real();
+    if (this->ext  != 0) E_ext =  this->ext->trace(Phi).real();
 
     double E_eex    = E_ee  + E_x;
     double E_orbxc2 = E_orb - E_xc2;
-    E_kin = E_orbxc2 - 2.0*E_eex - E_en;
-    E_el  = E_orbxc2 -     E_eex + E_xc;
+    E_kin = E_orbxc2 - 2.0*E_eex - E_en - E_ext;
+    E_el  = E_orbxc2 -     E_eex + E_xc + E_ext;
```
stigrj: Should `E_ext` be added here? It should already be present in `E_orb`

ilfreddy: No it should not. I have removed it after that commit, but still it did not help. The result changes but it does not seem correct anyway :-(

stigrj: Hmm, how far off?

ilfreddy: Too far. Dipole: 1.11au. Finite Difference (E(+)-E(-))/2h: 0.84au
And it is origin dependent. This are the values in the centre of charge such that the nuclear contribution is zero.

src/qmoperators/two_electron/FockOperator.cpp
```diff
@@ -99,18 +109,21 @@ void FockOperator::rotate(const ComplexMatrix &U) {
...
+    double E_xc2 = 0.0; // Trace of the XC operator
+    double E_ext = 0.0; // External field contribution to the electronic energy
+    double E_nex = 0.0; // External field contribution to the nuclear energy
 
     // Nuclear part
     if (this->nuc != 0) {
+        std::cout << "Computing nuc conts " << ext << std::endl;
         Nuclei &nucs = this->nuc->getNuclei();
         int nNucs = nucs.size();
         for (int i = 0; i < nNucs; i++) {
```
stigrj: Can you try to move this part into a `trace(nucs)` function in `NuclearOperator`?

ilfreddy: If you mean the comment, that as just a debugging thing and can be removed. If you mean the lines under the `if`statement, I will have a look.

stigrj: The latter :slightly_smiling_face:

src/qmoperators/one_electron/NuclearOperator.cpp
```diff
@@ -59,4 +59,28 @@ void NuclearPotential::clear() {
     clearApplyPrec(); // apply_prec = -1
 }
 
+    /** @brief computes the repulsion self energy of a set of nuclei
+     * 
+     * @param[in] nucs the set of nuclei
+     *
+     */
+    
+double NuclearOperator::trace(const Nuclei &nucs) {
```
stigrj: This function should rather compute the repulsion between `*this` nuclear potential and the provided charge distribution (in this case a collection of point charges), e.i. `nuc_i` should be taken from `this->nuclei` and `nuc_j` from the `nucs` input. However, then the second loop should _not_ be truncated, which results in double counting including self-repulsion whenever `this->nuclei` coincides with the input `nucs`. How do we make this generally applicable?

src/qmoperators/one_electron/NuclearOperator.cpp
```diff
@@ -59,4 +60,27 @@ void NuclearPotential::clear() {
     clearApplyPrec(); // apply_prec = -1
 }
 
+/** @brief computes the repulsion self energy of a set of nuclei
+ * 
+ * @param[in] nucs the set of nuclei
+ *
+ */    
+double NuclearOperator::trace(const Nuclei &nucs) {
```
stigrj: This function should rather compute the repulsion between `*this` nuclear potential and the provided charge distribution (in this case a collection of point charges), e.i. `nuc_i` should be taken from `this->nuclei` and `nuc_j` from the `nucs` input. However, then the second loop should _not_ be truncated, which results in double counting including self-repulsion whenever `this->nuclei` coincides with the input `nucs`. How do we make this generally applicable?

ilfreddy: I think there should then be two functions. One for the self-repulsion interaction (not sure it should be part of the NuclearOperator either... and one for the interaction of NuclearOperator with a second set of charge (no double-counting here).

ilfreddy:
@stigrj do you like it better with a chem_utils file which computes the nuclear self-repulsion?

stigrj:
You need to rebase and resolve some conflicts before we can proceed

ilfreddy:
I'm on it now...

ilfreddy:
This can be merged in now.

src/SCFDriver.cpp
```diff
@@ -352,42 +375,58 @@ void SCFDriver::setup() {
...
-        xcfun->setDensityCutoff(dft_cutoff);
+        fock->setExchangeOperator(K);
+    }
+    //For hybrid DFT we need a partial HF exchange
+    if (wf_method == "DFT" and (dft_x_fac > mrcpp::MachineZero)) {
+        K = new ExchangeOperator(*P, *phi, dft_x_fac);
+        fock->setExchangeOperator(K);
+    }
+    //For DFT we need the XC operator
+    if (wf_method == "DFT") {
```
stigrj: Please clean up this section (remove `der_dft`, print statement and comment). Also, the XCFunctional is not set up correctly, you need the `setUseGamma()` `setDensityCutoff()` `evalSetup(1)` and `setupInitialGrid()`.

ilfreddy: OK, but why did you remove those from the constructor?

stigrj: Yes, `gamma` and `cutoff` could have been in the constructor, but we anyway have to actively set the functionals after construction, so I decided to include only the necessary (`const`) parameters in the constructor arguments

ilfreddy:
Fixed and checked that He (BLYP) and Li (PBE) give reasonable results.

PR #114, opened 2018-06-27
[WIP] New mpi

stigrj:


stigrj:
Yes, you are right. We should first loop through the orbitals and figure out which to send and which to receive, and then loop through MPI ranks following your algorithm and do the actual communication.

gitpeterwind:
The simpler way to to the MPI "pairing" in the simplest case, is to define:

	int other_rank = (max_rank + this->iter - my_rank)%max_rank;

Then the send and receive ranks are always equal, and the communication is more effective.
(note that the case other_rank=my_rank is not at the same iteration for all)
I have also a version which makes the maps of who owns what (not sure it is needed).
I will try to generalize this, so that it can handle symmetry and only send one orbital at a time. We need all this for the Exchange potential.

PR #115, opened 2018-08-06
New Density class to handle complex functions (for response calculations)

ilfreddy:
The current Density is a simple typedef and it is not flexible enough to handle perturbed densites which could also be complex-valued. We re-implement a density class which allows for more flexibility.

- [x] copy the Orbital class
- [x]  cleanup of unnecessary functionality
- [x]  parsing of the code to exchange the old typedef with the new class
- [x]  testing


ilfreddy:
@stigrj could you have a look at this? I am running a He calculation (I will attach the input file) which crashes in `compute_density`, between `wtf2` and `wtf3` ;-) 
As far as I can see all necessary objects are properly allocated. But I'm for sure overlooking something here.

ilfreddy:
Here is the input file....

# vim:syntax=sh:

rel_prec = 1.0e-4                 # Overall relative precision

MRA {
  min_scale = -5                  # Root scale defines world size 2^{-scale}
  corner = [ -1, -1, -1 ]         # Translation of corner root box [x,y,z]
  boxes = [ 2, 2, 2 ]             # Number of root boxes [x,y,z]
}

Molecule {
$coords
He   0.0     0.0     0.0
$end
}

WaveFunction {
  method = DFT                    # Wave function method (HF or DFT)
  restricted = true
}

DFT {
spin = false
use_gamma = true
density_cutoff = 1.0e-10
$functionals
SLATERX
VWN5C
$end
}

Properties {
  scf_energy = true               # Compute ground state energy
}

SCF {
  kain = 3                        # Length of KAIN iterative history
  initial_guess = SAD_DZ
}

KineticFree {
  run = false
  property_thrs = 1.0e-6          # Convergence threshold in total energy
}

src/qmfunctions/Density.h
```diff
@@ -0,0 +1,79 @@
+#pragma once
+
+#include "QMFunction.h"
+#include "density_utils.h" //LUCA Is this kosher?
```
stigrj: Why is this necessary?

src/qmfunctions/density_utils.cpp
```diff
@@ -0,0 +1,100 @@
...
+        mrcpp::add(-1.0, rho.real(), dens_vec, 0);
+    }
+    mrcpp::clear(dens_vec, true);
+
+    mpi::reduce_density(rho, mpi::comm_orb);
+    mpi::broadcast_density(rho, mpi::comm_orb);
+}
+
+void density::project(double prec, Density &rho, mrcpp::GaussExp<3> &dens_exp, int spin) {
+    FunctionTree<3> dens(*MRA);
```
stigrj: This needs to be a pointer, otherwise it is deallocated by the end of this routine

src/qmoperators/two_electron/XCPotential.cpp
```diff
@@ -60,24 +64,32 @@ void XCPotential::clear() {
 void XCPotential::setupDensity() {
     if (this->functional->hasDensity()) return;
     if (this->orbitals == nullptr) MSG_ERROR("Orbitals not initialized");
-
     OrbitalVector &Phi = *this->orbitals;
+    std::cout << "wtf" << std::endl;
     if (this->functional->isSpinSeparated()) {
         Timer time_a;
-        Density &rho_a = this->functional->getDensity(mrdft::DensityType::Alpha);
+        FunctionTree<3> tmp_a = this->functional->getDensity(mrdft::DensityType::Alpha);
```
stigrj: This must be a reference `&tmp_a`, otherwise the copy constructor is invoked

src/qmoperators/two_electron/XCPotential.cpp
```diff
@@ -60,24 +64,32 @@ void XCPotential::clear() {
...
+        Density rho_a;
+        rho_a.setReal(&tmp_a);
         density::compute(-1.0, rho_a, Phi, DENSITY::Alpha);
         time_a.stop();
         Printer::printTree(0, "XC alpha density", rho_a.getNNodes(), time_a.getWallTime());
 
         Timer time_b;
-        Density &rho_b = this->functional->getDensity(mrdft::DensityType::Beta);
+        FunctionTree<3> tmp_b = this->functional->getDensity(mrdft::DensityType::Beta);
```
stigrj: `&tmp_b`

src/qmoperators/two_electron/XCPotential.cpp
```diff
@@ -60,24 +64,32 @@ void XCPotential::clear() {
...
+        Density rho_b;
+        rho_b.setReal(&tmp_b);
         density::compute(-1.0, rho_b, Phi, DENSITY::Beta);
         time_b.stop();
         Printer::printTree(0, "XC beta density", rho_b.getNNodes(), time_b.getWallTime());
     } else {
         Timer time_t;
-        Density &rho_t = this->functional->getDensity(mrdft::DensityType::Total);
+        FunctionTree<3> tmp_t = this->functional->getDensity(mrdft::DensityType::Total);
```
stigrj: `&tmp_t`

src/qmfunctions/density_utils.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+void density::compute(double prec, Density &rho, OrbitalVector &Phi, int spin) {
+    double mult_prec = prec;            // prec for \rho_i = |\phi_i|^2
+    double add_prec = prec/Phi.size();  // prec for \sum_i \rho_i
+
+    FunctionTreeVector<3> dens_vec;
+    for (int i = 0; i < Phi.size(); i++) {
+        if (mpi::my_orb(Phi[i])) {
+            Density *rho_i = new Density(); //LUCA: Is it the right creator here (it was Density(*MRA);
+            rho_i->setReal(new mrcpp::FunctionTree<3>(*MRA));
```
stigrj: You could make an `allocReal()` function for this

src/qmfunctions/Density.h
ilfreddy: Because otherwise it does not compile ;-)
(seriously) I acknowledge this is not logically correct (the class should not depend on the utils), but something is not correct. This applies btw also to the `Orbital` class

src/qmfunctions/density_utils.cpp
ilfreddy: Yepp!

src/qmoperators/two_electron/XCPotential.cpp
ilfreddy: Indeed!
I think we need to have a discussion about references, pointers and similar stuff in one of the coming meetings!

ilfreddy:
@stigrj can you have a look at this one?

PR #116, opened 2018-09-03
Restablish some of the MPI functionalities

gitpeterwind:
- OrbitalIterator: changed algorithm, allow symmetric computations at half the price, allow send one orbital at a time when memory use is an issue (exchange). NB: the routine is short but to understand what happens in practice is not trivial (odd/even number of MPI or orbitals, send/recv to oneself,avoid situation where the process is in a deadlock, no situation where processors must wait for more than one other MPI, load balancing, etc.)
- Parallelized exchange. Uses symmetri, send one exchange-orbital at a time, (but rotation of exchange still use vectors)
- Use symmetry for overlap matrix when possible
- Parallelized load and save orbitals (did not work for MPI)
- Revised localization of orbitals. Changed algorithm for better scaling. Only use diagonal of the Hessian first, then in the minimum region use a few iterations of Conjugated Gradient (with exact Hessian) and diagonal of Hessian as preconditioner.
- Bug in center of mass and center of charge
- Write out torque and total force of nuclear derivatives.
- Note: the creation of directories (very start of mrchem, before c++ code start) can fail under MPI if the directory does not exist, because many MPI can try to create the directory at the same time.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -5,4 +5,4 @@ Peter Wind        <peter.wind@uit.no>
 Magnar Bjorgve    <magnar.bjorgve@uit.no>
 Rune Monstad      <mo019@post.uit.no>
 Eirik Fossgaard
-Antione Durdek
+Antoine Durdek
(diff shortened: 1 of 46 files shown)


stigrj:
This has been fully accounted for in #122, #123 and #124.

PR #117, opened 2018-09-05
Cleanup includes and forward declarations

stigrj:
I changed the typedef
```
typedef std::vector<Orbital> OrbitalVector;
```
into an empty derived class
```
class OrbitalVector final : public std::vector<Orbital> { };
```
This was the only way I could find that allows for forward declaration of `OrbitalVector`.


src/qmfunctions/orbital_utils.h
```diff
@@ -4,9 +4,12 @@
...
 namespace mrchem {
 
+/* The following container classes are defined as derived classes rather than
+ * typedefs in order to be able to forward declare them. */
+
 class Orbital;
-typedef std::vector<Orbital> OrbitalVector;
-typedef std::vector<std::tuple<int, Orbital> > OrbitalChunk;
+class OrbitalChunk final : public std::vector<std::tuple<int, Orbital> > { };
+class OrbitalVector final : public std::vector<Orbital> { };
```
robertodr: I think this is fine as long as you are OK with always using the default allocator for a `std::vector`.

src/qmfunctions/density_utils.h
```diff
@@ -4,13 +4,21 @@
 
 namespace mrchem {
 
+/* The following container classes are defined as derived classes rather than
+ * typedefs in order to be able to forward declare them. */
+
+class Orbital;
+class OrbitalVector;
+class Density;
+class DensityChunk final : public std::vector<std::tuple<int, Density> > { };
```
robertodr: In C++11 you don't need to separate multiple `>`  when closing a template declaration. That is, this is perfectly OK:
`std::vector<std::tuple<int, Density>>`

stigrj: Really? Well that's a relief :slightly_smiling_face:

stigrj:
Yes, the typedef header is a definite possibility

stigrj:
Made separate `typedef` headers for the different directories:
- `chemistry/chemistry.h`
- `properties/properties.h`
- `qmfunctions/qmfunctions.h`
- `qmoperators/qmoperators.h`

This avoided the empty derived classes.

Changed to `using` syntax rather than `typedef`.

Feel free to squash and merge if approved.

PR #122, opened 2018-09-06
Geometric derivative

stigrj:
Squashed the geometric derivative part of Peters PR #116


robertodr:
I think it needs a quick rebase.

PR #123, opened 2018-09-06
Improved MPI functionalities

stigrj:
Squashed the MPI part of Peter's PR #116.

Fixed the merge conflicts and the bug we discovered at the group meeting, and the code now runs successfully in parallel.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,16 @@ jobs:
       - *configure-omp
       - *build
       - *tests
+  mpi-py3:
+    <<: *ubuntu-1804
+    environment:
+      - PYTHON_VER: "3"
+    steps:
+      - checkout
+      - *setup
+      - *report
+      - *configure-mpi
+      - *build
 
 workflows:
   version: 2
@@ -100,3 +110,4 @@ workflows:
     jobs:
       - serial-py3
       - omp-py3
+      - mpi-py3
--- a/src/SCFDriver.cpp
+++ b/src/SCFDriver.cpp
@@ -398,12 +398,12 @@ void SCFDriver::setup() {
     }
     // For HF we need the full HF exchange
     if (wf_method == "HF") {
-        K = new ExchangeOperator(*P, *phi);
+        K = new ExchangeOperator(P, phi);
         fock->setExchangeOperator(K);
     }
     //For hybrid DFT we need a partial HF exchange
     if (wf_method == "DFT" and (dft_x_fac > mrcpp::MachineZero)) {
-        K = new ExchangeOperator(*P, *phi, dft_x_fac);
+        K = new ExchangeOperator(P, phi, dft_x_fac);
         fock->setExchangeOperator(K);
     }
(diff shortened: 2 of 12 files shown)


stigrj:
@gitpeterwind can you run this version on some of your test cases and see if everything looks ok? Note that I have done *a lot* of meddling with the git history, so please don't try to merge back into some of your versions.

robertodr:
@stigrj I put an embryo of MPI set up in `.circleci/config.yml`. It can be activated putting this stanza under `jobs`:
```yaml
  mpi-py3:
    <<: *ubuntu-1804
    environment:
      - PYTHON_VER: "3"
    steps:
      - checkout
      - *setup
      - *report
      - *configure-mpi
      - *build
      - *tests
```
and then ilsting it under the `workflows`. The tests will run serial though.

gitpeterwind:
When I tried to compile it, I got this error:
``/home/mifapw/forkmrchem/mrchem/src/SCFDriver.cpp(9): catastrophic error: cannot open source file "getkw/Getkw.hpp"
  #include "getkw/Getkw.hpp"``
I might have an incomplete version of something. When/where was getkw/Getkw.hpp introduced?

stigrj:
Then you have some old headers in the build directory. Try to do a `git submodule sync` and a new `./setup`

gitpeterwind:
Thanks, that worked (although it made the older branch not work anymore).
The results are not identical (but the same within required precision). I think it is because in the squash-mpi the symmetrical treatment of the overlap matrix is not included.

stigrj:
Is the symmetric overlap missing in this PR relative to your original?

gitpeterwind:
yes, it is included in #116 but not in #123 (in qmfunctions.cpp)

stigrj:
No, the symmetry is there, but in `orbital_utils.cpp`, and the `symmetric=true` is moved from `iter.next(true)` into the iterator constructor. This is because the symmetry property is not likely to change during an iteration procedure. Actually, I think I will move the `max_recv` parameter back into `next()` since this one could conceivably change. Also, I think I will move the symmetric iteration explicitly into `calc_overlap(Phi)` and let the `calc_overlap(Phi_a, Phi_b)` remain unsymmetric even if `Phi_a == Phi_b`.

gitpeterwind:
Then I do not know what can cause the difference. Do you know other things that might have changed between the versions?

stigrj:
I included the MPI build on CircleCI, but without the tests since there is no MPI initialization in the current test cases. However, it is nice to confirm that the code at least builds with MPI, even without the tests.

stigrj:
I will do some more test calculations next week to convince myself that the current version is correct. Hopefully you don't have a lot of uncommitted changes on top of PR #116, @gitpeterwind . Otherwise we might have to manually import them.

gitpeterwind:
The discrepancies between the  #116 and #123 were due to differences in the localization scheme: when the new scheme is included the results get identical.
The new localization scheme is not yet included in #123. It should be included later then (it scales better and use less memory).

PR #125, opened 2018-09-07
Energy optimizer

stigrj:
The helium example crashed in the second SCF procedure (energy optimization) due to a faulty initialization of the n+1 FockOperator. Also fixed a memleak in CoulombPotential.


ilfreddy:
I guess I should have just approved it :-)

PR #126, opened 2018-09-07
Add infrastructure for integration testing

robertodr:
Hooks up with [`runtest`](https://github.com/bast/runtest) for integration testing.

~**Warning** Getkw doesn't ship anymore a copy of `pyparsing`, so running `mrchem` might fail if that's not somewhere on your system.~
I've updated to latest Getkw which ships `pyparsing.py`.


stigrj:
I have some runtime problems I have to resolve before I can approve this PR, probably related to python 2vs3. The helium integration test runs successfully, but I'm not able to run the other examples manually.

robertodr:
What problems do you see? The set up for the tests is likely to change soon, but I wanted to bring this is in sooner rather than later, to have a baseline of working functionality.

stigrj:
```
(venv) stig@pilates :: mrchem-stig/install-python3/bin ‹superbuild-mrchem*› » python2 ./mrchem                                                                                     1 ↵
Traceback (most recent call last):
  File "./mrchem", line 547, in <module>
    main()
  File "./mrchem", line 46, in main
    inpfil = os.path.join(root_dir, inpfil)
  File "/home/stig/src/mrchem-stig/venv/lib/python2.7/posixpath.py", line 68, in join
    if b.startswith('/'):
AttributeError: 'NoneType' object has no attribute 'startswith'
```
```
(venv) stig@pilates :: mrchem-stig/install-python3/bin ‹superbuild-mrchem*› » python3 ./mrchem                                                                                     1 ↵
Traceback (most recent call last):
  File "./mrchem", line 547, in <module>
    main()
  File "./mrchem", line 46, in main
    inpfil = os.path.join(root_dir, inpfil)
  File "/usr/lib/python3.6/posixpath.py", line 94, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib/python3.6/genericpath.py", line 149, in _check_arg_types
    (funcname, s.__class__.__name__)) from None
TypeError: join() argument must be str or bytes, not 'NoneType'
```

robertodr:
Is it supposed to work without passing positional arguments? `os.path.join` gets confused when generating the full path to the input file because the second argument to the function call is empty.

stigrj:
Aha, we used to have `mrchem.inp` as default, now it runs if I pass this as argument. Thanks!

robertodr:
Should I revert to having a default? Or should I put a more intelligible error higher up, right after command-line parsing?

stigrj:
It's probably better with explicit argument, but an intelligible error would be nice :slightly_smiling_face:

stigrj:
There's something weird with the `printlevel` input, though. It defaults to zero in the `mrchem` script, but the output is as if `printlevel=1`, i.e. a bunch of lines like this:
```
 Applied QMMomentum                    456      3.93566e-02
 Applied QMMomentum                    584      4.51354e-02
 Applied QMMomentum                    456      3.32929e-02
 Applied QMMomentum                    584      4.44285e-02
 Applied QMMomentum                    456      3.44789e-02
 Applied QMMomentum                    584      4.67404e-02
 Applied QM potential                  328      7.81547e-02
 Applied QM potential                  328      8.24981e-02
```
When I explicitly set `printlevel=0` in `mrchem.inp` the output is as expected.

robertodr:
Fixes incoming. Anything else you'd like to see changed?

robertodr:
The logic to set the `printlevel` is:
```python
    printlevel = int(opts.printlevel)

    debug = opts.debug
    if debug:
        printlevel = 1
```
and `debug` defaults to `MRCHEM_DRIVER_DEBUG`. So I guess you're building the debug configuration? I think I've followed what seemed to be the existing logic in refactoring the script, or did I mess it up?

stigrj:
Yes, these lines seems to trigger it, but I don't understand how. My `MRCHEM_DRIVER_DEBUG' is OFF and `debug` defaults to `false`. Not sure if we need the `debug` option, though? Currently it doesn't do anything other than set `printlevel=1`, which can anyway be done with `--print=1` (and printlevel 1 doesn't provide much debug information anyway).

stigrj:
Hmm, now it tries to build mrchem immediately without any of the external projects built :confused:

stigrj:
A dot appears in the install path:
```
Install the project...
-- Install configuration: "release"
-- Installing: /home/stig/src/mrchem-stig/install-super/.
-- Installing: /home/stig/src/mrchem-stig/install-super/./include
-- Installing: /home/stig/src/mrchem-stig/install-super/./include/XCFun
-- Installing: /home/stig/src/mrchem-stig/install-super/./include/XCFun/functional.hpp
-- Installing: /home/stig/src/mrchem-stig/install-super/./include/XCFun/densvars.hpp
-- Installing: /home/stig/src/mrchem-stig/install-super/./include/XCFun/xcfun.h
```

robertodr:
The `.` is intended. It's a bit ugly, but it's the only way to tell CMake to install at the root of the install prefix with a relative path. For the background of the change: I was setting `CMAKE_INSTALL_PREFIX` (absolute path) instead as `DESTINATION`, but I've since learnt that using absolute paths in `install` commands is a very bad idea.

As for the use of `debug` and `printlevel`. I agree that either one would serve the purpose and I'd vote for keeping `--print`. It's surprising that the launch script gets hardcoded to a certain behavior at configuration time and it's not a very transparent configuration to override. Explicit is better than implicit.

I'll take care of dev-cafe/libgetkw#21 and then update Getkw here.

stigrj:
Note that `printlevel` is much more fine grained than the 0 vs 1 indicated by the `debug` flag, and proper debug information doesn't appear until `printlevel=10`, so the `debug` flag is currently completely useless. Can you please remove this flag altogether, along with `MRCHEM_DRIVER_DEBUG`?

robertodr:
Much obliged! Configuring execution scripts is a bit iffy, in my opinion.

stigrj:
Now just waiting for new getkw?

robertodr:
Yep

robertodr:
Make sure to run `git submodule sync` and `git submodule update --init --recursive`

stigrj:
Greatness!

PR #127, opened 2018-09-12
reduce density one mpi at a time, instead of all at once

gitpeterwind:
Reduces drastically memory usage for mpi runs

### DIFF
--- a/src/parallel.cpp
+++ b/src/parallel.cpp
@@ -46,7 +46,7 @@ void mpi::initialize(int argc, char **argv) {
 
     // divide the world into groups
     // each group has its own group communicator definition
- 
+
     // split world into groups that can share memory
     MPI_Comm_split_type(MPI_COMM_WORLD, MPI_COMM_TYPE_SHARED, 0, MPI_INFO_NULL, &comm_share);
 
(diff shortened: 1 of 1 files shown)


src/parallel.cpp
```diff
@@ -230,24 +231,15 @@ void mpi::reduce_density(Density &rho, MPI_Comm comm) {
...
-        rho_0->alloc(NUMBER::Real);
-        mrcpp::copy_grid(rho_0->real(), rho.real());
-        mrcpp::copy_func(rho_0->real(), rho.real());
-        rho.real().clear();
-
-        mrcpp::FunctionTreeVector<3> rho_vec;
-        rho_vec.push_back(std::make_tuple(1.0, &rho_0->real()));
-        for (int src = 1; src < comm_size; src++) {
-            Density *rho_i = new Density;
+	Density *rho_i = new Density;
```
stigrj: Your indentation settings are still wrong :slightly_smiling_face:

gitpeterwind: What is wrong? I have 4 spaces as default identation

gitpeterwind: It looks fine in my file. Either it is not correctly pushed, or github's diff does not show properly

stigrj: The lines you have changed seems to have one tab + 4 spaces instead of 12 spaces. Perhaps 8 spaces automatically turns into a tab?

gitpeterwind: yes, it seems there is some mixing of spaces and tabs that make the problems. Emacs puts tabs I think.

PR #129, opened 2018-09-13
Include headers relatively

robertodr:
Header includes are now done relative to `src`. This simplifies carrying around the include directories with `target_include_directories`.


src/chemistry/Molecule.h
```diff
@@ -27,6 +26,16 @@
 
 namespace mrchem {
 
+class DipoleMoment;
```
stigrj: `properties/properties.h` could not be used for the forward declarations?

robertodr: I picked this up from the #96 so probably `properties.h` didn't exist yet in that "universe"? Should I put back `properties.h`? Slightly off-topic suggestion: header files which only contain forward declaration could maybe carry a `fwd` prefix or suffix. I was convinced `properties.h` did a blanket-include :smile_cat:

PR #130, opened 2018-09-14
Formatting, licensing, ang Git hooks

robertodr:
- [x] Propose a `.clang-format` file. @MRChemSoft/mrchemdev I have formatted `src/scf_solver` Please look at the diff to check whether you like it or not. If not, let me know what you'd like changed. I ran the formatting with: `cd src/scf_solver; git ls-files -- '*.cpp' '*.h' | xargs clang-format -i`
- [x] Add (optional) Git pre-commit hooks to check that licensing headers and formatting are correct.

To add the license header on top of the source files:
```
./.githooks/pre-commit-license-maintainer
```
I am not doing since it would touch all files and I think it's better @stigrj does something like that.

### DIFF
--- a/src/scf_solver/Accelerator.cpp
+++ b/src/scf_solver/Accelerator.cpp
@@ -20,9 +20,9 @@ extern mrcpp::MultiResolutionAnalysis<3> *MRA; // Global MRA
  * @param sep: solve separate eqations for each orbital
  */
 Accelerator::Accelerator(int max, int min, bool sep)
-        : minHistory(min),
-          maxHistory(max),
-          sepOrbitals(sep) {
+        : minHistory(min)
+        , maxHistory(max)
+        , sepOrbitals(sep) {
     if (this->minHistory < 0) this->minHistory = 0;
     if (this->maxHistory < this->minHistory) MSG_ERROR("Invalid argument");
 }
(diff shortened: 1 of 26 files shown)


src/scf_solver/Accelerator.cpp
```diff
@@ -41,12 +41,8 @@ void Accelerator::clear() {
         orbital::free(this->dOrbitals[0]);
         this->dOrbitals.pop_front();
     }
-    while (this->fock.size() > 0) {
-        this->fock.pop_front();
-    }
-    while (this->dFock.size() > 0) {
-        this->dFock.pop_front();
-    }
+    while (this->fock.size() > 0) { this->fock.pop_front(); }
```
stigrj: For one-liners I like to skip the `{` `}`, but it's probably not for the formatter to remove those?

stigrj:
Looks very good to me. What's the condition for line break? Some of the argument lists look awfully long

src/scf_solver/Accelerator.cpp
robertodr: Let me see if I can convince the formatter to not put braces for single-line constructs.

robertodr:
The breaks are computed based on the column number. This is now set to 120, trying with 85 didn't help that much. And less than 85 seems silly.

stigrj:
Ok, just keep 120 then

src/scf_solver/Accelerator.cpp
robertodr: Can't manage to convince to remove braces...

robertodr:
I've unmade some of the argument list changes by hand. These are now protected with `// clang-format off` and `// clang-format on`

stigrj:
Hmm, I think I prefer the long arguments over the extra comments :stuck_out_tongue:

src/scf_solver/Accelerator.cpp
stigrj: That's fine, it doesn't seem to enforce them either so we just have to avoid writing them :slightly_smiling_face:

stigrj:
So what would you like me to do next?

robertodr:
This one can be merged. You can then apply license header repo-wide and that shouldn't cause too many conflicts. Formatting might generate a lot of conflicts for the rest of the team, so that might have to be discussed more carefully.

PR #132, opened 2018-09-19
Kinetic matrix MPI

stigrj:
Kinetic matrix
-----------------
The kinetic part of the Fock matrix is now computed as:
<&phi;<sub>i</sub> | -&nabla;<sup>2</sup> | &phi;<sub>j</sub>> = <&nabla;&phi;<sub>i</sub> | &nabla;&phi;<sub>j</sub>>

The matrix is computes as:
```
OrbitalVector Phi_x = p_x(Phi);
ComplexMatrix T_x = orbital::calc_overlap_matrix(Phi_x);
```
which means that the momentum operator is first applied locally on each MPI (only own orbitals, no communication), then the overlap matrix is computed in the usual distributed way, including symmetry.

Coulomb potential
-----------------------
I also increased the precision (scaled by number of electrons) of the application of the Poisson operator in the CoulombPotential. For larger systems the Coulomb potential became way too small using the regular relative precision, typically ~200 nodes for the potential while the density could be ~100 000 nodes and the nuclear potential ~30 000 nodes. The proper precision parameter needs to be fine tuned at a later stage in order to be consistent for all system sizes and target precisions.


stigrj:
Integration test (He atom SCF) fails because the Coulomb potential is slightly different, giving an energy difference of ~1.0e-6 Ha. We need to adjust the success criterion.

robertodr:
The filters are defined [here](https://github.com/MRChemSoft/mrchem/blob/master/tests/he/test#L14-L27) You want to change the `rel_tolerance` parameter to something more lenient. I guess 10^-5?

stigrj:
Actually, I think we should keep a strict threshold for the integration tests to detect any changes that might influence the final precision in the calculation, even if the diff is well below the requested precision. If we decide that such a change is still acceptable, then the reference should be updated along with the code changes in the PR.

PR #134, opened 2018-09-20
License header

stigrj:
Add license header to all source files


ilfreddy:
Fine by me, but I cannot write a review for now. I guess I need to wait for the checks to pass...

stigrj:
Doh, we need to merge the new integration tests before this is accepted by Travis

robertodr:
And as I wrote in the `README.md`, if you install the Git hooks then every commit will be checked for the presence of the license header. If the header is present, the hook will check that year is correct.

PR #136, opened 2018-09-20
Fix faulty link to documentation page

stigrj:


stigrj:
Ah, yes it also worked to change the default from `master` to `latest`. Scrapping this PR

PR #137, opened 2018-09-25
Activate MPI test case on CircleCI

stigrj:
Unit tests
------------
- Added `MPI_Init` and `MPI_Finalize` so that the tests run without failing, but all test cases are fully sequential

Integration tests
--------------------
- Update `runtest` to allow for `mpirun` launch agent
- The test cases are run with a fixed number of MPI processes when compiled with MPI options
- Relaxed the success criterion for the properties, since we currently cannot expect identical results when running with different number of MPI procs

Not at all sure if this is a sensible way of passing arguments from CMake -> CTest -> runtest

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -110,6 +110,7 @@ jobs:
       - *report
       - *configure-mpi
       - *build
+      - *tests
 
 workflows:
   version: 2
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,7 @@ Profiling*
 build*
 debug*
 release*
+install*
 intel*
 pilot
 mrcpp.config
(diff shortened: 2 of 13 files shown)


stigrj:
Yes, I was wondering about that, but it was @robertodr who included runtest in this way.

bast:
In (LS)Dalton and DIRAC it is done this way since there is no `requirements.txt` ecosystem/culture yet but here I would go for PyPI.

robertodr:
Nice! As for runtest, I agree with @bast that it is much better to track from PyPI. But this needs proper documentation or a CMake check whether the module is available before making the integration tests available to `ctest`

stigrj:
Ok, so you think it is fine to keep it like this for the time being?

robertodr:
For the moment, yes. If you open an issue we can decide a bit better what to do. The change would mean a slight change of workflow for everybody (get a virtual environment and activate it before doing any MRChem work) which is minimal, but might discourage running these tests. More than an issue of coding it is and issue of tooling, documentation, and culture.

Issue #138, opened 2018-09-25
install runtest through PyPI


stigrj:

robertodr:
See #137 for context. The sources for `runtest` are currently part of MRChem. @bast suggested adding a `runtest` to `requirements.txt` to install it from PyPI instead. 
We have four alternatives:
1. Leave things as they are. Not very clean, in my opinion (and I was the one to do it this way in the first place)
2. The `requirements.txt` way.
3. Skip integration tests if `runtest` is not available.
4. Install globally.
In all cases there is need for documentation. The other three need some changes to tooling and culture.

#### Approach 2
Requires some changes in the workflow:
1. Create a Python virtual environment
2. Activate it
3. Install required Python dependencies

This is not terrible, but forgetting to activate the environment might result in surprising test failures. And it is an additional step.
The step can be automatized using something like [`direnv`](https://direnv.net/) so that anytime you do `cd mrchem` the environment gets activated. Of course, this is problematic when working on a cluster, where direnv is not installed (it's not hard to install it though, I have it on fram)

#### Approach 3
The alternative is to _only_ activate integration tests _if_ `runtest` is available. Code-wise this amounts to a CMake check. Drawback (in my opinion a **huge one**) is that integration tests suddenly are only run during CI.

#### Approach 4
Yet another alternative, which I do not recommend, is to install `runtest` globally. Again, this probably won't work on a cluster.

**My preference** is for following approach 2 (I've been using direnv for a year already and it works smooth as vodka) or keep approach 1. Either way, documentation is needed.

bast:
I like a solution which does not deactivate tests but rather screams that I forgot to activate the venv. I would favor 2). Alternatively pipenv. And we need to push clusters towards these good solutions.

bast:
I agree. I like both the high-level aspect of pipenv as well as the locking and hashing. It is safer and better reproducible.

stigrj:
I agree with the second approach, although I'm not yet familiar with `pipenv` and `direnv`. We already have some requirements listed for building documentation.

PR #139, opened 2018-09-26
Remove runtest sources

robertodr:
As discussed in #138. I have updated the Pipfile (I had forgotten that's already used on CircleCI!) and removed the sources of `runtest`. I have also added `.envrc` which sets up `direnv`.

### Using _without_ `direnv`
```
cd mrchem
pipenv shell
```
This will install Python dependencies into a virtual environment under `$HOME/.local/virtualenvs` and activate it.

### Using _with_ `direnv`
1. Install `direnv`. Either from your package manager or [download the tarball](https://github.com/direnv/direnv/releases) and put it on `PATH`
2. On first use:
```
cd mrchem
direnv allow
```
3. For all subsequent uses, `cd mrchem` will activate the virtual environment automatically.

**Note** The `.envrc` has a `case` switch because I was already using it locally.


stigrj:
Thanks, this looks great! I have to test it out first on different platforms.

robertodr:
You'd need to install direnv locally and then adjust `.envrc` to run the operations necessary to load the correct environment (like loading modules and so forth)

PR #135, opened 2018-09-20
[WIP] Shared QM functions

gitpeterwind:
Shared memory for Coulomb density.
The present version is consistent, however it could be an advantage to set the "shared memory" into the density at a higher level: 
define a creator "Density(sharedmem)" corresponding to what we have in the FunctionTree.

PR #140, opened 2018-10-02
Extra functionality to the Denisty class

ilfreddy:
The density class is now augmented with functionality required for linear response calculations:
We need to be able to compute these objects:

1. |phi><x| + |y><phi|
2. |phi><x| + cc

And sums of 1 and 2 over all orbitals.

The single orbital routines are now hidden to avoid problems with the MPI implementation.

### DIFF
--- a/src/qmfunctions/density_utils.cpp
+++ b/src/qmfunctions/density_utils.cpp
@@ -81,6 +81,45 @@ void density::compute(double prec, Density &rho, Orbital phi, int spin) {
     mrcpp::clear(sum_vec, true);
 }
 
+void density::compute(double prec, Density &rho, Orbital phi, Orbital xi, int spin) {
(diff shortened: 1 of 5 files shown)


stigrj:
The open-shell integration test fails with MPI. I will look into it

stigrj:
You are trying to move MRCPP beck to an older commit than what is currently on `master`

ilfreddy:
> You are trying to move MRCPP back to an older commit than what is currently on `master`

I see... How can I prevent that?

stigrj:
I have realized some design flaws in the whole QMFunction - Orbital - Density hierarchy and I am working on a fix. This also relates to some problems I found when looking into a generalization on shared memory MPI in #135.

ilfreddy:
I see. Let me know if I can help with anything.

stigrj:
These changes have been incorporated in #144

PR #141, opened 2018-10-02
Update Doxyfile.in

ilfreddy:
Fix path to source files in documentation build


stigrj:
Great, it works! Now we just need some documentation :thinking:

PR #144, opened 2018-10-05
Dens class

stigrj:
I restructured the `QMFunction` to make the `Orbital` and `Density` functionality simpler and more consistent. This will also help in the generalization of shared memory that should be done in #135. It's probably a bit too much for a proper review, but @ilfreddy please have a look at the changes it lead to in the `density::compute()` routines.

### DIFF
--- a/src/initial_guess/core.cpp
+++ b/src/initial_guess/core.cpp
@@ -40,6 +40,7 @@
 #include "chemistry/Nucleus.h"
 #include "qmfunctions/Orbital.h"
 #include "qmfunctions/orbital_utils.h"
+#include "qmfunctions/qmfunction_utils.h"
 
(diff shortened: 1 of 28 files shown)


stigrj:
Note that `OrbitalVector` (`std::vector<Orbital>`) is now conceptually different from `QMFunctionVector` (`std::vector<QMFunction>`). The former should be a collection of orbitals defining a proper Slater determinant, while the latter is any arbitrary collection of functions (could also be orbitals) used e.g. when adding several orbitals into one. The difference is important when MPI is activated, as `OrbitalVectors` are distributed while `QMFunctionVectors` are not.

robertodr:
Maybe add these notes as Doxygen documentation strings? :smile_cat:

PR #147, opened 2018-10-17
Write perturbed orbitals on disk

ilfreddy:
Perturbed orbitals are written on disk in the same folder as the ground state orbitals. In order to distinguish them, they bear a suffix in the name, which stems from the perturbation. The name is derived from the corresponding `ResponsePerturbation` object, which in turns gets it from the corresponding perturbation operator. The naming is for the time being done "by hand" but it would be better to assign a name directly to the operators and then use it.


stigrj:
Incorporated in #155

PR #151, opened 2018-11-02
add Cavity class

Gabrielgerez:


src/chemistry/Cavity.cpp
```diff
@@ -0,0 +1,31 @@
+#include "Cavity.h"
+#include <cmath> 
+#include <vector>
+#include <array>
+
+namespace mrchem {
+
+Cavity::Cavity(std::vector<std::vector<double>> coord, std::vector<double> R, double slope){
```
robertodr: I think either `Eigen::Matrix3Xd` or `std::vector<std::array<double, 3>>` would be more expressive containers for the coordinates of the sphere centers.

src/chemistry/Cavity.h
```diff
@@ -0,0 +1,20 @@
...
+namespace mrchem {
+
+
+class Cavity : public mrcpp::RepresentableFunction<3> {
+public: 
+  Cavity(std::vector<std::vector<double>> coord, std::vector<double> R, double slope);
+  double evalf(const double *r) const;
+  double evalf(const std::array<double, 3> &r) const {return 0;}
+protected:
+  std::vector<std::vector<double>> pos;
```
robertodr: I think either `Eigen::Matrix3Xd` or `std::vector<std::array<double, 3>>` would be more expressive containers for the coordinates of the sphere centers.

src/chemistry/Cavity.h
```diff
@@ -0,0 +1,20 @@
...
+#include "MRCPP/MWFunctions"
+
+namespace mrchem {
+
+
+class Cavity : public mrcpp::RepresentableFunction<3> {
+public: 
+  Cavity(std::vector<std::vector<double>> coord, std::vector<double> R, double slope);
+  double evalf(const double *r) const;
+  double evalf(const std::array<double, 3> &r) const {return 0;}
```
robertodr: This method should call the one taking the pointer, rather than just returning zero.

src/chemistry/Cavity.h
Gabrielgerez: changed it now

src/chemistry/Cavity.h
Gabrielgerez: I think it is fixed.

stigrj:
Actually, the `mrcpp::Coord<D>` has not made its way into MRChem yet. Should be done ASAP. This means that the old pure virtual `evalf(double *r)` is still necessary in the current version.

src/chemistry/Cavity.h
```diff
@@ -0,0 +1,20 @@
...
+#include <array>
+#include "MRCPP/MWFunctions"
+
+namespace mrchem {
+
+
+class Cavity : public mrcpp::RepresentableFunction<3> {
+public: 
+  Cavity(std::vector<std::array<double, 3>> coord, std::vector<double> R, double slope);
+  double evalf(const double *r) const;
```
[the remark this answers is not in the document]
Gabrielgerez: as @stigrj pointed out, it seems this is not yet a feature in mrchem, so i will keep it as it was for now. will make sure to add `Coord<3>` later down the line though.

src/chemistry/Cavity.h
```diff
@@ -0,0 +1,20 @@
...
+  double evalf(const std::array<double, 3> &r) const {return evalf(r.data());}
```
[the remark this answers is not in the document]
Gabrielgerez: Done

src/chemistry/Cavity.cpp
```diff
@@ -0,0 +1,31 @@
...
+  this->pos = coord;
+  this->R = R;
+  this->d = slope; 
+}
+
+
+double Cavity::evalf(const double *r) const {
+  double C = 1.0;
+  double s, O;
+  int size = pos.size();
```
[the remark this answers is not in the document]
Gabrielgerez: Done.

PR #152, opened 2018-11-05
MRCPP coord

stigrj:
Updated to latest MRCPP with the new `mrcpp::Coord<3>` type instead of `const double *`


src/chemistry/Cavity.cpp
```diff
@@ -1,30 +1,54 @@
...
-    C *= 1 - (1 - O);
-  }
-  C = 1 - C;
-  return C;
+double Cavity::evalf(const mrcpp::Coord<3> &r) const {
+    double C = 1.0;
+    double s, O;
+
+    for (int i = 0; i < pos.size(); i++) {
+        s = math_utils::calc_distance(pos[i], r);
```
Gabrielgerez: `s` has to be the signed normal distance from the atom at  `pos[i]`. meaning that it has to be the distance minus the VdW radius of the atom. Should be enough to just write `math_utils::calc_distance(pos[i], r) - R[i]`

stigrj: Thanks!

PR #153, opened 2018-11-15
Shared memory

stigrj:
Several things to consider:

1. Generalized shared memory functionality to `QMFunction`, which means that both `QMPotential`, `Density` and `Orbital` can be shared. At the same time generalized `mpi::send_orbital` and `mpi::send_density` to `mpi::send_function` (as well as `recv`).

2. In-place addition of `QMFunction` had to be changed in order to allow for shared data. Previously it was a "fake" in-place in the sense that we first added out-of-place and then reassigned to the original function. Now it's a proper in-place union grid addition. Used extensively in `density::compute()`.

3. Backtracked the DFT grid algorithm. Now the grid is properly cleared in between SCF cycles and re-grown in `XCOperator::setup(prec)` based on precision. This algorithm needs to be reconsidered in connection with linear response. Shared memory for DFT functions is NOT possible at the moment.

4. The `density::compute()` algorithm is simplified relative to Peter's original shared memory version. The shared masters no longer writes directly into the final shared memory but to a temporary local density in the same way as the non-masters. This was done mainly for readability, but also because the old setup was not compatible with some of the other changes. Should be reconsidered along with a better scaling (not all-to-one) reduction algorithm.

5. Added MPI input parameters: `numerically_exact = true` means that we guarantee invariance wrt number of MPI ranks (requires much more memory than `false`). You can also decide which of the global functions (densities and potentials) to share, e.g. `share_nuclear_potential = true`.

6. With `numerically_exact = true` the density grids are grown extensively to the union of all orbital grids before the final grid is cropped to the desired precision. This should be associative within double precision. With `numerically_exact = false` the intermediate grids are cropped along the way, which makes the final result dependent on the order of addition and number of MPI ranks. The same applies to `orbital::rotate()`.


stigrj:
Good point. No I haven't tested shared memory for actual complex functions. I will try to compose a unit test on this.

PR #154, opened 2018-11-20
added override to isZeroOnInterval and isVisibleAtScale

ilfreddy:


src/analyticfunctions/NuclearFunction.cpp
```diff
@@ -48,7 +48,7 @@ double NuclearFunction::evalf(const mrcpp::Coord<3> &r) const {
     return result;
 }
 
-bool NuclearFunction::isVisibleAtScale(int scale, int nQuadPts) const {
+bool NuclearFunction::isVisibleAtScale(int scale, int nQuadPts) const override {
```
stigrj: The `override` keyword should only be in the header file

stigrj:
Can you also add the `template class project<1>...` fix to mrcpp, and update mrcpp as part of this PR?

ilfreddy:
Yes, but I guess I cannot do that directly on github or?

stigrj:
Probably not

ilfreddy:
Good to go

stigrj:
Is mrchem/mrcpp pointing to ilfreddy/mrcpp now? I think we have to wait for the mrcpp PR to be merged first, then point to MRChemSoft/mrcpp.

ilfreddy:
Good now? :-)

stigrj:
Well, you could have cleaned up your history a bit :wink:

ilfreddy:
And you have spent too much time listening to @robertodr lately...

PR #155, opened 2018-11-20
Write RSP

stigrj:
Did a rebase on master and implemented reading of response orbitals. Had to make a new PR rather than push to ilfreddy due to the rebase...

### DIFF
--- a/src/SCFDriver.cpp
+++ b/src/SCFDriver.cpp
@@ -155,6 +155,10 @@ SCFDriver::SCFDriver(Getkw &input) {
 
     file_start_orbitals = input.get<std::string>("Files.start_orbitals");
     file_final_orbitals = input.get<std::string>("Files.final_orbitals");
+    file_start_x_orbs = input.get<std::string>("Files.start_x_orbs");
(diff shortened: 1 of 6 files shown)


robertodr:
Suggestion: consider using https://github.com/rogersce/cnpy to write arrays to NumPy format. The files are cross-platform and can be loaded directly into Python. We've been using that library in PCMSolver and I wrote some [adapters for Eigen types](https://github.com/PCMSolver/pcmsolver/blob/master/src/utils/MathUtils.hpp#L359-L434) that you can reuse. Note that if you want to save to compressed NumPy format (basically a dictionary of NumPy arrays) you'll need to link against zlib.

stigrj:
@robertodr any idea why CircleCI sometimes fails to launch the builds? The checks below just hangs with "Expecting", and the only way I've found to relaunch is to push something new

robertodr:
No idea. Isn't there a restart button you can click in the web UI?

stigrj:
Can't find one. The PR doesn't show up in CircleCI at all, not even the current successful build...

PR #156, opened 2018-11-20
Simplified input for standard DFT functionals

stigrj:
The whole `DFT` section of the input can now be omitted for the most standard functionals. Currently implemented for LDA, BLYP, B3LYP, PBE, PBE0, BP86 and PW91.

Old setup:
```
WaveFunction {
  method = DFT
}

DFT {
  exact_exchange = 0.2
  $functionals
  BECKEX 0.80
  LYPC   1.00
  $end
}
```

New setup:
```
WaveFunction {
  method = B3LYP
}
```

### DIFF
--- a/src/mrchem.in
+++ b/src/mrchem.in
@@ -198,7 +198,7 @@ def setup_keywords(input_dir):
     # yapf: enable
     top.add_sect(molecule)
 
-    wavefunction = getkw.Section('WaveFunction')
+    wavefunction = getkw.Section('WaveFunction', callback=verify_wf)
     # yapf: disable
     wavefunction.add_kw('restricted',   'BOOL', True)
     wavefunction.add_kw('method',       'STR', 'HF')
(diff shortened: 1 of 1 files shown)


ilfreddy:
Are you sure this is necessary? The functional specifications already work (you can already say e.g. LDA) in the `$functional` section of `DFT` and the right combination is obtained from the `aliases.cpp` list in `xcfun`. The only thing truly missing is to make sure one takes the weight of the exact exchange from `xcfun`, unless one wants to override the default, of course. 
Then you avoid to duplicate specifications (which is error-prone and less general) and you don't clutter the input pursing script.
Suggestion for modifications
1. IF you specify `METHOD = B3LYP`, this will simply copy the string over to the list of functionals (one item only) with weight 1.0 .
2. You make sure the correct HF exchange weight is retrieved UNLESS a non-default value has been specified from input.

stigrj:
Made some improvements:

1. If one of the default functionals is given in `WaveFunction.method` it is copied as is into `DFT.functionals` and the correct `DFT.exact_exchange` is set. If `DFT.functionals` **or** `DFT.exact_exchange` is already set an error is flagged.
2. If you want a non-default amount of exact exchange you have to be specific, e.g. 50% in B3LYP:
```
WaveFunction {
  method = DFT
}

DFT {
  exact_exchange = 0.5
  $functionals
  BECKEX 0.5
  LYPC 1.0
  $end
}

**Comment by ilfreddy** [comment]
Did you commit the changes?

**Comment by stigrj** [comment]
Yes, I updated the old commit... I'm also trying to replace the `exact_exchange` with the internal `exx` of XCFun, which means that 50% B3LYP would be written:
```
DFT {
$functionals
  EXX 0.5
  BECKEX 0.5
  LYPC 1.0
  $end
}
```
But I don't see a way to retrieve the `exx` parameter from XCFun. The `xc_get(func, "exx", &value)` is not part of the API for some reason.

ilfreddy:
I will open an issue on XCFun about that.

stigrj:
I'm preparing a PR in XCFun where I just add `XCFun_API` to the function. If they don't like it I'm sure they have other suggestions.

ilfreddy:
You are too fast for my standards!

PR #161, opened 2018-11-29
DFT Hessian (take 2)

ilfreddy:
Linear response calculations need the second derivative of the different energy components.

- [x] Coulomb Hessian
- [x] XC Hessian (LDA)
- [ ] XC Hessian (GGA)
- [ ] Exchange Hessian
- [x] Testing


ilfreddy:
Linear response calculations need the second derivative of the different energy components.

- [x] Coulomb Hessian
- [x] XC Hessian (LDA)
- [ ] XC Hessian (GGA)
- [ ] Exchange Hessian
- [x] Testing

ilfreddy:
Still GGA and HF exchange missing, but I think this can go now. So I avoid the rebase-hell next time.

ilfreddy:
I see that all checks fail, but I have no clue how to fix them.

src/mrdft/XCFunctional.cpp
```diff
@@ -641,6 +645,7 @@ void XCFunctional::calcPotentialLDA(FunctionTreeVector<3> &potentials) {
  * and whether explicit or gamma-type derivatives have been used in xcfun.
  */
 void XCFunctional::calcPotentialGGA(FunctionTreeVector<3> &potentials) {
+    if(this->order > 1) MSG_FATAL("Not implemented: abort!");
```
stigrj: We have a `NOT_IMPLEMENTED_ABORT` message

src/qmfunctions/orbital_utils.cpp
```diff
@@ -375,6 +375,7 @@ void orbital::orthogonalize(OrbitalVector &Phi, OrbitalVector &Psi) {
     }
 }
 
+//LUCA: should this happen for all spins at the same time?
```
stigrj: There's nothing wrong with calling this function with the full `OrbitalVector`, but the resulting matrix will have a lot of zeros as `<alpha|beta> = 0`

src/qmfunctions/orbital_utils.cpp
```diff
@@ -459,20 +460,74 @@ ComplexMatrix orbital::calc_lowdin_matrix(OrbitalVector &Phi) {
...
+
+Localization is done for each set of spins separately (we don't want to mix spins when localizing).
+
+*/
+ComplexMatrix orbital::localize(double prec, OrbitalVector &Phi) {
+    Printer::printHeader(0, "Localizing orbitals");
+    Timer timer;
+    if(not orbital_order_is_sane(Phi)) {
+        orbital::print(Phi);
+        MSG_FATAL("Wrong orbital order: should be paired/alpha/beta");
```
stigrj: Do we have the necessary functionality to guarantee this ordering? I general the order of the orbitals can change during optimization.

src/qmfunctions/orbital_utils.h
```diff
@@ -50,6 +50,8 @@ OrbitalVector param_copy(const OrbitalVector &Phi);
 OrbitalVector adjoin(OrbitalVector &Phi_a, OrbitalVector &Phi_b);
 OrbitalVector disjoin(OrbitalVector &Phi, int spin);
 
+OrbitalVector filter_spin(const OrbitalVector &Phi, int spin);
```
stigrj: I suspect the above function `disjoin` does something similar to `filter_spin`

src/qmfunctions/orbital_utils.cpp
ilfreddy: I don't see why the order should change. In the diagonalization scheme (canonical orbitals) we diagonalize each Fock matrix block by itself (I implemented it myself in the new version). And in the localization scheme now we localize each set by itself. How can a reordering happen?

src/qmoperators/two_electron/CoulombDerivative.h
```diff
@@ -0,0 +1,45 @@
...
+ * If a set of orbitals has NOT been given in the constructor, the density
+ * MUST be explicitly computed prior to setup(). The density will be computed
+ * on-the-fly in setup() ONLY if it is not already available. After setup() the
+ * operator will be fixed until clear(), which deletes both the density and the
+ * potential.
+ */
+
+namespace mrchem {
+
+class CoulombDerivative: public QMPotential {
```
stigrj: Is the `CoulombDerivative` class actually in use?

ilfreddy: Guess not...

src/mrdft/XCFunctional.cpp
ilfreddy: Fixed. But I noticed such functionality is inherited from MRCPP. Maybe we should have a separate messaging library used by both MRCPP and MRChem?

stigrj: Yes, we should modernize the messaging, the current version using macros is not very nice. @robertodr, do you have anything nice in PCMSolver?

src/qmfunctions/orbital_utils.h
ilfreddy: I see an issue here. `filter_spin` does not transfer ownership, whereas `disjoin` does it. I use `filter_spin` only as a temporary `OrbitalVector` to construct the localization matrix. Such a vector is then discarded. But if I use `disjoin`, I cannot discard it aftrwards. Maybe we should discuss about how to proceed here.

ilfreddy: I like it!

ilfreddy:
@stigrj can you have a look again?

ilfreddy:
Tests pass, besides code coverage. @stigrj  do you want me to fix that too?

stigrj:
I don't understand why Travis fails. The serial run fails in H2 polarizability while the OMP run fails in Li polarizability.

src/SCFDriver.cpp
```diff
@@ -364,7 +362,10 @@ void SCFDriver::setup() {
         for (int i = 0; i < pol_frequency.size(); i++) {
             double omega = pol_frequency[i];
             molecule->initPolarizability(omega);
-            NOT_IMPLEMENTED_ABORT;
+            for (int d = 0; d < 3; d++) {
+                if (rsp_directions[d] == 0) continue;
```
robertodr: Do I understand correctly that `0`, `1`, `2` map to `x`, `y`, and `z` dipole components here?

src/SCFDriver.cpp
```diff
@@ -403,7 +404,7 @@ void SCFDriver::setup() {
     }
     //For DFT we need the XC operator
     if (wf_method == "DFT") {
-        xcfun = setupFunctional(1);
+        xcfun = setupFunctional(1); //LUCA why do we need second derivative here??
```
robertodr: For readability's sake, what about having an `enum` to encode the order? Is the comment still relevant?

src/SCFDriver.cpp
```diff
@@ -863,6 +877,7 @@ void SCFDriver::calcGroundStateProperties() {
 
 void SCFDriver::calcLinearResponseProperties(const ResponseCalculation &rsp_calc) {
     int j = rsp_calc.dir;
+    double freq = rsp_calc.freq;
 
     if (calc_magnetizability and rsp_calc.pert == h_B) {
```
robertodr: This is probably unrelated to this PR, but the `ResponseCalculation` object knows what it property it represents, doesn't it? Why not store a function pointer to assemble the property from the response vector and property gradients in the `ResponseCalculation` object and just call `rsp_calc.compute(/* X_b */, /* E_1 */)` here? This function might become a very long list of `if` otherwise...

And still unrelated, but probably relevant. The `ResponseCalculations` object is quite confusing: it just wraps a `std::vector<ResponseCalculation>` and I think that is unnecessary. It should undergo the same makeover `OrbitalVector` got over the summer.

To summarize, the final implementation for `calcLinearResponseProerties` should just loop over the response calculations, dispatch the respective perturbed orbitals and property gradients to the right `compute` methods and store the results where appropriate.

src/chemistry/Molecule.cpp
```diff
@@ -248,8 +249,6 @@ void Molecule::clearSpinSpinCoupling(int k, int l) {
 
 /** @brief Delete property Polarizability */
 void Molecule::clearPolarizability() {
-    NOT_IMPLEMENTED_ABORT;
-    /*
     int nPol = this->polarizability.size();
     for (int i = 0; i < nPol; i++) {
         if (this->polarizability[i] != 0) {
```
robertodr: Is this checking for the null pointer? If yes, please use `!= nullptr` (disambiguates for the reader and the compiler)

src/chemistry/Molecule.cpp
```diff
@@ -593,6 +585,9 @@ void Molecule::printProperties() const {
     if (this->energy != 0) println(0, *this->energy);
     if (this->dipole != 0) println(0, *this->dipole);
     if (this->geomderiv != 0) println(0, *this->geomderiv);
+    for (int i = 0; i < polarizability.size(); i++) {
+        if (this->polarizability[i] != 0) println(0, *this->polarizability[i]);
```
robertodr: Ditto here about `nullptr`

src/properties/Polarizability.h
```diff
@@ -0,0 +1,86 @@
...
+
+#pragma once
+
+#include "mrchem.h"
+
+namespace mrchem {
+
+class Polarizability final {
+public:
+    Polarizability(double w = 0.0, const double *o = nullptr, bool v = false) {
```
robertodr: As a general rule, CTORs should use [initialization lists](https://isocpp.org/wiki/faq/ctors#init-lists)
```
Polarizability(double w = 0.0, const double *o = nullptr, bool v = false) : frequency(w), origin(o), velocity(v), tensor(DoubleMatrix::Zero(3, 3) {}
```
In this case, this is basically the default CTOR,  provided you default-initialize data members (see below comment)
Additionally, the gauge origin is a 3-dimensional vector hence the proper data type is `std::array<double, 3>`, `Coord<3>` might be even better. This will avoid the need for the `this->origin` lines [which hurt my eyes pretty bad]

src/properties/Polarizability.h
```diff
@@ -0,0 +1,86 @@
...
+        o<<std::setw(19)<<tens(1,0)<<std::setw(20)<<tens(1,1)<<std::setw(20)<<tens(2,1)<<std::endl;
+        o<<std::setw(19)<<tens(2,0)<<std::setw(20)<<tens(1,2)<<std::setw(20)<<tens(2,2)<<std::endl;
+        o<<"                                                            "<<std::endl;
+        o<<"============================================================"<<std::endl;
+        o<<"                                                            "<<std::endl;
+        mrcpp::Printer::setPrecision(oldPrec);
+        return o;
+    }
+protected:
+    double frequency;
```
robertodr: Why `protected` and not `private`? Are you planning on inheriting from `Polarizability`? Does it make sense to inherit from it? If you conclude that it doesn't, mark these members `private` and the whole class `final`.

FYI, the compiler can perform some optimizations if data members are laid out smaller to larger.

tests/h_el_field/CMakeLists.txt
```diff
@@ -0,0 +1,12 @@
+set(_h_el_field_launcher " ")
+
+if(ENABLE_MPI)
+    set(_h_el_field_launcher "mpirun -np 1")
```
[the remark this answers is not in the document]
stigrj: Our MPI parallelization distributes orbitals among MPI processes and with only one orbital it doesn't make sense to use more processes.

stigrj:
Thanks for the thorough review @robertodr, I will do some polishing. Most of these issues are leftovers from the old code which are overdue to be fixed.

We have not yet settled for a good structure around the D1, D2,... classes, so your input here is appreciated. The current implementation is just a dirty one to accommodate ground-state and linear response calculations. Perhaps we can spend some time on this design at Haraldvollen?

The "love triangle" is a bit obscure. Every fundamental operator (in this case  `CoulombPotential`) needs to be wrapped in a `TensorOperator` (in this case `CoulombOperator`) before it can be used, and this is done with the assignment operator in the constructor. It looks a bit weird for such a simple operator with only one component, perhaps it's a bit clearer here:
```
class AngularMomentumOperator final : public RankOneTensorOperator<3> {
public:
    AngularMomentumOperator(mrcpp::DerivativeOperator<3> &D, const mrcpp::Coord<3> &o)
            : r(o)
            , p(D) {
        RankOneTensorOperator<3> &l = (*this);
        l[0] = (r[1]*p[2] - r[2]*p[1]);
        l[1] = (r[2]*p[0] - r[0]*p[2]);
        l[2] = (r[0]*p[1] - r[1]*p[0]);
    }

protected:
    PositionOperator r;
    MomentumOperator p;
};
```

src/SCFDriver.cpp
ilfreddy: Yes

src/qmfunctions/orbital_utils.cpp
```diff
@@ -600,7 +639,7 @@ int orbital::size_singly(const OrbitalVector &Phi) {
 int orbital::size_doubly(const OrbitalVector &Phi) {
     int nDoubly = 0;
```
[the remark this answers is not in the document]
ilfreddy: This is beyond my limited C++ skills.

src/qmoperators/two_electron/CoulombPotentialD1.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "CoulombPotential.h"
+
+namespace mrchem {
+
+class CoulombPotentialD1 final : public CoulombPotential {
```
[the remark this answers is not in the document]
ilfreddy: Actually not. This one computes the GS coulomb potential. Which is the first derivative of the coulomb energy in the functional sense. But this proves your point (need doxygen) because this is far from clear :-)

src/qmoperators/two_electron/XCHessian.cpp
```diff
@@ -0,0 +1,171 @@
...
+    println(0, " XC grid size   " << std::setw(26) << inpNodes << std::setw(17) << inpPoints);
+    println(0, " XC grid change " << std::setw(26) << newNodes << std::setw(17) << newPoints);
+}
+
+/** @brief Return FunctionTree for the input density from the XCFunctional
+ *
+ * @param[in] type Which density to return (alpha, beta or total)
+ */
+Density &XCHessian::getDensity(int spin) {
+    if (spin == DENSITY::Total) return this->functional->getDensity(mrdft::DensityType::Total);
```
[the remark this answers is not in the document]
ilfreddy: XCHessian is a leftover and will be removed

tests/h_el_field/CMakeLists.txt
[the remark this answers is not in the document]
robertodr: :sweat_smile: was looking at the code more closely than at the chemistry...

src/properties/Polarizability.h
robertodr: Also `DoubleMatrix` is dynamic in both row and column dimension, however you know up front this will be 3x3. Using `Eigen::Matrix3d` should skip any resizing.

robertodr:
Thanks for the explanation @stigrj, there might be some trick to be played to make those CTORs clearer. Of course we can look at the `Dn` design together later on.

ilfreddy:
@stigrj can you have a look now? 
This must indeed be the first time I manage to increase code coverage with my PR ;-)
I also have a test for the XC second derivative but that is segfaulting for now...
@robertodr I fixed some of the issues you have pointed out, but I have left open some others, which I thought were beyond my skills.

PR #165, opened 2018-12-04
Memory optimizations

stigrj:
Trying to reduce the memory footprint of SCF calculations.

Helmholtz application algorithm
----------------------------------------

* Greatly simplified the `HelmholtzVector` class. Removed option to reuse Helmholtz operators by storing them between iterations. Now the operators are computed on-the-fly and deallocated immediately.
* Also compute the potential part of the Helmholtz argument on-the-fly. The matrix rotation part is still computed separately for the full `OrbitalVector` due to MPI communication.

Old algorithm (with *local* orbital I mean owned by **this** MPI):
1. Construct Helmholtz operators for all local orbitals `H_i`
2. Compute Helmholtz arguments for all local orbitals `h_i =  V*phi_i + sum_j (L_ij - F_ij)phi_j`
3. Apply Helmholtz operator for all local orbitals `phi_i = H_i [h_i]`
4. Delete all `h_i`'s and `H_i`'s

New algorithm:
1. Compute matrix part for all local orbitals `g_i = sum_j (L_ij - F_ij)phi_j`
2. For each local orbital `phi_i`, do
  a. Compute potential part `f_i = V*phi_i`
  b. Compute argument `h_i = f_i + g_i`
  c. Construct Helmholtz operator `H_i`
  d. Apply Helmholtz operator `phi_i = H_i[h_i]`
  e. Delete `f_i`, `h_i` and `H_i`
3. Delete all `g_i`'s

SAD initial  guess
-----------------------

* The Fock matrix calculation within the SAD guess used the wrong algorithm for the kinetic part (`<phi|nabla^2 phi>`), which lead to excessive memory consumption and CPU time. Switched to the appropriate `<nabla phi|nabla phi>` algorithm.

NuclearPotential
---------------------

* The nuclear potential is no longer re-projected in every `setup()` (e.i. every SCF iteration). Now it's projected only once in the constructor.


gitpeterwind:
I did a test run on ch4_004 with previous master (7cc33522) and this PR (49aaddc17620). And the new version takes more memory than the old :(
new (MaxRSS 11599188K):
stallo:/global/work/mifapw/mrchem_runs/ch4_004/slurm-1349357.out
old (MaxRSS 9833836K):
stallo:/global/work/mifapw/mrchem_runs/ch4_004/slurm-1348956.out

Btw, wasn't the on-the-fly Helmholtz already part of the code one year ago?

stigrj:
Need to test some more

stigrj:
I have done some more testing, both with `/usr/bin/time -v` and with heaptrack, and the memory seems to be consistently reduced relative to the current master, both for DFT and HF calculations. It is most noticeable when there's many orbitals per MPI process, with a single orbital/proc it should in principle be the same.

Ready to merge when tests complete.

PR #170, opened 2018-12-14
Shared pointer in QMFunction

stigrj:
Collected the `*re` and `*im` `FunctionTree` pointers into a simple `ComplexFunction` class, which in turn is added as a `std::shared_ptr` to the `QMFunction` class (applies to `Orbital`, `Density` and `QMPotential` classes. This means that `QMFunction`'s  can now be assigned at will, without the fear of leaking memory at the end of scope, e.g.
```
{
  Orbital phi_a;
  {
      Orbital phi_b = O(psi); // Compute phi_b from some operator
      phi_a = phi_b;          // phi_a and phi_b points to the same FunctionTree
  }                           // phi_b out of scope, phi_a still holds the FunctionTree
}                             // phi_a out of scope, FunctionTree deallocated
```
or "in place" operations
```
phi = O(phi);                  // Application of QMOperator
Phi = orbital::rotate(U, Phi); // Unitary rotation of OrbitalVector
```

To deallocate the `FunctionTree` pointers before the object goes out of scope (this will affect all other `QMFunction`'s that point to the same function) :
```
phi.free(NUMBER::Total);       // or NUMBER::Real, NUMBER::Imag 
```

To release the `shared_ptr` before the object goes out of scope (this will *not* affect any other `QMFunction` that happens to point to the same function, but it will trigger a `delete` if this is the *last* `shared_ptr` to the function):
```
phi.release();
```


src/mrchem.h
```diff
@@ -13,29 +13,30 @@
...
+const double beta_N  =    0.0002723;       // Nuclear magneton
+} // namespace PHYSCONST
 
 namespace MATHCONST {
-    const double pi      =    3.1415926535897932384;
-    const double sqrt_pi =    1.7724538509055160273;
-}
+const double pi      =    3.1415926535897932384;
+const double sqrt_pi =    1.7724538509055160273;
+} // namespace MATHCONST
```
ilfreddy: Unrelated from this PR, but I think we should implement the same design of physical constants that @robertodr has used in PCMSolver. It allows to select a group of constants from a given standard set, thus increasing reproducibility of results.

robertodr: For `pi`&co I would use `M_PI`&co as defined by the standard. For physical constants, have a look at this MolSSI project: https://github.com/MolSSI/QCElemental How to integrate with C++, I don't know.

ilfreddy:
Nice PR and excellent increase of code coverage. Great job!

PR #175, opened 2018-12-21
Bump C++ standard to 14

stigrj:
Since MRCPP now requires `c++14`


stigrj:
@robertodr, we need CMake 3.11 on CircleCI. How can I achieve this?

PR #182, opened 2019-01-05
Reorder initialization list

robertodr:
Fixes a compiler warning about data member reordering.


robertodr:
MPI job timed out, I guess a kick (aka "rerun workflow" on CircleCI) would do

PR #184, opened 2019-01-08
omp parallelization of Hessian matrix mult for localization

gitpeterwind:
Small gain in cpu time, but only a few lines extra.
Also set std=c++14 compiler flag (for compatibility with mrcpp)


stigrj:
The compiler flag has been fixed in a better way in `master`, so I suggest you do the following:
1. revert compiler flag change back to `c++11`
2. `git commit --amend` to redo your commit
3. `git fetch upstream`
4. `git rebase upstream/master`
5. `git push -f origin ompRR`

PR #193, opened 2019-01-22
Update Pipfile with pyyaml-4.2b1

stigrj:
Should remove security alert


stigrj:
Yes, we should add a section in `doc/installation.rst` on pipenv and the Pipfile. It's just that I'm not entirely sure how it works :slightly_smiling_face:

robertodr:
Can you open an issue and assign me to it?

PR #196, opened 2019-01-23
Danger systems

bjorgve:


ilfreddy:
I think the only danger here is when @stigrj @robertodr and @bjorgve get together to supposedly "improve" mrchem!!!

robertodr:
screenshot from 2019-01-23 20-21-28

.travis.yml
```diff
@@ -60,6 +66,11 @@ env:
     - CMAKE_VERSION="3.12.4"
     # Invalidate PYENV_ROOT
     - PYENV_ROOT=$HOME/fuffa
+    - |
```
robertodr: This goes under the `before_script` field.

robertodr: Or anywhere else but `env`. These lines are prefixed by `export` when the YAML is processed.

PR #198, opened 2019-01-27
Finally the superbuild

robertodr:
As the title says, _finally_ the superbuild. I've used the [`FetchContent` approach](https://github.com/dev-cafe/cmake-cookbook/blob/master/chapter-08/recipe-04/README.md), rather than `ExternalProject_Add`. The latter turned out to be quite tricky to get right (tests were failing) and required moving around basically the whole source tree.
The superbuild required some modifications in XCFun (see dftlibs/xcfun#91), Getkw (see dev-cafe/libgetkw#29) and MRCPP (see MRChemSoft/mrcpp#75).
To use precompiled versions of XCFun, Getkw and MRCPP you can do:
```
./setup --cmake-options="-DMRCPP_DIR=$HOME/Software/junk/share/cmake/MRCPP -DXCFun_DIR=$HOME/Software/junk/share/cmake/XCFun -Dgetkw_DIR=$HOME/Software/junk/share/getkw"
``` 
(the dependencies are assumed to have been installed under `$HOME/Software/junk`)
Eigen3 is also fetched if not already available. I've removed the submodules. Can the `pilot` folder be axed too?


external/upstream/fetch_getkw.cmake
```diff
@@ -0,0 +1,27 @@
...
+else()
+  message(STATUS "Suitable getkw could not be located. Fetching and building!")
+  include(FetchContent)
+
+  FetchContent_Populate(getkw_sources
+    QUIET
+    GIT_REPOSITORY
+      https://github.com/dev-cafe/libgetkw.git
+    GIT_TAG
+      master # Preferable to have a tag for a release
```
bast: I would change "master" to a hash.

external/upstream/fetch_xcfun.cmake
```diff
@@ -0,0 +1,33 @@
...
+  enable_language(C)
+  message(STATUS "Suitable XCFun could not be located. Fetching and building!")
+  include(FetchContent)
+
+  FetchContent_Populate(xcfun_sources
+    QUIET
+    GIT_REPOSITORY
+      https://github.com/dftlibs/xcfun.git
+    GIT_TAG
+      master # Preferable to have a tag for a release
```
bast: Same here: hash instead of "master" otherwise future versions of XCFun can break past versions of MRChem.

bast:
For my own education: what is a "sad" basis?

stigrj:
Is the GTO basis set used in our Superposition of Atomic Densities initial guess

bast:
Thanks!

external/upstream/fetch_xcfun.cmake
robertodr: Done.

external/upstream/fetch_getkw.cmake
robertodr: Done.

bast:
It seems the testing cannot locate binary/binaries.

robertodr:
Yep, I fixed one thing in MRCPP and broke the superbuild. It's the include paths that are flaky right now...

robertodr:
@stigrj, please double-check this works on stallo and/or fram with the Intel compilers. Can the `pilot` folder be removed? Are its contents important as an example? Maybe it should be moved to the docs?

@ilfreddy please double-check this works on your Mac.

stigrj:
Everything except Eigen works on Fram!

robertodr:
I'll wait for further feedback on the Mac build. I guess the only thing that's really missing is documentation on building and installing?

stigrj:
Same problem on Stallo, MRCPP works but MRChem fails on Eigen. Actually, also on my laptop if I remove all existing Eigen installations. It is downloaded and installed, but there's no `Eigen3Targets.cmake` in this folder:
```
stig@pilates :: ~/src/mrchem-super ‹finally-the-superbuild› » ll build-super/_deps/eigen3_sources-build/
total 52K
drwxr-xr-x 3 stig stig 4,0K feb.   2 20:32 bench
drwxr-xr-x 3 stig stig 4,0K feb.   2 20:32 blas
-rw-r--r-- 1 stig stig 1,6K feb.   2 20:32 cdashtesting.cmake
drwxr-xr-x 2 stig stig 4,0K feb.   2 20:32 CMakeFiles
drwxr-xr-x 3 stig stig 4,0K feb.   2 20:32 demos
drwxr-xr-x 7 stig stig 4,0K feb.   2 20:32 doc
drwxr-xr-x 3 stig stig 4,0K feb.   2 20:32 Eigen
-rw-r--r-- 1 stig stig 1,3K feb.   2 20:32 Eigen3Config.cmake
-rw-r--r-- 1 stig stig 1,6K feb.   2 20:32 Eigen3ConfigVersion.cmake
-rw-r--r-- 1 stig stig  249 feb.   2 20:32 eigen3.pc
drwxr-xr-x 3 stig stig 4,0K feb.   2 20:32 lapack
drwxr-xr-x 3 stig stig 4,0K feb.   2 20:32 scripts
drwxr-xr-x 5 stig stig 4,0K feb.   2 20:32 unsupported
```

robertodr:
Can you grep for `find_package(Eigen`? There should only be one call to that. The current set up is technically _not_ a superbuild. In a "proper" superbuild all subprojects are built separately and installed. The superproject is then fed the installation folders and calls `find_package` to get dependencies through CMake targets. In the current approach, there is no installation step. However, everything is built within the same tree, which means that `Eigen3::Eigen` and `MRCPP::mrcpp` are targets of the MRChem project. Once you've fetched, there is no need for additional `find_package` calls (and they won't work anyway) @bast can add more details or correct me if I am wrong.

In a very early iteration of this PR, I had put a second call to `find_package` before figuring out it would always fail when fetching. Could it be that somehow your history is focked up a little?

stigrj:
There's just one call in `external/upstream/fetch_eigen3.cmake`. Currently Eigen goes under `build/_deps/` as it does in the standalone mrcpp (and working), while mrcpp, xcfun and getkw are located directly under `build/`. Is this how it should be?

stigrj:
This is not the most pressing issue, we've managed to use an external Eigen up to this point. I just have to re-install those I have remove in order to test this 🙂

stigrj:
I still get include errors if I try to use an external MRCPP:
```
[  2%] Built target getkw
[  3%] Built target cxx_test
[ 43%] Built target xcfun
[ 43%] Built target xcfun-c-objlib
[ 44%] Built target xcfun-shared
[ 45%] Built target xcfun-static
[ 47%] Built target testall
[ 48%] Building CXX object src/CMakeFiles/mrchem.dir/parallel.cpp.o
In file included from /home/stig/dev/mrcpp-omp/include/MRCPP/trees/FunctionTree.h:33,
                 from /home/stig/dev/mrcpp-omp/include/MRCPP/trees/FunctionNode.h:30,
                 from /home/stig/src/mrchem-stig/src/parallel.cpp:3:
/home/stig/dev/mrcpp-omp/include/MRCPP/trees/MWTree.h:35:10: fatal error: utils/omp_utils.h: No such file or directory
 #include "utils/omp_utils.h"
          ^~~~~~~~~~~~~~~~~~~
compilation terminated.
make[2]: *** [src/CMakeFiles/mrchem.dir/build.make:62: src/CMakeFiles/mrchem.dir/parallel.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1523: src/CMakeFiles/mrchem.dir/all] Error 2
make: *** [Makefile:141: all] Error 2
```

robertodr:
On fram? Is this trying to fetch? Or did it find a preinstalled MRCPP?

stigrj:
No, I tried on my laptop. First install MRCPP separately, then `export MRCPP_DIR=<mrcpp-install-dir>` before `./setup` in MRChem. It's able to locate MRCPP, but fails in compilation due to include errors.

robertodr:
It's the latest MRCPP master, right? Just trying to exclude the obvious.

stigrj:
Yes :slightly_smiling_face:

robertodr:
OK, will test some more.

robertodr:
There was a wart in the definition of `INTERFACE_INCLUDE_DIRECTORIES` (testing is never enough, it seems, but thankfully we have code review!) This is fixed in MRChemSoft/MRCPP#82 If that fixes your problems (worked locally for me) then the value of the `GIT_TAG` record in the call to `FetchContent_Populate` should be updated.

stigrj:
Yes!! @luca approves! Do you want to work on the git history or merge as is?

robertodr:
I'll clean up a bit. Let me write some docs before merging.

robertodr:
Cool, cleaned up history and destroyed everything. 
no-fun-not-ever

robertodr:
Docs updated @stigrj @gitpeterwind @bjorgve have a look.

doc/installation.rst
```diff
@@ -84,6 +83,11 @@ The public version of MRChem is available on GitHub::
 
     $ git clone git@github.com:MRChemSoft/mrchem.git
 
+MRChem depends on `Eigen3 <http://eigen.tuxfamily.org/index.php?title=Main_Page>`_, `XCFun <https://github.com/dftlibs/xcfun>`_,
+`Getkw <https://github.com/dev-cafe/libgetkw>`_ and `MRCPP <https://github.com/MRChemSoft/mrcpp>`_.
+We have structured the build system such that these dependencies will be **fetched**
+when configuring the project if they are not already installed.
+
```
stigrj: The Eigen fetch still doesn't work on my laptop or on Fram if it's done directly in MRChem, so perhaps we should not advocate this? It works, however, if you first fetch it in a separate MRCPP before running `./setup` in MRChem. You don't even have to build or install MRCPP, you just need to do `./setup` in MRCPP first, then MRChem will find Eigen in the MRCPP folder.

bast:
"no fun not ever" :-) that's the best!

doc/installation.rst
robertodr: I'll remove that.

stigrj:
:tada: :tada: :tada:

robertodr:
Only one year after I said I'd do it! What a record!

PR #200, opened 2019-01-28
Update installation.rst

gitpeterwind:


doc/installation.rst
```diff
@@ -103,6 +103,12 @@ To build the code with hybrid MPI/OpenMP parallelization::
 
 The MRChem executables will be installed in ``<install-dir>/bin``.
 
+If you want only to recompile one of the external library (for example mrcpp), without rebuilding from scratch, try::
+
+   $ cd <build-dir>/subproject/Build/mrcpp_external
+   $ make install
+
```
stigrj: This will probably change with #198. Not sure how the superbuild approach will work with in-source development of MRCPP, @robertodr? We probably need to link to a separate MRCPP?

doc/installation.rst
```diff
@@ -58,17 +58,17 @@ Stallo
 Using the Intel tool chain on Stallo::
 
     $ module load intel/2017a
-    $ module load CMake/3.9.1
+    $ module load CMake/3.11.4-GCCcore-7.3.0
```
stigrj: I think we need to load `CMake` first, since `intel` loads a different GCCcore?

gitpeterwind: I couldn't find a set that does not create a conflict between libraries. (yes setting intel last is what I actually did in my tests, but was lazy editing)

gitpeterwind: Now, how do I modify/edit this, when I did the modification by editing the file on github/MRChemSoft?

stigrj: You should be able to do the same on the `gitpeterwind-patch-1` branch of MRChemSoft/mrchem

gitpeterwind: ah, yes, I didn't realize that a new branch had been created

doc/installation.rst
stigrj: We'll keep this for now until the superbuild is ready, might still take a while :stuck_out_tongue:

PR #203, opened 2019-01-31
[WIP] Solvent 

Gabrielgerez:
To do:
- add tests for Solvent effect
- fix memory leaks (done)
- fix performance


src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,104 @@
...
 
 double Cavity::evalf(const mrcpp::Coord<3> &r) const {
     double C = 1.0;
     double s, O;
-
+    double val;
     for (int i = 0; i < pos.size(); i++) {
-        s = math_utils::calc_distance(pos[i], r) - R[i];
+        if (abc) {
+            s = math_utils::calc_distance(pos[i], r) - R[i] * 1.2; // scale vdwr by 1.2
```
ilfreddy: A more flexible way of including scaling factors (like the 1.2 you have now) is to create a vector of factors which is as long as the vector of radii. Then you can simply do `R[i] * alpha[i]` avoiding the if statement.

src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,104 @@
...
+    } else {
+        if (is_linear) {
+            val = e_i * C + e_o * (1 - C);
+        } else {
+            val = e_i * std::exp(log(e_o / e_i) * (1 - C));
+        }
+    }
+
+    if (for_rho) { val = (1 - val) / val; }
+    return val;
```
ilfreddy: I assume a function like `evalf`is used many times. A such, you should try to avoid to have all these if statements here.
Suggestions:
1. the last if can be left outside of the function. The caller will know whether it needs `val` or `1-val/val` and you avoid the if altogether.
2.a you can substitute the ìf` statements for `is_inv` `is_linear` with a `switch` construct based on the bitwise representation of the two booleans. 
2.b ideally you should have one function for each of the cases, thus you can move the  `if` outside and loop on a simple function evaluation.

src/qmoperators/two_electron/ReactionOperator.h
```diff
@@ -0,0 +1,45 @@
...
+    ~ReactionOperator() {
+        if (this->potential != nullptr) delete this->potential;
+    }
+
+    ComplexDouble trace(OrbitalVector &Phi) { return RankZeroTensorOperator::trace(Phi); }
+
+    double &get_tot_Energy() { return this->potential->get_tot_Energy(); }
+    double &get_e_Energy() { return this->potential->get_e_Energy(); }
+    double &get_nuc_Energy() { return this->potential->get_nuc_Energy(); }
+
```
ilfreddy: try to adopt the MRChem naming convention: methods and data members of a class are written as a single string (no underscore) capitalizing the words (except the first). Words are generally written in full
Example `getTotalEnergy` instead of `get_tot_Energy`.

Gabrielgerez: I've changed the naming now

src/chemistry/Cavity.cpp
Gabrielgerez: added the vector now

src/chemistry/Cavity.cpp
Gabrielgerez: made it so the implementation of `1-val/val` is on ReactionPotential instead and added a switch statement.
i believe is is possible to shorten the code a bit more by adding a lambda expression that returns `1-cavity.evalf/cavity.evalf` which i haven't gotten to work yet

PR #204, opened 2019-01-31
Parallelize potentials

stigrj:
Switch parallelization algorithm for the Coulomb potential
----------------------------------------------------------------------
Old algorithm:
  1. All ranks compute local density from own orbitals
  2. Rank 0 collects local contributions into a global density
  3. Rank 0 distributes global density to all ranks
  4. All ranks compute global potential from global density

New algorithm:
  1. All ranks compute local density from own orbitals
  2. All ranks compute local potential from local density
  3. Rank 0 collects local contributions into a global potential
  4. Rank 0 distributes global potential to all ranks

~Switch between new and old algorithm with the new input keyword~
~`
MPI {
  local_coulomb_potential = true/false
}
`~

Switch between new and old algorithm with the keyword
```
MPI {
  numerically_exact = true/false (true = global, false = local)
}
```

Parallelize projection of nuclear potential
---------------------------------------------------
The nuclear potential is computed only once in the beginning of the calculation (twice if you use SAD or CORE initial guess), but it was previously not MPI parallelized, only OMP.

New algorithm:
  1. Distribute nuclei among MPI ranks
  2. All ranks project local potential from own nuclei
  3. Rank 0 collects local contributions into a global potential
  4. Rank 0 distributes global potential to all ranks

Misc
------
- Add printouts and timings for orbital rotations

A significant amount of the computation time for each SCF cycle was not accounted for in the printed output. This was due to orbital rotations in Löwdin orthonormalizations and the preprocessing of Helmholtz operator argument.

Future work
---------------
We should make all-to-one reduction and one-to-all broadcast algorithms that scales with number of MPI.

Status
--------
- [x] Extensive testing
- [x] Cleanup
- [x] Ready to review
- [x] Ready to merge


gitpeterwind:
Is the goal to avoid to store the complete density? (It is not that large, for example 900MB for CH4_040, while exchange and some part in the Helmoltz operations are larger and also grow with system size).
With "local", you mean here on the same MPI (or compute-node)? (one could define proper subsystems too)

stigrj:
Yes, with local I mean density from orbitals located on a given MPI, they are not necessarily localized in space.

The goal is to reduce the memory in the Coulomb part but also to improve efficiency/scalability. In the old version all MPIs compute the same full potentials, both for Coulomb and nuclear, e.i. no MPI parallelization at all. Also, since the Coulomb potential is much smaller in size than the density, both the communication and reduction addition should be more efficient.

The XC part is more problematic and requires a different solution.

gitpeterwind:
Now I understand. If the local Coulomb potential takes less space than the local densities it is a logical step.
BTW, we should also make the reduce/broadcast as log scaling; should be easy (we even had a working version some years ago).

gitpeterwind:
Could you explain what happens in the changes in OrbitalOptimizer.cpp, with Helmholtz H?

stigrj:
The Helmholtz operator involves the following (L is diagonal and typically cancels the diagonal in F):
`\phi_i ^{n+1} = H[V*\phi_i^n + \sum_j (L_ij - F_ij)\phi_j^n]`

This is done in two steps:
1. First the matrix part is done collectively since it involves all MPIs
`Psi_n = orbital::rotate(L-F, Phi_n);`

2. Then the rest is done one orbital at the time, locally on each MPI since it requires no communication
`Phi_np1[i] = H(V(Phi_n[i]) + Psi_n[i]);`

I haven't analyzed the first orbital rotation. Intuitively I wouldn't expect the resulting orbitals `Psi_n[i]` to be localized, but based on the properties of the Helmholtz operator I can't see that they can **not** be localized, since the resulting orbitals `Phi_np1[i]` are localized, and H is a smoothing operator.

gitpeterwind:
yes, but in your commit called "add rotation print", you seem to replace 
`OrbitalVector Psi = orbital::rotate(L_mat - F_mat, Phi_n);` 
with 
`OrbitalVector Psi = H(F_mat, Phi_n);`
Is does not look identical?

stigrj:
Ah, no I just moved the rotation into the `HelmholtzOperator` class to avoid cluttering the `OrbitalOptimizer::optimize()` function with timers and print statements. It should do the same thing.

gitpeterwind:
I think I understand. But it is confusing to use H() for so different things:`H(F_mat, Phi_n)'  just rotate orbitals without applying H, and `H(V, Phi_n, Psi)` applies H. Shouldn't the first have another name or notation?

stigrj:
Yes, you are right. I will come up with a different name

gitpeterwind:
I did a test on H2O with 5 MPI and it fails! 
Short after start in the Allreduce potential after Nuclear potential.
Maybe I did something wrong, but it worked fine for larger molecules!

gitpeterwind:
it works with water for mpi=1,2, or 3, but not 4 or 5. 
(Could there be something with the number of nuclei<number of MPI ?)

stigrj:
Yes, I didn't consider the case when some MPIs don't get any nuclear potential to project. Should be a simple fix.

gitpeterwind:
Shouldn't the "numerically_exact" option also be used in the potentials?

stigrj:
Yes, I will put that in as well

stigrj:
Do you have a working log reduction algorithm to put on top of this?

gitpeterwind:
There is both a reduction and broadcast algoritm

stigrj:
The `numerically_exact` option cannot be used with the new local algorithm for the Coulomb potential, so I have removed the `local_coulomb_potential` input parameter. Now the local vs global algorithm is chosen based on the `numerically_exact` keyword, where exact means global algorithm, otherwise local.

The reason the exact option **can** be used with the global algorithm but not with the local is that the addition is an exact operation in MW (provided union grid) while the Poisson operator is not. This means that the global algorithm

```V_tot = \int P(r-r') \sum_mpi \rho_loc(r') dr'```

is identical, irrespective of the MPI distribution of the orbitals, while the local algorithm

```V_tot = \sum_mpi \int P(r-r') \rho_loc(r') dr'```

is not identical, since there is **a lot** of local irreversible thresholding in the Poisson application.

gitpeterwind:
Except for the problems we had with shared densities, is there a reason why we should first sum the results in rho_loc before copying to rho_tot?

stigrj:
No other reason, I think

PR #207, opened 2019-02-14
[WIP] Fix path to getkw Python packages

### LINKED ISSUE #206: Preinstalled Getkw
The following copy paths are not correct when using a preinstalled `getkw`:
https://github.com/MRChemSoft/mrchem/blob/c67777a665e633a826913849d0db5b5ee2c6b078/cmake/custom/main.cmake#L10-L16

In particular, the `getkw_sources_BINARY_DIR` variable seems empty. There's also an error in the Python install path in `getkw` (https://github.com/dev-cafe/libgetkw/issues/30), but that's only half the issue.

stigrj:
Not sure this is the correct path to include from, but it works

### DIFF
--- a/cmake/custom/main.cmake
+++ b/cmake/custom/main.cmake
@@ -9,8 +9,8 @@ set(PYTHON_SITE_INSTALL_DIR ${CMAKE_INSTALL_LIBDIR}/python${PYTHON_VERSION_MAJOR
 include(${PROJECT_SOURCE_DIR}/external/upstream/fetch_getkw.cmake)
 file(
   COPY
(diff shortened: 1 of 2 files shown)


stigrj:
Noo, this does not work when `getkw` is fetched in the superbuild

stigrj:
Ok, this works, but it feels dirty

robertodr:
I think the CMake configuration module needs an update. I am not sure if I'll get the time today or later in the weekend, though.

robertodr:
I think #210 supersedes this one (and it's less dirty :wink: )

stigrj:
Fixed in #210

PR #209, opened 2019-02-14
Fix FetchContent

stigrj:
* Remove `GIT_SHALLOW 1` in MRCPP, XCFun and Getkw so that we can point to older revisions.
* Update MRCPP and Getkw to latest and greatest.

### DIFF
--- a/external/upstream/fetch_getkw.cmake
+++ b/external/upstream/fetch_getkw.cmake
@@ -11,7 +11,7 @@ else()
     GIT_REPOSITORY
       https://github.com/dev-cafe/libgetkw.git
     GIT_TAG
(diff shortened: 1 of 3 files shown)


stigrj:
Why does Circle always fail on the first try with MPI? :angry:

robertodr:
Probably the threshold is too tight and MPI generates slightly non-deterministic results on the last digits?

stigrj:
Oh wait, now it's actually a real error :slightly_smiling_face: But usually it's one integration test that times out in the first run, but works after restart.

stigrj:
I'm rewinding to previous MRCPP, the latest version is not `numerically_exact` with different number of MPIs. I will address this problem in #204.

PR #210, opened 2019-02-17
Fix path to getkw Python packages

robertodr:
My take on #207 in tandem with dev-cafe/libgetkw#32. I have tested with a prebuilt and a fetched getkw. When dev-cafe/libgetkw#32 is in, I'll switch back to the right commit on the canonical repo.


robertodr:
The MPI build on circle seems stuck. Can you give it a kick?

stigrj:
I don't have permissions to kick

robertodr:
It seems it failed, but I can't restart either... It will rerun when I update the repo in `fetch_getkw.cmake` :+1:

PR #218, opened 2019-03-05
Update MRCPP

stigrj:


stigrj:
Yes, we *really* should put some effort into proper versioning on both MRCPP and MRChem

PR #219, opened 2019-03-05
Update Dangerfile

robertodr:
Finesse the `has_code_changes` rule. Changes to the `cmake` folder do not mean that new tests should be added.


robertodr:
Nice, the rule is malformed :man_facepalming:

robertodr:
I suggest a squash-merge.

PR #220, opened 2019-03-08
Clang tidy

bjorgve:
Clang-tidy fixes on MRChem. Similar to the ones on mrcpp.

### DIFF
--- a/src/SCFDriver.cpp
+++ b/src/SCFDriver.cpp
@@ -321,8 +321,7 @@ void SCFDriver::setup() {
         }
     }
     if (calc_nmr_shielding) {
-        for (int k = 0; k < nmr_nucleus_k.size(); k++) {
-            int K = nmr_nucleus_k[k];
+        for (auto &K : nmr_nucleus_k) {
             molecule->initNMRShielding(K);
             if (nmr_perturbation == "B") {
                 for (int d = 0; d < 3; d++) {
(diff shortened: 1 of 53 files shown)


src/SCFDriver.cpp
```diff
@@ -426,34 +418,34 @@ mrcpp::DerivativeOperator<3> *SCFDriver::useDerivative(string derivative_name) {
 
 void SCFDriver::clear() {
```
robertodr: This function should be called in the destructor, to clear up resources automatically when the object goes out of scope.

src/analyticfunctions/NuclearGradientFunction.cpp
```diff
@@ -41,7 +41,7 @@ double NuclearGradientFunction::evalf(const mrcpp::Coord<3> &r) const {
 
 bool NuclearGradientFunction::isVisibleAtScale(int scale, int nQuadPts) const {
     double stdDeviation = std::pow(this->smooth, -0.5);
-    int visibleScale = int(std::floor(std::log2(nQuadPts * 5.0 * stdDeviation)));
+    auto visibleScale = int(std::floor(std::log2(nQuadPts * 5.0 * stdDeviation)));
```
robertodr: `static_cast<int>` here too.

src/SCFDriver.cpp
```diff
@@ -170,46 +170,46 @@ SCFDriver::SCFDriver(Getkw &input) {
...
-    dK = 0;
-    dXC = 0;
-    d_fock = 0;
-
-    xcfun = 0;
-
-    h_E = 0;
-    h_B = 0;
-    h_M = 0;
+    kain = nullptr;
```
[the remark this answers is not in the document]
stigrj: No worries, all this will be thoroughly revised with a new driver in #212

src/SCFDriver.cpp
```diff
@@ -321,8 +321,7 @@ void SCFDriver::setup() {
         }
     }
     if (calc_nmr_shielding) {
-        for (int k = 0; k < nmr_nucleus_k.size(); k++) {
-            int K = nmr_nucleus_k[k];
+        for (int K : nmr_nucleus_k) {
```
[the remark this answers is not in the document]
bjorgve: Yes good catch!

src/analyticfunctions/NuclearGradientFunction.cpp
bjorgve: Okei good catch :)

bjorgve:
It looks like it's timing out on circleci, am I doing something wrong or is it just circleci that's stupid?

src/SCFDriver.cpp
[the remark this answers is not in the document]
bjorgve: Okei, I'll leave this be then

stigrj:
The first MPI build tends to time out on Circle, restarting or pushing new commits usually helps

src/SCFDriver.cpp
bjorgve: You'll fix this in your PL @stigrj ?

bjorgve:
Think this is more or less ready to go.

PR #221, opened 2019-03-12
Lots of std::shared_ptr

### LINKED ISSUE #105: Use `std::shared_ptr` in TensorOperators

stigrj:
Extracted some of the noise from #212 which is unrelated to the input and driver.

Changes made
------------------
- Move orbitals and Fock matrix into the `Molecule` class
- Remove all `init`s of molecular properties, now they are allocated on the fly
- Introduce `std::shared_ptr` into QM operators, closes #105 
- Remove a lot of unnecessary state in the SCF solvers
- Add Fock matrix to argument list of `localize` and `orthonormalize` to make them consistent with `diagonalize`
- Update integration test references for calculations involving the `EnergyOptimizer` and `LinearResponseSolver`, since minor algorithmic changes made the results slightly different
- Adapt `SCFDriver` to all changes

Please ignore all changes in `SCFDriver`, this will be fully revised in #212.


src/SCFDriver.h
```diff
@@ -205,53 +206,39 @@ class SCFDriver final {
...
 
     // XCFun
-    mrdft::XCFunctional *xcfun;
+    std::shared_ptr<mrdft::XCFunctional> xcfun{nullptr};
 
     // Perturbation operators
-    H_E_dip *h_E;  // dH/dE
-    H_B_dip *h_B;  // dH/dB
-    H_M_pso **h_M; // dH/dM[K]
+    H_E_dip *h_E{nullptr};  // dH/dE
```
robertodr: Must these be pointers?

src/qmoperators/one_electron/DistanceOperator.h
```diff
@@ -21,13 +21,13 @@ class DistancePotential final : public QMPotential {
 class DistanceOperator final : public RankZeroTensorOperator {
 public:
     DistanceOperator(double pow, const mrcpp::Coord<3> &R, double S = 1.0e-7)
-            : r_pow(pow, R, S) {
+            : r_pow(new DistancePotential(pow, R, S)) {
```
robertodr: There might be copies incurred in doing the initialization like this. Probably can be moved to the body of the CTOR to use `make_shared` instead.

src/SCFDriver.h
stigrj: No, but it will be fixed later

src/qmoperators/one_electron/DistanceOperator.h
stigrj: Hmm, really? I will change to `make_shared` then

stigrj:
I will run some tests on Fram, but at this abstraction level I doubt we'll see any change in performance

stigrj:
No discernible change in performance

PR #223, opened 2019-03-18
Periodictable: better vdW. radii

Gabrielgerez:
Use PCMSolver's  Allinger's MM3 radii to correct the vdW. radii in `PerdiodicTable.cpp`


robertodr:
@Gabrielgerez you might want to check with @ilfreddy which radii should be default, especially when using them to build a cavity for continuum solvation. Bondi radii (times a scaling factor of 1.2) have historically been the default in many codes, including PCMSolver.

Gabrielgerez:
There are Bondi radii that are `0.0`, should i just match those with the ones in `PeriodicTable.cpp` or just keep the old ones?

robertodr:
My strategy was to throw an error in those cases and tell the user to provide a radius that seemed sensible.

Issue #228, opened 2019-06-18
Pilot code not compiling


Gabrielgerez:
- There is a case error in line 11 of `mrchem.cpp.sample`: `Input` should be `input`.

- Following instructions from `README`on pilot gives me the following error:
```
CMakeFiles/mrchem-pilot.dir/mrchem.cpp.o: In function `main':
mrchem.cpp:(.text.startup+0x1c): undefined reference to `mrchem::mpi::initialize(int, char**)'
mrchem.cpp:(.text.startup+0x29): undefined reference to `mrenv::initialize(int, char**)'
mrchem.cpp:(.text.startup+0x4b): undefined reference to `mrenv::finalize(double)'
mrchem.cpp:(.text.startup+0x50): undefined reference to `mrchem::mpi::finalize()'
collect2: error: ld returned 1 exit status
pilot/CMakeFiles/mrchem-pilot.dir/build.make:84: recipe for target 'bin/mrchem-pilot' failed
make[2]: *** [bin/mrchem-pilot] Error 1
CMakeFiles/Makefile2:2329: recipe for target 'pilot/CMakeFiles/mrchem-pilot.dir/all' failed
make[1]: *** [pilot/CMakeFiles/mrchem-pilot.dir/all] Error 2
Makefile:140: recipe for target 'all' failed
make: *** [all] Error 2
```
-  `cp pilot/mrchem.cpp.sample src/mrchem.cpp` enables the pilot code when compiling as standard, and one can run it by using `./mrchem mrchem.inp`. one must keep the real `mrchem.cpp` elsewhere if one wants to do normal calculations, at which point one just replaces the pilot `mrchem.cpp` with the old one and compile. 

We should try to avoid this file copying and renaming, if possible

PR #229, opened 2019-06-19
Issue with pilot

robertodr:
This should solve #228 Please try it out. Note that the Python front-end and the actual executable are saved to `<build_dir>/bin` and that the example input is in `<build_dir>/pilot`.

@stigrj is anyone else using `.envrc`? If not, I'll remove it.


stigrj:
Not sure what `.envrc` does, so probably not

src/CMakeLists.txt
```diff
@@ -116,7 +113,7 @@ install(
 # <<< Build initial guess executables >>>
 add_subdirectory(initial_guess)
 
-set(MRCHEM_EXECUTABLE ${CMAKE_INSTALL_PREFIX}/${CMAKE_INSTALL_BINDIR}/mrchem.x)
+set(MRCHEM_EXECUTABLE $<TARGET_FILE:mrchem.x>)
```
stigrj: It seems this `TARGET_FILE` stuff is not correctly expanded. It appears as `default='$<TARGET_FILE:mrchem.x>'` in the `mrchem` (and `mrchem-pilot`) script. The integration tests goes through since they set the executable explicitly with `--executable`.

robertodr:
Should be fixed now. When merging, please *squash*

stigrj:
@Gabrielgerez, the `mrchem-pilot` should now work as expected

PR #230, opened 2019-06-24
MolPlot -> MolPlotter

stigrj:
Adapting the `MolPlotter` class to the new `mrcpp::Plotter` and add input option to plot converged orbitals and densities in SCF. The output files `.cube` can be directly visualized in the browser by [blob](https://github.com/densities/blob). For a 20x20x20 plot in the volume [-4,4]^3:
```
Plotter {
  points = [20, 20, 20]  # number of grid points
  O = [-4.0,-4.0,-4.0]   # plot origin
  A = [8.0, 0.0, 0.0]    # boundary vector
  B = [0.0, 8.0, 0.0]    # boundary vector
  C = [0.0, 0.0, 8.0]    # boundary vector
}

SCF {
  plot_density = true    # plot converged density (including spin for open-shell)
  plot_orbital = [1,2,4] # plot converged phi_1, phi_2 and phi_4
}
```


stigrj:
Done! The docs will have to be thoroughly revised soon, though

gitpeterwind:
I wanted to try it out on Fram, but it seems python needs pathlib:
``` mrchem -D mrchem.inp
Traceback (most recent call last):
  File "./mrchem", line 37, in <module>
    from pathlib import Path
ImportError: No module named pathlib
```
Any simple solution?

stigrj:
Note that this version relates to the upcoming release v1.0 with `parselglossy` input instead of `getkw`, which also requires Python3. Have you used this version before? It requires that you do
```
$ pipenv install --dev
$ pipenv shell
```
inside the `mrchem` directory. I think `pipenv` is already available on Fram, but you need to load a Python3 version. Some input parameters have also been changed or moved, but you can have a look in `src/input/template.yml` for a comprehensive list.

gitpeterwind:
I made a clone of your rep. That should include "parselglossy" (whatever that is)?
I did not find pipenv on Fram. Do you have a complete recipe?

stigrj:
Try the following (not sure if it is completely kosher):
```
$ module load intel/2017a Python/3.6.1-intel-2017a CMake/3.12.1 git/2.14.1-GCCcore-6.4.0
$ pip install --user pipenv
$ cd mrchem
$ git checkout molplot
$ pipenv install --dev
$ pipenv shell
$ ./setup --prefix=install-dir --cxx=icpc --omp build-dir
$ cd build-dir
$ make install
$ ctest
```

stigrj:
Actually, the integration part of `ctest` fails on Fram, but a manual run seems to work...

stigrj:
Works on Saga, though :sunglasses:

gitpeterwind:
Thanks, almost there... How do I get 'parselglossy' ?

stigrj:
it should be there after `pipenv shell`, no need to do anything

gitpeterwind:
```
$pipenv shell
Shell for /cluster/home/mifapw/.local/share/virtualenvs/mrchem-kindcob0 already activated.
No action taken to avoid nested environments.
(mrchem) login-1-2 /cluster/projects/nn4654k/peter/h2o $ mrchem -D mrchem.inp
Traceback (most recent call last):
  File "./mrchem", line 43, in <module>
    from parselglossy.api import parse
ModuleNotFoundError: No module named 'parselglossy' ```

**Comment by stigrj** [comment]
See if you can find "parselglossy" anywhere in `Pipfile.lock`, and perhaps try another `pipenv install --dev`. Do you get an error from
```
$ parselglossy --version
```

**Comment by gitpeterwind** [comment]
´´´
$ parselglossy --version
parselglossy, version 0.3.0
´´´
So, yes it is installed somewhere.

No parselglossy in Pipfile.lock.

**Comment by gitpeterwind** [comment]
I had a `module load Boost/1.63.0-intel-2017a-Python-2.7.13` in my .bashrc, which was causing the problems

**Comment by gitpeterwind** [comment]
According to cube format, the atoms lines should have the format
`int float 3xfloat`
i.e line 75 in MolPlotter.h should be something like:

`o << int(Z) << Z << coord[0] << "  " << coord[1] << "  " << coord[2] << std::endl;`

**Comment by gitpeterwind** [comment]
The oribtal plots are produced by individual MPI processes. Then we can get the "Exiting MRChem" message before everybody is finished (it take some time to write on disk). 
It would be good to add a barrier (also as a general feature) before the finalizes at the end of mrchem.cpp:
```
mpi::barrier(comm_orb);
mrenv::finalize(timer.elapsed());
mpi::finalize();```

gitpeterwind:
` o << static_cast<int>(Z) << Z << coord[0] << "  " << coord[1] << "  " << coord[2] << std::endl; `
trenger vel mellomrom mellow intZ og Z, og etter Z?

stigrj:
Doh :man_facepalming:

gitpeterwind:
> Doh man_facepalming

and while you are there: kan we print less decimals? The files can get big

PR #231, opened 2019-07-05
MRChem-0.2.0

stigrj:
This finalizes MRChem v0.2.0. This will basically be the same as v1.0.0, but I think we should wait until `parselglossy` is handled internally before we call it v1.0. Properties are currently limited to ground state energy and dipole moment, e.i. no response. This will be included once #197 is finished and we have all methods (including open-shell GGA).

New or updated input keywords are explained with a comment below:

```
world_prec = 1.0e-4     # Former rel_prec
world_size = 4          # Replaces the world part of former MRA, total size [2^4]^3
                        # always symmetric around the origin, e.i. [-8,8]^3 in this case
                        # A negative value will estimate the size based on molecule

Basis {                 # Replaces the basis part of former MRA
  order = 6
  type = Interpolating
}

Printer {
  print_level = 0         # Level of detail in the written output
  print_input = true      # Print the user input file at the start of the calculation
}

Precisions {
  nuc_prec = 1.0e-4       # Precision for smoothing and projection of nuclear potential
  poisson_prec = 1.0e-4   # Precision for construction of Poisson operators
  helmholtz_prec = 1.0e-4 # Precision for construction of Helmholtz operators
}

Properties {
  scf_energy = true
  dipole_moment = true
}

WaveFunction {
  method = B3LYP       # Default functionals can now be written here without DFT section
  restricted = true
}

Molecule {
  translate = true     # Translate center-of-mass to origin, i.e. z = +/-0.7
$coords
H   0.0   0.0   0.0
H   0.0   0.0   1.4
$end
}

SCF {
  kain = 3
  guess_prec = 1.0e-3  # Precision used in initial guess calculation
  start_prec = 1.0e-3  # Incremental precision in SCF iterations, initial value
  final_prec = 1.0e-4  # Incremental precision in SCF iterations, final value
  localize = false
  orbital_thrs = 1.0e-3
  property_thrs = 1.0e-5
}
```


doc/input_reference.rst
```diff
@@ -0,0 +1,439 @@
+.. raw:: html
```
robertodr: Have you tested locally that the documentation renders correctly? I think it does, but just to be extra sure.

external/upstream/fetch_mrcpp.cmake
```diff
@@ -9,9 +9,9 @@ else()
   FetchContent_Populate(mrcpp_sources
     QUIET
     GIT_REPOSITORY
```
robertodr: Apparently, using `GIT_REPOSITORY` can be a security issue. In Psi4 we switched to use this instead:
```
URL https://github.com/MolSSI/QCElemental/archive/v0.4.2.tar.gz
```
where you point directly to a release (or commit or tag) tarball. I never tried whether it works with `FetchContent` though.

stigrj: Seems to work, at least on Circle

PR #235, opened 2019-08-22
Linear response: Closed Shell and LDA Open Shell

ilfreddy:
Reactivating PR after rebase.

- [x] Check that nothing is broken (should be working up to GGA closed shell)
- [ ] Fix GGA open shell
- [ ] Make unit and integration tests

### DIFF
--- a/src/driver.cpp
+++ b/src/driver.cpp
@@ -370,7 +370,11 @@ bool driver::run_rsp(const json &json_rsp, Molecule &mol) {
         solver.setOrbitalPrec(start_prec, final_prec);
         solver.setThreshold(orbital_thrs, property_thrs);
 
+        mrcpp::print::header(1, "Unperturbed Fock operator");
+        Timer t_fock;
         F_0.setup(rsp_prec);
+        mrcpp::print::footer(1, t_fock, 2);
+
         for (int d = 0; d < 3; d++) {
             if (directions[d]) {
                 F_1.perturbation() = h_1[d];
@@ -495,8 +499,7 @@ void driver::plot_scf_quantities(const json &json_plot, Molecule &mol) {
 void driver::calc_scf_properties(const json &json_prop, Molecule &mol) {
     Timer t_tot, t_lap;
     auto plevel = Printer::getPrintLevel();
-    print_utils::headline(1, "Computing Ground State Properties");
-    if (plevel == 1) mrcpp::print::header(1, "Calculating Molecular Properties");
+    if (plevel == 1) mrcpp::print::header(1, "Computing Ground State Properties");
 
     auto &nuclei = mol.getNuclei();
     auto &Phi = mol.getOrbitals();
@@ -578,30 +581,35 @@ void driver::calc_scf_properties(const json &json_prop, Molecule &mol) {
  * input section, and will compute all properties which are present in this input.
  */
 void driver::calc_rsp_properties(const json &json_prop, Molecule &mol, int dir, double omega) {
+    Timer t_tot, t_lap;
+    auto plevel = Printer::getPrintLevel();
+    if (plevel == 1) mrcpp::print::header(1, "Computing Linear Response Properties");
+
     auto &Phi = mol.getOrbitals();
     auto &X = mol.getOrbitalsX();
     auto &Y = mol.getOrbitalsY();
 
     auto json_pol = json_prop.find("polarizability");
     if (json_pol != json_prop.end()) {
-        mrcpp::print::header(1, "Calculating polarizability");
+        t_lap.start();
+        mrcpp::print::header(2, "Calculating polarizability");
         auto prec = (*json_pol)["setup_prec"].get<double>();
         auto r_O = (*json_pol)["origin"].get<Coord<3>>();
 
         Polarizability &alpha = mol.getPolarizability(omega);
 
-        Timer timer;
         H_E_dip h(r_O);
         h.setup(prec);
         alpha.getTensor().row(dir) = -h.trace(Phi, X, Y).real();
         h.clear();
-        timer.stop();
-        mrcpp::print::footer(1, timer, 2);
+        mrcpp::print::footer(2, t_lap, 2);
+        if (plevel == 1) mrcpp::print::time(1, "Polarizability", t_lap);
     }
 
     auto json_mag = json_prop.find("magnetizability");
     if (json_mag != json_prop.end()) {
-        mrcpp::print::header(1, "Calculating paramagnetic magnetizability");
+        t_lap.start();
+        mrcpp::print::header(2, "Calculating paramagnetic magnetizability");
         auto prec = (*json_mag)["setup_prec"].get<double>();
         auto r_O = (*json_mag)["origin"].get<Coord<3>>();
         auto pert_diff = (*json_mag)["derivative"].get<std::string>();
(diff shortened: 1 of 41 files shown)


src/scf_solver/SCFSolver.cpp
```diff
@@ -250,6 +250,7 @@ void SCFSolver::printOrbitals(const DoubleVector &norms,
         o_row << std::setw(w6) << std::setprecision(pprec) << std::scientific << errors(i);
         o_row << std::setw(w5) << conv_i;
         println(2, o_row.str());
+        std::cout << Phi[i].real() << std::endl;
```
stigrj: If you want to keep this output you should do
```
if (Phi[i].hasReal()) println(0, Phi[i].real());
```

ilfreddy:
@stigrj All checks pass. Can you review this PR?

src/driver.cpp
```diff
@@ -664,7 +676,7 @@ void driver::build_fock_operator(const json &json_fock, Molecule &mol, FockOpera
             auto J_p = std::make_shared<CoulombOperator>(P_p, Phi_p, X_p, Y_p, shared_memory);
             F.getCoulombOperator() = J_p;
         } else {
-            MSG_ABORT("Invalid perturbation order");
+            MSG_ABORT("Invalid perturbation orde>r");
```
stigrj: ``if (orde > r)`` then what?

src/mrenum.h
```diff
@@ -0,0 +1,14 @@
...
+ *          UiT - The Arctic University of Norway
+ *
+ *  \breif Global enumerator structures
+ */
+
+#pragma once
+
+namespace DENSITY {
+enum class DensityType { Total, Spin, Alpha, Beta };
+}
```
stigrj: I think ``DENSITY::DensityType`` is "butter on bacon", as we say in Norwegian :bacon: :slightly_smiling_face:

tests/qmoperators/CMakeLists.txt
```diff
@@ -7,7 +7,7 @@ target_sources(mrchem-tests
     ${CMAKE_CURRENT_SOURCE_DIR}/nuclear_operator.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/coulomb_operator.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/coulomb_hessian.cpp
-    ${CMAKE_CURRENT_SOURCE_DIR}/xc_operator.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/xc_operator_lda.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/xc_operator_blyp.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/xc_hessian.cpp
```
stigrj: It really bugs me that this file is not called ``xc_hessian_lda.cpp`` :slightly_smiling_face:

src/mrenum.h
ilfreddy: How about `DENSITY::Type`?

tests/qmoperators/CMakeLists.txt
ilfreddy: I think you have been spending too much time with @robertodr lately... 
(I will change it)

src/mrenum.h
stigrj: I think the purpose of `enum class` is to replace the namespace, so `DensityType::` should be fine

src/scf_solver/SCFSolver.cpp
```diff
@@ -249,16 +250,28 @@ void SCFSolver::printOrbitals(const DoubleVector &norms,
         o_row << std::setw(w6) << std::setprecision(pprec) << std::scientific << errors(i);
         o_row << std::setw(w5) << conv_i;
         println(2, o_row.str());
+        if (Phi[i].hasReal()) println(0, Phi[i].real());
```
stigrj: I don't think belongs in any of the 'standard' printlevels in the mrchem output, but if you want to keep it you can make it printlevel 5.

stigrj: and yes, I could have suggested printlevel 5 in the first place, but I didn't realize at the time :clown_face:

tests/qmoperators/CMakeLists.txt
robertodr: Why is it always my fault?

src/qmoperators/two_electron/XCPotential.cpp
```diff
@@ -25,46 +25,109 @@ void XCPotential::setupDensity(double prec) {
...
-        print_utils::qmfunction(2, "XC beta density", rho_b, time_b);
-        rho_b.setReal(nullptr);
-
-        // Extend to union grid
+        buildDensity(Phi, DensityType::Alpha, prec);
+        buildDensity(Phi, DensityType::Beta, prec);
+        FunctionTree<3> &func_a = this->getDensity(DensityType::Alpha);
+        FunctionTree<3> &func_b = this->getDensity(DensityType::Beta);
+        println(5, static_cast<mrcpp::MWTree<3> &>(func_a));
+        println(5, static_cast<mrcpp::MWTree<3> &>(func_b));
```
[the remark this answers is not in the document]
stigrj: No longer necessary after MRCPP bugfix

src/driver.cpp
```diff
@@ -575,45 +581,51 @@ void driver::calc_scf_properties(const json &json_prop, Molecule &mol) {
...
         Polarizability &alpha = mol.getPolarizability(omega);
 
-        Timer timer;
         H_E_dip h(r_O);
         h.setup(prec);
         alpha.getTensor().row(dir) = -h.trace(Phi, X, Y).real();
         h.clear();
-        timer.stop();
-        mrcpp::print::footer(1, timer, 2);
+        mrcpp::print::footer(2, t_lap, 2);
```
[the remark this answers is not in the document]
stigrj: No longer necessary, it will just print the `elapsed()` time since the clock was started

src/input/template.yml
```diff
@@ -79,6 +79,11 @@ sections:
         docstring: |
           Precision parameter used in construction of Helmholtz operators.
           Negative value means it will follow the dynamic precision in SCF.
+      - name: orth_prec
+        type: float
+        default: 1.0e-14
+        docstring: |
+          Precision parameter used in the response density projection operator (1 - rho_0).
```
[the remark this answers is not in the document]
stigrj: For the time being this is not related to the regular MW precisions but rather to floating point precision. It is a threshold for what is considered orthogonal wrt `<i|j>`, which matters for reproducibility as the overlap for "orthogonal" functions might vary (irreproducibly) in the range 1.0e-12 - 1.0e-16, and then it is the question whether we should re-orthogonalize it or not.

It could possibly be relate to the MW precision, though, as `<i|j> = 1.0e-6` should be considered orthogonal when the MW precision is `prec = 1.0e-5`...

Anyway, a predicate could still be useful

src/qmfunctions/density_utils.cpp
```diff
@@ -115,11 +120,17 @@ void density::compute(double prec, Density &rho, OrbitalVector &Phi, int spin) {
...
+                      Density &rho,
+                      OrbitalVector &Phi,
+                      OrbitalVector &X,
+                      OrbitalVector &Y,
+                      DensityType spin) {
     int N_el = orbital::get_electron_number(Phi);
     double rel_prec = prec;        // prec for rho_i = |x_i><phi_i| + |phi_i><x_i|
     double abs_prec = prec / N_el; // prec for rho = sum_i rho_i
 
+    // LUCA: rho_loc should get the "general" grid in order to make sure all densities are available on the same grid.
```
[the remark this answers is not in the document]
ilfreddy: Removed

src/mrchem.h
```diff
@@ -39,10 +40,8 @@ const double sqrt_pi =    1.7724538509055160273;
 
 namespace SPIN { enum type { Paired, Alpha, Beta }; }
```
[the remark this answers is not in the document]
ilfreddy: This is rather annoying. In many places this is recast as an integer and if I use the enum class, all those are now flagging a compilation error. Do you have a simple way to fix this without changing all those instances (roughly 150) by hand?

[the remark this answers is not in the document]
ilfreddy: Not to mention all those cases where I would have to change `int spin` to `SpinType spin`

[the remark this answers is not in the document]
ilfreddy: In conclusion, if you don't deem this necessary, I would rather wait and focus on more pressing matters...

PR #238, opened 2019-09-06
Bank and norm overlap

gitpeterwind:
Cleaned up and ready to merge.
A Bank is defined to store remotely (through MPI) orbitals, functions or double arrays.
The number of MPI processes reserved for running the bank is set by bank_size (zero default, no bank is used) in the MPI section.
For now only the Helmolz application uses the bank, if bank_size>0.
Some reorganization in initializations, since the bank is part of MPI initializations, and it now needs to know the value of bank_size. So the order is:  init_mra, init_mpi, mpi::initialize(), init_printer.

Functions for calculating the norm overlap included. Uses pairwise common grid. Two method: exact, or uses only norm of nodes. Norm of nodes is much faster (20 seconds vs 570 seconds for C20H42) and gives same qualitative results. Example of differences in distribution of the norm_overlap values (C20H42):
norm of nodes method:
 between 1.0e-01 and 1.0e+00:   881 
 between 1.0e-02 and 1.0e-01:   1210 
 between 1.0e-03 and 1.0e-02:   1380 
 between 1.0e-04 and 1.0e-03:   1248
 between 1.0e-05 and 1.0e-04:   1196 
 between 1.0e-06 and 1.0e-05:   602 
 between 1.0e-07 and 1.0e-06:   44 
 between 1.0e-08 and 1.0e-07:   0 

exact method:
 between 1.0e-01 and 1.0e+00:   595 
 between 1.0e-02 and 1.0e-01:   1332
 between 1.0e-03 and 1.0e-02:   1280 
 between 1.0e-04 and 1.0e-03:   1320 
 between 1.0e-05 and 1.0e-04:   1206 
 between 1.0e-06 and 1.0e-05:   752 
 between 1.0e-07 and 1.0e-06:   76 
 between 1.0e-08 and 1.0e-07:   0


src/scf_solver/OrbitalOptimizer.cpp
```diff
@@ -25,6 +25,9 @@
 
 #include "MRCPP/Printer"
 #include "MRCPP/Timer"
+#include <chrono>
+#include <thread>
+#include <unistd.h>
```
stigrj: Are these includes used?

gitpeterwind: probably not!

gitpeterwind: How to link to the new mrcpp on github? (the tests fail). Also I did not manage to rebase because of conflicts

stigrj: Temporarily you can change the `GIT_TAG` in `external/upstream/fetch_mrcpp.cmake` from v1.0.1 to the latest commit hash. I will make a new patch v1.0.2 when https://github.com/MRChemSoft/mrcpp/pull/98 is merged.

stigrj: Actually, this involves new features so it will be v1.1.0

stigrj:
At first glance this looks great! I will have a closer look and try it out next week

src/scf_solver/OrbitalOptimizer.cpp
gitpeterwind: Thanks, now the tests have passed!

PR #242, opened 2019-10-04
New solvent

Gabrielgerez:
Solvent effect that has been brought up to date to the mrchem code. Less commits this time


src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,101 @@
...
+    this->d = slope;
+    this->abc = atom_based_cavity;
+    readCoordinateString(coord_str);
+}
+
 double Cavity::evalf(const mrcpp::Coord<3> &r) const {
     double C = 1.0;
     double s, O;
-
+    double val;
```
robertodr: I think this is unused

src/qmoperators/two_electron/ReactionPotential.cpp
```diff
@@ -0,0 +1,284 @@
...
+        setGamma(inv_eps_func, this->gamma, *V_tot_func, d_cavity);
+
+        poissonSolver(*rho_eff_func, &diff_func, &V_np1_func, &error);
+
+        if (iter > 1 and this->history > 0) { accelerateConvergence(diff_func, temp, kain); }
+
+        V_np1_func.free(NUMBER::Real);
+        qmfunction::add(V_np1_func, 1.0, temp, 1.0, diff_func, -1.0);
+        temp = V_np1_func;
+        std::cout << "error:\t" << error << "\n"
```
robertodr: Shouldn't `Printer` be used to write to the output file?

src/qmoperators/two_electron/ReactionPotential.cpp
```diff
@@ -0,0 +1,284 @@
...
+        this->d_coefficient = e_i - e_o;
+        setRhoEff(rho_eff_func, eps);
+
+    } else if (not is_lin) {
+        auto eps = [C_tmp, eps_i, eps_o](const mrcpp::Coord<3> &r) {
+            return eps_i * std::exp(std::log(eps_o / eps_i) * (1 - C_tmp.evalf(r)));
+        };
+        this->d_coefficient = std::log(e_i / e_o);
+        setRhoEff(rho_eff_func, eps);
+    }
```
robertodr: Are you sure you wrote the branching correctly here? It seems to me the `else if` is redundant?

src/qmoperators/two_electron/ReactionPotential.cpp
```diff
@@ -0,0 +1,284 @@
...
+            std::cout << "error:\t" << error << std::endl;
+        } else {
+            SCRF(&V_tot_func, &V_vac_func, &rho_eff_func, temp, inv_eps_func, d_cavity);
+        }
+    }
+
+    if (gammanp1.hasReal()) gammanp1.free(NUMBER::Real);
+    if (V_tot_func.hasReal()) V_tot_func.free(NUMBER::Real);
+    gammanp1.alloc(NUMBER::Real);
+    V_tot_func.alloc(NUMBER::Real);
```
robertodr: It's probably a good idea to have a `reset` method to abstract this pattern away and avoid boilerplate code.

src/qmoperators/two_electron/ReactionPotential.cpp
```diff
@@ -0,0 +1,284 @@
...
+        V_np1_func.free(NUMBER::Real);
+        qmfunction::add(V_np1_func, 1.0, temp, 1.0, diff_func, -1.0);
+        temp = V_np1_func;
+        std::cout << "error:\t" << error << "\n"
+                  << "iter.:\t" << iter << std::endl;
+        iter++;
+    }
+}
+
+void ReactionPotential::setup(double prec) {
```
robertodr: Is this method only called once when initialising the object? If yes, I think some of the code later on is redundant. If not, then the name of the method is a bit confusing.

src/qmoperators/two_electron/ReactionPotential.cpp
```diff
@@ -0,0 +1,284 @@
...
+            return eps_i * std::exp(std::log(eps_o / eps_i) * (1 - C_tmp.evalf(r)));
+        };
+        this->d_coefficient = std::log(e_i / e_o);
+        setRhoEff(rho_eff_func, eps);
+    }
+
+    V_vac_func.alloc(NUMBER::Real);
+    mrcpp::apply(prec, V_vac_func.real(), *poisson, rho_tot.real());
+
+    if (not temp.hasReal()) {
```
robertodr: See my initial comment on this method. If this is guaranteed to only be called once, then I don't see why the conditional with two branches is needed.

src/chemistry/Cavity.cpp
Gabrielgerez: I think so too

src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,101 @@
...
+            ss >> coord[0];
+            ss >> coord[1];
+            ss >> coord[2];
+            ss >> Rad;
+            this->pos.push_back(coord);
+            this->R.push_back(Rad);
+        }
+        this->alpha.assign(R.size(), 1.0);
+    }
+}
```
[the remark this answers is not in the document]
Gabrielgerez: No longer needed and thus removed. Everything is done in the parser level.

src/chemistry/Molecule.cpp
```diff
@@ -272,4 +272,9 @@ void Molecule::printProperties() const {
     }
 }
 
+void Molecule::initCavity(std::vector<mrcpp::Coord<3>> &coords, std::vector<double> &R, double slope) {
+    if (cavity != nullptr) MSG_ABORT("Cavity already initialized");
```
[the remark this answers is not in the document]
Gabrielgerez: so i can replace `(cavity != nullptr)` with  `(cavity)` and it will still give the same check?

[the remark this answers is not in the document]
robertodr: Yes: https://coliru.stacked-crooked.com/a/2ea2d3788ceac058

src/qmoperators/two_electron/ReactionPotential.cpp
Gabrielgerez: It is called on the first SCF iteration, but not every time a new `ReactionOperator` instance is created.
I run this 0th iteration once to create a potential which is carried on to the next SCF cycle. 
From what i gather a new `ReactionOperator` instance is created each SCF iteration, and i don't want to do a new 0th iteration everytime, so i just have a test to see if there already has been a SCF iteration before.

src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,101 @@
...
+
+bool Cavity::isVisibleAtScale(int scale, int nQuadPts) const {
+
+    double stdDeviation = d;
```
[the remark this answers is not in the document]
Gabrielgerez: Using d directly now

src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,101 @@
 #include <vector>
 
 #include "Cavity.h"
+#include "Element.h"
+#include "MRCPP/Printer" //testing only, remove when done
```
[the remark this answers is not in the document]
Gabrielgerez: was it `<vector>` that was unused?

src/qmoperators/two_electron/ReactionPotential.cpp
stigrj: All QM operators come with a pair of functions `setup(double prec)` and `clear()`. The `setup` will prepare the operator for application and the `clear` will undo. General workflow
```
QMOperator O;              // initialize operator with necessary parameters
for (each SCF cycle) {
    O.setup(current_prec); // prepare operator, e.g. compute potential V
    Orbital psi = O(phi);  // apply operator, e.g. multiply potential V*phi
    O.clear();             // undo the setup, e.g. delete potential V
}
```
This operator is a bit special since the first `setup` is different from later `setup`s.

src/qmoperators/two_electron/ReactionPotential.cpp
stigrj: There should be created only one `ReactionOperator` at the start of the calculation, but then the underlying potential should be re-computed at every iteration in the `setup`

src/chemistry/Cavity.cpp
```diff
@@ -28,27 +28,101 @@
...
+    double stdDeviation = d;
+    auto visibleScale = static_cast<int>(-std::floor(std::log2(nQuadPts * 2.0 * stdDeviation)));
+
+    if (scale < visibleScale) { return false; }
+
+    return true;
+}
+// maybe necessary to state value outside the boundaries of the cavity, that is
+// eps_o and eps_i
+bool Cavity::isZeroOnInterval(const double *a, const double *b) const {
```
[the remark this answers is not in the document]
stigrj: Yes, it should ideally be `std::array` but this is a virtual `override` from mrcpp so it has to be changed there first.

src/chemistry/Cavity.h
```diff
@@ -23,23 +23,30 @@
...
+    void changeRadius(double r) { this->R[0] = r; }
+
 protected:
     std::vector<mrcpp::Coord<3>> pos;
     std::vector<double> R;
     double d;
-};
 
+    bool isVisibleAtScale(int scale, int nQuadPts) const;
+    bool isZeroOnInterval(const double *a, const double *b) const;
```
stigrj: You should add a `override` to these functions

src/chemistry/Cavity.cpp
```diff
@@ -41,14 +42,37 @@ Cavity::Cavity(std::vector<mrcpp::Coord<3>> &coord, std::vector<double> &R, doub
 double Cavity::evalf(const mrcpp::Coord<3> &r) const {
     double C = 1.0;
     double s, O;
```
stigrj: You should rather declare `s` and `O` inside the loop to limit their scope

src/chemistry/Cavity.h
```diff
@@ -23,23 +23,30 @@
  * <https://mrchem.readthedocs.io/>
  */
 
+#pragma once
+
 #include "MRCPP/MWFunctions"
 #include <array>
+#include <cmath>
 #include <vector>
```
stigrj: Do you need all these includes? Try to remove or move to .cpp file

src/chemistry/Cavity.h
```diff
@@ -23,23 +23,30 @@
...
 namespace mrchem {
 
 class Cavity final : public mrcpp::RepresentableFunction<3> {
 public:
     Cavity(std::vector<mrcpp::Coord<3>> &coords, std::vector<double> &R, double slope);
     double evalf(const mrcpp::Coord<3> &r) const override;
 
+    std::vector<mrcpp::Coord<3>> getCoordinates() { return pos; }
+    std::vector<double> getRadius() { return R; }
+    void changeRadius(double r) { this->R[0] = r; }
```
stigrj: This will only change the first entry?

src/scf_solver/OrbitalOptimizer.cpp
```diff
@@ -118,8 +123,37 @@ bool OrbitalOptimizer::optimize(Molecule &mol, FockOperator &F) {
...
+
+            // variational implementation of solvent effect
+            gamma.QMFunction::operator=(Phi_n.back());
+            Phi_n.pop_back();
+            dgamma.QMFunction::operator=(dPhi_n.back());
+            dPhi_n.pop_back();
+
+            gammanp1.free(NUMBER::Real);
+            qmfunction::add(gammanp1, 1.0, dgamma, 1.0, gamma, -1.0);
+            F.getReactionOperator()->setGammanp1(gammanp1);
```
stigrj: We need to find a more elegant solution for this

src/chemistry/Cavity.cpp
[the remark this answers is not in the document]
Gabrielgerez: I think i have removed all the unused headers

src/chemistry/Cavity.cpp
Gabrielgerez: Done!

src/chemistry/Molecule.cpp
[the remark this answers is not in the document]
Gabrielgerez: ok

src/qmoperators/two_electron/ReactionPotential.cpp
Gabrielgerez: Added it in the newest commit

src/scf_solver/OrbitalOptimizer.cpp
Gabrielgerez: agree, @robertodr adviced to do the substraction in the potential, which is included in the latest commit, we still need to add it to the vector and back to the potential, as of now.

robertodr:
@Gabrielgerez you might want to rebase on top of #293 once it's in, to fix the error from the bot.

src/qmfunctions/orbital_utils.cpp
```diff
@@ -599,6 +599,7 @@ ComplexMatrix orbital::calc_norm_overlap_matrix(OrbitalVector &BraKet, bool exac
  * Computes the inverse square root of the orbital overlap matrix S^(-1/2)
  */
 ComplexMatrix orbital::calc_lowdin_matrix(OrbitalVector &Phi) {
+
```
robertodr, line 602: Remove this space.

src/chemistry/Cavity.cpp
```diff
@@ -22,33 +22,89 @@
...
+                double DCi = -(1.0 / (d * std::sqrt(MATHCONST::pi))) * std::exp(-std::pow(s / d, 2.0)) * ds;
+
+                double numerator = DCi;
+                double denominator = 1.0 - Ci;
+
+                if (((1.0 - Ci) < 1.0e-12) and ((1.0 - Ci) >= 0)) {
+                    denominator = 1.0e-12;
+                } else if (((1.0 - Ci) > -1.0e-12) and ((1.0 - Ci) <= 0)) {
+                    denominator = -1.0e-12;
+                }
```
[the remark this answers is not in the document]
Gabrielgerez: If i did not do that then the function would go towards infinity when the denominator was small enough. this at least seems to set a limit. I'm not so sure of its safety though

src/qmoperators/two_electron/ReactionOperator.h
```diff
@@ -0,0 +1,49 @@
...
+
+    ComplexDouble trace(OrbitalVector &Phi) { return RankZeroTensorOperator::trace(Phi); }
+
+    double &getTotalEnergy() { return this->potential->getTotalEnergy(); }
+    double &getElectronicEnergy() { return this->potential->getElectronicEnergy(); }
+    double &getNuclearEnergy() { return this->potential->getNuclearEnergy(); }
+    bool &getRunVariational() { return this->potential->getRunVariational(); }
+    QMFunction &getGamma() { return this->potential->getGamma(); }
+    QMFunction &getDiffFunc() { return this->potential->getDiffFunc(); }
+    QMFunction &getPotential() { return *this->potential; }
```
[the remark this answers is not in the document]
Gabrielgerez: If I remember correctly it was required by other methods in `FockOperator.cpp` and `Driver.cpp` at the time they were created. @stigrj is there a reason for this?

src/qmoperators/two_electron/ReactionPotential.cpp
```diff
@@ -0,0 +1,239 @@
...
+    diff_func.QMFunction::operator=(dPhi_n[0]);
+
+    phi_n.clear();
+    dPhi_n.clear();
+}
+
+void ReactionPotential::poissonSolver(QMFunction rho_eff_func,
+                                      QMFunction *diff_func,
+                                      QMFunction *V_np1_func,
+                                      double *error) {
```
robertodr, line 83: Why pass by pointer? Can you instead `return error;`?

src/qmoperators/two_electron/ReactionPotential.h
```diff
@@ -0,0 +1,79 @@
...
+                      double eps_i,
+                      double eps_o,
+                      bool islin);
+    ~ReactionPotential() override { free(NUMBER::Total); }
+    friend class ReactionOperator;
+
+    double &getTotalEnergy();
+    double &getElectronicEnergy();
+    double &getNuclearEnergy();
+    double &getElectronIn();
```
robertodr, line 29: Return by value, not by reference. And mark as `const`.

src/qmoperators/two_electron/ReactionPotential.h
```diff
@@ -0,0 +1,79 @@
...
+    QMFunction &getGamma() { return gamma; }
+    QMFunction &getDiffFunc() { return diff_func; }
+    bool &getRunVariational() { return variational; }
```
robertodr, line 32: Return by value and mark as `const`.

src/qmoperators/two_electron/ReactionPotential.h
```diff
@@ -0,0 +1,79 @@
...
+    int history;
+
+    double d_coefficient; // factor with which rescale the derivative of the cavity function
+    double electronicEnergy{0.0};
+    double nuclearEnergy{0.0};
+    double totalEnergy{0.0};
+    double electronsIn{0.0};
+    double e_i;       // permitivity of free space, 1 inside the cavity
+    double e_o;       // dielectric constant characteristic of the solvent
+    bool is_lin;      // determines if the dielectric function will be implemented linearly or exponentially
```
robertodr, line 66: I suggest you implement a permittivity profile class to make this part more general and more easily extended in the future.

src/qmoperators/two_electron/ReactionPotential.h
```diff
@@ -0,0 +1,79 @@
...
+    bool variational; // determines if the Reaction potential will be optimized in its own loop each SCF iteration or if
```
robertodr, line 67: The solver, variational or iterative, can be abstracted into two separate classes that perform the variational step or the microiterations. This is probably more generality than we need, but worth a thought.

src/scf_solver/GroundStateSolver.cpp
```diff
@@ -276,8 +280,32 @@ bool GroundStateSolver::optimize(Molecule &mol, FockOperator &F) {
...
+            // Employ KAIN accelerator
+            kain.accelerate(orb_prec, Phi_n, dPhi_n);
+
+            // variational implementation of solvent effect
+            V_r.QMFunction::operator=(Phi_n.back());
+            Phi_n.pop_back();
+            diff_func.QMFunction::operator=(dPhi_n.back());
+            dPhi_n.pop_back();
+
+            F.getReactionOperator()->setDiffFunc(diff_func);
```
robertodr, line 304: This looks rather contrived and will surely come back at us with a vengeance. 
What this does (correct me if wrong):
1. Push the potential at the current iterate for`V_r` and the difference with the previous iterate `diff_func` in the global KAIN history, which includes the orbitals.
2. Extrapolate.
3. Set extrapolated `V_r` and `diff_func` into the `ReactionOperator` object instance.
4. Pop them out off the global KAIN history.

Where is the KAIN history for the reaction potential?

src/input/template.yml
```diff
@@ -250,6 +250,40 @@ sections:
...
+      Define cavity.
+    keywords:
+      - name: epsilon_in
+        type: float
+        default: 1.0
+        docstring: |
+          Permitivity of vacuum.
+      - name: epsilon_out
+        type: float
+        default: 2.0
```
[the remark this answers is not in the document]
stigrj: But then the user has to set it, regardless if it's used or not?

src/input/template.yml
```diff
@@ -250,6 +250,40 @@ sections:
...
+      - name: angstrom
+        type: bool
+        default: false
+        docstring: |
+          Coordinates given in angstrom rather than bohr.
+      - name: coords
+        type: str
+        default: ""
+        docstring: |
+          Coordinates and radii in xyz format.
```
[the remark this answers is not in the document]
stigrj: Again, without a default every user will have to set this

src/input/template.yml
```diff
@@ -377,6 +411,11 @@ sections:
         default: false
         docstring: |
           Compute hyperfine coupling tensors.
+      - name: reaction_energy
```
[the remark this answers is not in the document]
stigrj: This is currently used to trigger the addition of the `ReactionPotential` to the `FockOperator`, right? So this should be moved either to the `Environment` section as a keyword that says "use this environment", or perhaps to the `WaveFunction` section
```
WaveFunction {
  method = HartreeFock
  environment = pcm          # or solvent = true or whatever
  external_field = electric  # <- this is currently not in this section, but
}                            # could be more appropriate to add here as well
```

src/driver.cpp
```diff
@@ -132,7 +134,15 @@ void driver::init_molecule(const json &json_mol, Molecule &mol) {
         auto xyz = coord["xyz"].get<mrcpp::Coord<3>>();
         nuclei.push_back(atom, xyz);
     }
+    std::vector<double> radii;
+    std::vector<mrcpp::Coord<3>> spheres;
+    for (const auto &coord : json_mol["cavity_coords"].get<json>()) {
+        radii.push_back(coord["radius"].get<double>());
+        spheres.push_back(coord["center"].get<mrcpp::Coord<3>>());
+    }
+    auto width = json_mol["cav_width"].get<double>();
```
stigrj, line 143: Use full name `cavity_width`

src/input/template.yml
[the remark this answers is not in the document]
robertodr: True that...

src/input/template.yml
```diff
@@ -716,3 +755,18 @@ sections:
...
+  - name: SolventEffect
+    docstring: |
+      Includes parameters related to the computation of the reaction field
+      energy of a solvent system.
+    keywords:
+      - name: variational
+        type: bool
+        default: false
+        docstring: |
+          Run a variational computation of the reaction field energy.
```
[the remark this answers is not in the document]
Gabrielgerez: This is done now

src/input/template.yml
```diff
@@ -716,3 +755,18 @@ sections:
         docstring: |
           Path to where converged orbitals will be written in connection with
           the 'write_orbitals' keyword.
+  - name: SolventEffect
```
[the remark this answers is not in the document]
Gabrielgerez: its done

src/input/template.yml
```diff
@@ -716,3 +755,18 @@ sections:
...
+      - name: variational
+        type: bool
+        default: false
+        docstring: |
+          Run a variational computation of the reaction field energy.
+      - name: kain
+        type: int
+        default: 0
+        docstring: |
+          History for convergence accelerator
```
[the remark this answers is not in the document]
Gabrielgerez: added this to the latest commit

src/qmoperators/two_electron/ReactionOperator.h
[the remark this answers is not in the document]
Gabrielgerez: seems like it wasn't needed anyways, at least for now. All of the return by reference methods here have been changed.

Gabrielgerez:
Have now separated the nested procedure from the variational procedure. 
TODO:
- Rearrange Cavity.cpp
- Set tests for solvent

src/chemistry/Cavity.cpp
```diff
@@ -22,33 +22,89 @@
...
-    this->d = slope;
+Cavity::Cavity(std::vector<mrcpp::Coord<3>> &coord, std::vector<double> &R, double slope)
+        : pos(coord)
+        , R(R)
+        , d(slope) {
+    setGradVector();
+}
+
+void Cavity::setGradVector() {
+    for (int j = 0; j < 3; j++) {
```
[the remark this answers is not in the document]
Gabrielgerez: wouldn't the loop at the end just not run?
`this->gradvector` has size 0 at the start of the loop. The size of the vector should be 3 when it is finished, as it is a 3D gradient of a scalar function with three arguments.

external/upstream/fetch_mrcpp.cmake
```diff
@@ -42,7 +42,7 @@ else()
 
   FetchContent_GetProperties(mrcpp_sources)
 
-  set(CMAKE_BUILD_TYPE Release)
+  #set(CMAKE_BUILD_TYPE Release)
```
stigrj, line 45: Why commented? I guess it doesn't matter since `Release` is default, but still

src/qmoperators/two_electron/ReactionPotential.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
+#include "chemistry/Cavity.h"
+#include "chemistry/Nucleus.h"
+#include "qmfunctions/Density.h"
+#include "qmfunctions/Orbital.h"
+#include "qmoperators/one_electron/QMPotential.h"
+#include "qmoperators/two_electron/SCRF.h"
+
+using OrbitalVector_p = std::shared_ptr<mrchem::OrbitalVector>;
```
stigrj, line 9: Please don't use this in headers

src/qmoperators/two_electron/SCRF.h
```diff
@@ -0,0 +1,91 @@
...
+#include "chemistry/Permittivity.h"
+#include "qmfunctions/Density.h"
+#include "qmfunctions/Orbital.h"
+#include "qmfunctions/QMFunction.h"
+#include "qmfunctions/qmfunction_fwd.h"
+#include "scf_solver/KAIN.h"
+
+using PoissonOperator_p = std::shared_ptr<mrcpp::PoissonOperator>;
+using DerivativeOperator_p = std::shared_ptr<mrcpp::DerivativeOperator<3>>;
+using OrbitalVector_p = std::shared_ptr<mrchem::OrbitalVector>;
```
stigrj, line 13: remove from header

stigrj (changes requested):
Can you remove _all_ changes in the `pilot` directory?

I think we should modularize and move the following files to a new `solvent` or `environment` directory, i.e. leave only the `ReactionPotential`:
- SCRF
- Permittivity
- Cavity

Opinion on the name @robertodr? `environment` is more general, but can be confused with build environment, and we already have a `mrenv` file which initializes the MR world. 

We also need to think carefully about the input parameters before this is officially released, but that can wait for now.

stigrj:
Looks like the MPI build fails. I will try to debug

external/upstream/fetch_mrcpp.cmake
robertodr: This might have been I suggestion I made to run the debugger and get sensible backtraces.

robertodr:
I agree with reorganizing files. As for the name of the new folder, I prefer `environment`.

stigrj:
I can't find any MPI errors locally, so probably just a CircleCI glitch

stigrj:
@robertodr can you tell what's wrong with Minazo? Looks like it can't find the PR branch

robertodr:
> @robertodr can you tell what's wrong with Minazo? Looks like it can't find the PR branch

I don't know what's wrong, it could be a problem with the secret token. Please disregard it for the moment being.

src/scf_solver/KAIN.cpp
```diff
@@ -85,7 +85,8 @@ void KAIN::setupLinearSystem() {
                 orbB(i) += orbital::dot(dPhi_im, fPhi_m);
             }
         }
-        A_matrices.push_back(orbA);
+        double alpha = (this->scaling.size() == nOrbitals) ? scaling[n] : 1.0;
+        A_matrices.push_back(orbA * alpha);
```
robertodr, line 89: Are we using this at all for the reaction potential? If yes, is there a way to set the scaling parameter from the input?

robertodr (changes requested):
Some more changes needed before this goes in.

When including header files follow this order:
1. Headers from the standard library, *i.e.* `#include <string>`
2. Headers from external libraries, *i.e.* `#include <Eigen/Dense>` (or MRCPP or nlhomann_json or XCFun)
3. Headers from the current project, *i.e.* `#include "chemistry/Nucleus.h"`
4. Headers from the current folder, *i.e.* `#include "FockOperator.h"` (if in `qmoperators/two_electron` for example)

The only exception to this ordering is in the implementation file of a class. Then the corresponding header **must** come first.

Another point to consider in header files is to forward-declarer whenever possible, to reduce compilation times. See here for a guide: https://pcmsolver.readthedocs.io/en/latest/programmers/coding-standards.html#including-header-files

pilot/mrchem.cpp
```diff
@@ -3,9 +3,23 @@
 #include <MRCPP/Printer>
 #include <MRCPP/Timer>
 
+#include "MRCPP/MWFunctions"
+#include "MRCPP/MWOperators"
```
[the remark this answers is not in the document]
stigrj: All the `pilot/` commits should be removed from this PR

external/upstream/fetch_mrcpp.cmake
Gabrielgerez: ok

src/qmoperators/two_electron/ReactionPotential.h
Gabrielgerez: do you mean the alias or the includes?

robertodr: The alias. It will be visible in any other files that does `#include "ReactionPotential.h"` You can put it in the `public` part of the class and you wouldn't need to modify the cpp file (I think!)

python/template.yml
```diff
@@ -682,3 +682,74 @@ sections:
...
+        type: bool
+        default: true
+        docstring: |
+          Adjust the convergence threshold for the nested procedure with regards to the orbital residual.
+          When the orbitals are close to convergence the convergence threshold will be equal to the orbital threshold.
+      - name: extrapolate_Vr
+        type: bool
+        default: true
+        docstring: |
+          Extrapolate on the reaction potential if true, or on the surface charge distribution in the convergence acceleration.
```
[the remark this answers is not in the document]
Gabrielgerez: yes, but i'm not sure what result that would give

python/template.yml
```diff
@@ -682,3 +682,74 @@ sections:
...
+          - name: epsilon_out
+            type: float
+            default: 2.0
+            docstring: |
+              Permittivity outside the cavity.
+          - name: linear
+            type: bool
+            default: false
+            docstring: |
+              Run linear formulation of the Permittivity function, false runs exponential.
```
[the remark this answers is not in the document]
Gabrielgerez: The linear option is deprecated, I'll remove this option as the only one used now is exponential. 
Alternatively, I could keep this and modify it to a string with only exponential available, in case of future development

python/template.yml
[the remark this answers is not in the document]
robertodr: Then it's maybe best to remove the option completely.

python/template.yml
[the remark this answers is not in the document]
robertodr: Let's do the string thing, for future-proofing.

pilot/mrchem.cpp
```diff
@@ -3,9 +3,23 @@
...
+
+#include "analyticfunctions/HydrogenFunction.h"
+#include "chemistry/Cavity.h"
+#include "chemistry/Element.h"
+#include "chemistry/Nucleus.h"
+#include "chemistry/PeriodicTable.h"
+#include "chemistry/Permittivity.h"
```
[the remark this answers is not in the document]
Gabrielgerez: most of them weren't needed

stigrj:
The `ReactionOperator` and `ReactionPotential` should still be in `qmoperators` :slightly_smiling_face: The `environment` should contain everything needed to produce the `ReactionPotential`: `Cavity`, `Permittivity` and `SCRF` for now

python/template.yml
```diff
@@ -682,3 +682,74 @@ sections:
...
+          parameters for the permittivity function.
+        keywords:
+          - name: epsilon_in
+            type: float
+            default: 1.0
+            docstring: |
+              Permittivity inside the cavity.
+          - name: epsilon_out
+            type: float
+            default: 2.0
```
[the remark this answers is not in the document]
Gabrielgerez: It seems a bit of bugs appeared when removing the default for the permittivity outside the cavity, which i found hard to solve. It is back in now :/
I agree as well that it would be better to not have a default

python/template.yml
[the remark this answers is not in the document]
Gabrielgerez: done

pilot/mrchem.cpp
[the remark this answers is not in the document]
Gabrielgerez: I think all of them are gone now, and I have changed all external libraries as Roberto suggested

pilot/mrchem.cpp
```diff
@@ -3,9 +3,23 @@
...
 #include "mrchem.h"
 #include "mrenv.h"
 #include "parallel.h"
+#include "qmfunctions/Orbital.h"
+#include "qmfunctions/orbital_utils.h"
+#include "qmfunctions/qmfunction_utils.h"
+#include "qmoperators/two_electron/ReactionOperator.h"
+#include "qmoperators/two_electron/SCRF.h"
```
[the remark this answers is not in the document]
Gabrielgerez: they weren't needed, they are gone now :)

src/chemistry/Permittivity.h
```diff
@@ -0,0 +1,80 @@
...
+ */
+
+class Cavity;
+
+class Permittivity final : public mrcpp::RepresentableFunction<3> {
+public:
+    Permittivity(const Cavity cavity, double epsilon_in, double epsilon_out = 1.0); //!< Standard constructor.
+    double evalf(const mrcpp::Coord<3> &r) const override;
+
+    /** @brief Changes the value of #inverse. */
```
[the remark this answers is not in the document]
Gabrielgerez: which function? the constructor, evalf or flipFunction?

src/scf_solver/KAIN.cpp
Gabrielgerez: We are not using it on the internal loop.

src/chemistry/Cavity.cpp
```diff
@@ -22,33 +22,119 @@
...
+/**  @relates mrchem::Cavity
+ *   @brief Constructs a single element of the gradient of the Cavity.
+ *
+ *   This constructs the analytical partial derivative of the Cavity \f$C\f$ with respect to \f$x\f$, \f$y\f$ or \f$z\f$
+ * coordinates and evaluates it at a point \f$\mathbf{r}\f$. This is given for \f$x\f$ by \f[ \frac{\partial
+ * C\left(\mathbf{r}\right)}{\partial x} = \left(1 - C{\left(\mathbf{r} \right)}\right) \sum_{i=1}^{N} - \frac{\left(x
+ * -{x}_{i}\right) e^{- \frac{\operatorname{s_{i}}^{2}{\left(\mathbf{r}  \right)}}{\sigma^{2}}}}{\sqrt{\pi} \sigma
+ * \left(0.5 \operatorname{erf}{\left(\frac{\operatorname{s_{i}}{\left(\mathbf{r}  \right)}}{\sigma} \right)} +
+ * 0.5\right) \left| \mathbf{r} - \mathbf{r}_{i} \right|} \f] where the subscript \f$i\f$ is the index related to each
+ * sphere in the cavity, and \f$\operatorname{s}\f$ is the signed normal distance from the surface of each sphere.
```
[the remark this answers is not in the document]
Gabrielgerez: Is it better now?

python/template.yml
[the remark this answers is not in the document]
robertodr: Can you show what happens when you remove the default? This should be fixed properly.

src/scf_solver/KAIN.cpp
robertodr: If it's not used, should it then be removed completely? It makes the code harder to understand.

python/mrchem/api.py
```diff
@@ -154,11 +154,46 @@ def write_molecule(user_dict, origin):
...
+    cav_coords_dict = []
+    if len(user_dict["Environment"]["Cavity"]["spheres"]) > 0:
+        cav_coords_raw = user_dict["Environment"]["Cavity"]["spheres"]
+        for line in cav_coords_raw.split('\n'):
+            sp = line.split()
+            if len(sp) > 0:
+                center = list(map(float, sp[:-1]))
+                radius = float(sp[-1])
+                if len(center) != 3:
+                    print("Invalid coordinate: " +  " " + str(center))
```
robertodr, line 167: Use an f-string here: `print(f"Invalid coordinate {center}")

python/template.yml
```diff
@@ -682,3 +682,99 @@ sections:
...
+        type: bool
+        default: false
+        docstring: |
+          Perform the reaction field calculation of the reaction potential of the
+          interaction between environment and molecule. 
+      - name: algorithm
+        type: str
+        default: "scrf"
+        predicates:
+            - "value.lower() in ['scrf', 'variational']"
```
robertodr, line 705: Should `'variational'` be here?

src/chemistry/chemistry_utils.h
```diff
@@ -26,11 +26,13 @@
 #pragma once
 
 #include "chemistry_fwd.h"
+#include "qmfunctions/qmfunction_fwd.h"
 
 namespace mrchem {
 namespace chemistry {
 
 double compute_nuclear_repulsion(const Nuclei &nucs);
+Density compute_nuclear_density(double prec, const Nuclei &nucs, double alpha);
```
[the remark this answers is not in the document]
robertodr: This function is still undocumented

python/template.yml
Gabrielgerez: It shouldn't be there

src/environment/Permittivity.h
```diff
@@ -0,0 +1,84 @@
...
+    auto getGradVector() const { return this->cavity.getGradVector(); }
+
+    /** @brief Returns the value of #epsilon_in. */
+    auto getEpsIn() const { return this->epsilon_in; }
+
+    /** @brief Returns the value of #epsilon_out. */
+    auto getEpsOut() const { return this->epsilon_out; }
+
+private:
+    bool inverse = false;    //!< State of #evalf
```
robertodr, line 77: I'm being really dense, but can you explain what `inverse` is supposed to achieve? I'm a bit worried of this kind of stateful object...

robertodr (commented):
I've left some more comments. We're really close to merging this :smile_cat: Did you try running a test with a classical charge only that @ilfreddy suggested?

src/environment/Permittivity.h
Gabrielgerez: if `false` evalf returns epsilon(r)
if `true` evalf returns 1/epsilon(r)

Gabrielgerez: @stigrj mentioned once that I could avoid this by using a map method, but i'm not sure how it was implemented

Gabrielgerez:
Not yet. I had actually forgotten about that suggestion :sweat_smile: I can include some options to the code for this types of test

src/chemistry/Permittivity.h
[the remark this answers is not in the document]
Gabrielgerez: I've put a more extended explanation of the function now

stigrj:
Ok, it's about time to get this one in. Some of this will likely change a bit in the future, but I think we should just merge it. One last question: how do you scale the vdW radius? Shouldn't there be a simplified input for this, if you want to use vdW spheres, but with scaled radius?

Gabrielgerez:
> Ok, it's about time to get this one in. Some of this will likely change a bit in the future, but I think we should just merge it. One last question: how do you scale the vdW radius? Shouldn't there be a simplified input for this, if you want to use vdW spheres, but with scaled radius?

The cavity coordinates are either explicitly given (both position and radius size) by the user, or are constructed from the atomic coordinates with the vdw-radii from the periodic table.
the scaling happens per sphere radius, that is, multiply each radius by the scaling factor. This is done by the user beforehand.
There is no scaling factor in the code yet, but I think it might be a good idea to add a scaling parameter in the input, although it might increase confusion in the event of both cavity coordinates and scaling factor given.

Gabrielgerez:
@stigrj @robertodr What do you guys think, should we merge it as it is?

stigrj (approved):

stigrj:
Sorry, got distracted on friday and forgot about it :slightly_smiling_face:

stigrj:
Needs approval by @robertodr as well, it seems

robertodr (approved):
Very good!

PR #243, opened 2019-10-21
Periodic branch

bjorgve:
Working implementation of Luca's far-field approach and RC approach from Pask et. al. 2018.

The best way to run the code is to do a non-periodic calculation on the desired system then use that as
a starting guess for the periodic calculation.

To add:

- [ ]  Automatically run non-periodic starting guess 
- [ ] Auto-shift of starting guess, this is needed for starting guesses where the nucleis are placed on the cell-boundary
- [ ] Periodic integration tests

A few of the non-periodic tests fail, since the new result is not numerically the same as the previous version of the code. But they are correct within the requested precision. I have not figured out why it is not numerically the same.


src/driver.cpp
```diff
@@ -657,12 +658,17 @@ void driver::build_fock_operator(const json &json_fock, Molecule &mol, FockOpera
...
             auto J_p = std::make_shared<CoulombOperator>(P_p, Phi_p, shared_memory);
             F.getCoulombOperator() = J_p;
-        } else if (order == 1) {
+        } else if (order == 1 and not periodic) {
             auto J_p = std::make_shared<CoulombOperator>(P_p, Phi_p, X_p, Y_p, shared_memory);
             F.getCoulombOperator() = J_p;
+        } else if (periodic) {
+            auto nuc_prec = (*json_nuclear)["nuc_prec"].get<double>();
+            // auto J_p = std::make_shared<CoulombOperator>(P_p, Phi_p, nuclei, nuc_prec);
+            auto J_p = std::make_shared<CoulombOperator>(P_p);
```
stigrj: So you mean it segfaults if you uncomment 669? Line 670 will fail later unless you manually compute the density before `CoulombOperator::setup()`.

stigrj:
Try to run c++ executable manually as
```
mrchem -D mrchem.inp
mrchem.x mrchem.json
```
and see if some error message appears. Sometimes the `MSG_ERROR` `MSG_ABORT` messages are not printed when you run
```
mrchem mrchem.inp
```

bjorgve:
The following tests appear to fail:

```
 3/22 Test #19: H2_static_polarizability_LDA .....***Failed   12.02 sec
      Start 20: H2O_energy_BLYP
 4/22 Test #20: H2O_energy_BLYP ..................***Failed   82.23 sec
      Start 21: Li_SCF_PBE0
 5/22 Test #21: Li_SCF_PBE0 ......................***Failed   58.44 sec
      Start 22: Li_static_polarizability_LDA
 6/22 Test #22: Li_static_polarizability_LDA .....***Failed   45.93 sec

```

Looking at the output it seems like the answers are still correct, only lacking in precision. 

I don't understand why these tests are affected, as far as I can see I haven't touched code that can affect them.

PR #246, opened 2019-10-30
z function implemented

ilfreddy:
This is the LR version implementing the logarithmic ground state density to compute the gradient which is used by XCFun to get the functional derivatives.

This fix simply does the following:
1. computes zeta = log(rho)
2. computes \grad rho = rho \grad zeta

Everything else is unchanged in the LR algorithm

WARNING: this comes on top of the XCFun derivatives storage which is not relevant.
The relevant commits are from the following one:
4179738153a939c1c6289058942111bfbc531f56


stigrj:
There must be a missing `rebase` somewhere

stigrj:
The `zeta = log(rho)` implementation has been included in #248

PR #249, opened 2019-11-18
MPI fixes

stigrj:
- Allow oversubscription in `mpirun` of tests, so that `mpirun -np 5` works on a dual core laptop.
- `MPI_DOUBLE_COMPLEX` -> `MPI_CXX_DOUBLE_COMPLEX`. The former fails in some OpenMPI implementations.

### DIFF
--- a/src/parallel.cpp
+++ b/src/parallel.cpp
@@ -147,7 +147,7 @@ void mpi::allreduce_vector(DoubleVector &vec, MPI_Comm comm) {
 void mpi::allreduce_vector(ComplexVector &vec, MPI_Comm comm) {
 #ifdef HAVE_MPI
     int N = vec.size();
-    MPI_Allreduce(MPI_IN_PLACE, vec.data(), N, MPI_DOUBLE_COMPLEX, MPI_SUM, comm);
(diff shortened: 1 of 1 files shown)


gitpeterwind:
The `MPI_CXX_DOUBLE_COMPLEX` is fine!
  `-oversubscribe` is not always a valid flag (it is not on iimpi/2018b on Stallo for example).
On the other hand I thought it was always allowed to oversubscribe.

stigrj:
Hmm, no on my laptop I get (OpenMPI 3.1.3)
```
$ mpirun -np 5 echo "foo"
--------------------------------------------------------------------------
There are not enough slots available in the system to satisfy the 5 slots
that were requested by the application:
  echo

Either request fewer slots for your application, or make more slots available
for use.
--------------------------------------------------------------------------
```

stigrj:
Looks like `--oversubscribe` only exist for OpenMPI. With Intel you are always allowed to oversubscribe, and with OpenMPI you get this default by setting the variable
```
export OMPI_MCA_rmaps_base_oversubscribe=1
```
so I have removed the oversubscription part.

PR #251, opened 2019-11-26
Bug fix #244: Core initial guess

### LINKED ISSUE #244: Core Hamiltonian initial guess fails badly for open-shell

stigrj:
Fixing issue #244. Problem was not open-shell but rather that the core guess orbitals were not correctly distributed with MPI. Moved the `initial_guess::rotate_orbitals` from sad.cpp to core.cpp since the latter is more fundamental.

- [x] Ready to merge

### DIFF
--- a/src/initial_guess/core.cpp
+++ b/src/initial_guess/core.cpp
@@ -39,6 +39,7 @@
 #include "chemistry/Molecule.h"
 #include "chemistry/Nucleus.h"
 #include "qmfunctions/Orbital.h"
+#include "qmfunctions/OrbitalIterator.h"
 #include "qmfunctions/orbital_utils.h"
 #include "qmfunctions/qmfunction_utils.h"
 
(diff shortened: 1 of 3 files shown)


src/initial_guess/core.cpp
```diff
@@ -276,4 +256,32 @@ OrbitalVector initial_guess::core::project_ao(double prec, const Nuclei &nucs, i
     return Phi;
 }
 
+OrbitalVector initial_guess::core::rotate_orbitals(double prec, ComplexMatrix &U, OrbitalVector &Phi, int N, int spin) {
```
robertodr: Shouldn't this belong in `Orbital.{h,cpp}`?

stigrj: Yes, there is a `orbital::rotate` as well, but the current version is a bit special for the initial guess generation as it populates the output vector differently than the input. I will try to merge the two rotations and use only one.

robertodr: Thanks for the clarification!

PR #259, opened 2020-01-23
Derivative of the exact exchange operator

ilfreddy:
Implementation of the linear response version of the Exchange operator. It follows closely the implementation for the Coulomb operator. The challenge is that the Exchange operator requires the full density matrix rather then the density function.

- [x] Refactor ExchangePotential in `ExchangePotential` and `ExchangePotentialD1`
- [x] Create `ExchangePotentialD2`
- [x] Remove ij-ji symmetry to make the two implementations (D1 and D2) as similar as possible
- [x] Implement the core methods in `ExchangePotentialD2`
- [x] Enable Exchange in response calculations
- [ ] Complete the features of the class
- [ ] Fix renormalization in computation of exchange
- [ ] Fix unit tests (they are too long)
- [x] Test more orbitals 
- [x] Test higher precision 
- [x] Test open shell systems
- [X] Make unit tests
- [ ] Make integration tests
- [ ] Write/rewrite documentation in all `Exchange` classes


stigrj:
Seems to work. Now the error was due to timeout, which happens every now and then

ilfreddy:
Cont. Int. also times out. On the exchange_operator test. I guess there is something fishy there...

gitpeterwind:
Note that my next exchange Potential calculation (https://github.com/MRChemSoft/mrchem/pull/260), will use a different way of organizing the different parts. It should be possible to put more in common for D1 and D2. Therefore we should not spend to much time documenting the different classes at this stage.

stigrj:
Try to add `travis_wait` to the `ctest` command in `.travis.yml`: https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received
```
travis_wait ctest --output-on-failure --verbose
```
The other Travis fail is due to style violations

stigrj:
`WARNING`: Did a rebase to clean up some messy history, do NOT merge this with your own hess_exchange branch.

I think this is ready to merge once the tests passes.

ilfreddy:
> `WARNING`: Did a rebase to clean up some messy history, do NOT merge this with your own hess_exchange branch.
> 
> I think this is ready to merge once the tests passes.

It just failed again :-(  

What shall I do with the `hess_exchange` I have on my laptop(s)? I was updating the documentation. Should I just do it later in a separate branch?

Pipfile
```diff
@@ -6,6 +6,7 @@ name = "pypi"
 [packages]
 parselglossy = "*"
 runtest = "*"
+pyparsing = "*"
```
robertodr: I'm not sure this should be here. `pyparsing` is not used directly anywhere in MRChem

ilfreddy: Not sure where this comes from. @gitpeterwind, did you introduce this one?

gitpeterwind: Never touch this (at least that I am aware of)

ilfreddy: I seem to recall it came from your PR to my fork. But we can happily remove it if it is not needed.

ilfreddy: It comes from this one:
https://github.com/ilfreddy/mrchem/commit/6a73627759277db9afb0ec6af1e30608128067f0#diff-1e61a31bf9b94805f869dc4137ec1885

ilfreddy:
This is not clear to me. I have added two brand new unit tests, still the code coverage goes down. Is it only calculated with the `debug` job?

stigrj:
Yes, coverage is done only in `debug` as it gives the most accurate description relative to the actual source code

ilfreddy:
This can go for my part now. Only `exchange_hessian` is disabled to make sure the `debug` travis job finishes within 50min. Once we agree on a better strategy, we can enable it again.

Pipfile
robertodr: But is it really needed?

robertodr:
> This can go for my part now. Only `exchange_hessian` is disabled to make sure the `debug` travis job finishes within 50min. Once we agree on a better strategy, we can enable it again.

#262 Probably helps with this. Let's see.

src/qmoperators/two_electron/ExchangePotentialD2.h
```diff
@@ -0,0 +1,60 @@
...
+
+#include "ExchangePotential.h"
+#include "qmfunctions/qmfunction_fwd.h"
+#include "qmoperators/QMOperator.h"
+
+namespace mrchem {
+
+/** @class ExchangePotentialD2
+ *
+ *  @brief Hartree-Fock exchange potential defined by a set of unperturbed orbitals
```
robertodr: Class documentation is not correct, I think?

ilfreddy: That's right, but @stigrj had to rebase the branch (see comment above) and I am therefore postponing the documentation commit, which is on my local branch, to avoid another rebase mess. I will cherry-pick it and put it in another branch once this is merged.

stigrj: I think you can just cherry-pick right away if you have already finished the docs

ilfreddy: It is not finished. I just postponed finalizing it to get done with this PR first :-)

Pipfile
ilfreddy: If not, I can remove it. No problem.

ilfreddy:
Rebased and ready to go. @stig @robertodr

PR #260, opened 2020-02-11
New poisson

gitpeterwind:
Centered around `ExchangePotential::calc_i_Int_jk_P` which is much more effective than the previous method for computing the exchange contributions. 
It use adaptive precision for Poisson application and the new mrccp multiplication, that only use the refinements of tree where it is needed. Bank is used to allow asynchronous calculation of the exchange contributions (i.e. the MPI processes do not have to talk directly until they are done with all their exchange calculations).
The `calc_i_Int_jk_P` is general, so that it can be used for other Poisson applications too. Should it be moved somewhere else?


gitpeterwind:
The build fails, because it requires the updated version of mrcpp. Otherwise it is ready for review.

src/utils/RRMaximizer.cpp
```diff
@@ -44,7 +44,6 @@ RRMaximizer::RRMaximizer(double prec, OrbitalVector &Phi) {
     this->total_U = DoubleMatrix::Identity(this->N, this->N);
     this->N2h = this->N * (this->N - 1) / 2;
     this->gradient = DoubleVector(this->N2h);
-    this->hessian = DoubleMatrix(this->N2h, this->N2h);
```
stigrj, line 47: So we don't compute/need the Hessian at all? Did the algorithm change, or was the Hessian never actually in use?

src/parallel.cpp
```diff
@@ -113,7 +116,11 @@ void mpi::initialize() {
 
 void mpi::finalize() {
 #ifdef HAVE_MPI
-    if (mpi::bank_size > 0 and mpi::grand_master()) mpi::orb_bank.close();
+    if (mpi::bank_size > 0 and mpi::grand_master()){
+        auto plevel = Printer::getPrintLevel();
+        if (plevel > 1) std::cout<<" max data in bank "<<mpi::orb_bank.get_maxtotalsize()<<" MB "<<std::endl;
```
stigrj, line 121: ` println(1, " max data in bank " << mpi::orb_bank.get_maxtotalsize() << " MB");`

src/parallel.cpp
```diff
@@ -628,6 +651,22 @@ void Bank::close() {
...
+    int maxtot=0;
+    int datasize;
+    for (int i = 0; i < mpi::bank_size; i++) {
+        MPI_Send(&GETMAXTOTDATA, 1, MPI_INTEGER, mpi::bankmaster[i], 0, mpi::comm_bank);
+        MPI_Recv(&datasize, 1, MPI_INTEGER, mpi::bankmaster[i], 1171, mpi::comm_bank, &status);
+        maxtot = std::max(maxtot, datasize);
+    }
+    return maxtot;
+#endif
+}
```
stigrj, line 666: This generates a compiler warning for missing return statement when MPI is deactivated. Should perhaps return 0 or -1 by default?

src/qmoperators/two_electron/ExchangePotential.cpp
```diff
@@ -157,13 +349,7 @@ Orbital ExchangePotential::dagger(Orbital inp) {
...
 double ExchangePotential::getScaledPrecision(int i, int j) const {
-    double scaled_prec = this->apply_prec;
-    if (this->screen) {
-        double tNorm = this->tot_norms(i);
-        double pNorm = std::max(this->part_norms(i, j), this->part_norms(j, i));
-        if (tNorm > 0.0) scaled_prec *= tNorm / pNorm;
-    }
-    return scaled_prec;
+    return this->apply_prec;
 }
```
stigrj, line 353: Probably remove this function altogether?

src/qmfunctions/QMFunction.cpp
```diff
@@ -211,6 +211,56 @@ void QMFunction::add(ComplexDouble c, QMFunction inp) {
     mpi::share_function(out, 0, 9911, mpi::comm_share);
 }
 
+/** @brief In place addition of absolute values.
+ *
+ * Output is extended to union grid.
+ *
+ */
+void QMFunction::absadd(ComplexDouble c, QMFunction inp) {
```
stigrj, line 219: This function is currently not used?

src/qmoperators/two_electron/ExchangePotential.cpp
```diff
@@ -149,6 +143,204 @@ Orbital ExchangePotential::dagger(Orbital inp) {
...
+        phi_ij_tmp.alloc(NUMBER::Imag);
+    }
+
+    double precf = prec / 10; // multiplication1 precision
+
+    if (phi_i.hasReal() and phi_j.hasReal()) {
+        mrcpp::multiply(precf, phi_ij.real(), 1.0, phi_i.real(), phi_j.real(), -1, true, true);
+    }
+    if (phi_i.hasImag() and phi_j.hasImag()) { // multiply by +1.0 for complex conjugate and i*i
+        mrcpp::multiply(precf, phi_ij_tmp.real(), 1.0, phi_i.real(), phi_j.real(), -1, true, true);
```
stigrj, line 189: I think there's a typo, should be `imag()` on the two arguments

src/qmoperators/two_electron/ExchangePotential.cpp
```diff
@@ -149,6 +143,204 @@ Orbital ExchangePotential::dagger(Orbital inp) {
...
+    int Nj = phi_i.getNNodes(NUMBER::Total);
+    int N0 = phi_ij.getNNodes(NUMBER::Total);
+    timermult.stop();
+
+    // if the product is smaller than the precision, the result is expected to be negligible
+    if (phi_ij.norm() < prec) {
+        phi_ij.release();
+        phi_ij_tmp.release();
+        return;
+    }
```
stigrj, line 211: The complex multiplication is implemented in `qmfunction_utils.cpp`, so you could get away with
```
Orbital phi_ij = phi_i.paramCopy();
qmfunction::multiply(phi_ij, phi_i, phi_j.dagger(), prec);
```
but it is missing the absolute precision feature, which needs to be added

src/qmoperators/two_electron/ExchangePotential.cpp
```diff
@@ -149,6 +143,204 @@ Orbital ExchangePotential::dagger(Orbital inp) {
...
+
+    std::vector<mrcpp::FunctionTree<3> *> phi_opt_vec; // used to steer precision of Poisson application
+    phi_opt_vec.push_back(&phi_k.real());
+    if (phi_k.hasImag()) { phi_opt_vec.push_back(&phi_k.imag()); }
+    if (phi_out_jij != nullptr) {
+        phi_opt_vec.push_back(&phi_j.real());
+        if (phi_j.hasImag()) { phi_opt_vec.push_back(&phi_j.imag()); }
+    }
```
stigrj, line 219: How will this work if orbitals are complex, i.e. real and imaginary components are mixed in the `phi_opt_vec`? We might also have orbitals without real component, should there be a `if (phi_k.hasReal())` before the first `push_back`?

src/qmoperators/two_electron/ExchangePotential.cpp
```diff
@@ -149,6 +143,204 @@ Orbital ExchangePotential::dagger(Orbital inp) {
...
+    phi_ij.release();
+    phi_ij_tmp.release();
+    double normV = V_ij.norm();
+    timerV.stop();
+    int N2 = V_ij.getNNodes(NUMBER::Total);
+
+    // compute phi_out_kij = phi_k * V_ij
+    Timer timermult2;
+
+    precf = prec / 100; // multiplication2 precision.
```
stigrj, line 245: These precision parameters should be adjustable from input

stigrj:
This rebases seamlessly on `master`, please to
```
git fetch upstream
git rebase upstream/master
git push -f origin newPoisson
```

stigrj:
It would be nice to discuss this once more in person. Will you be in office any time in the next couple of weeks?

gitpeterwind:
Yes, there are many points to discuss! I will be at office most of next week (and probably tomorrow).

src/utils/RRMaximizer.cpp
gitpeterwind: The Hessian elements are computed on the fly (this is not new for this commit). 
When testing for 1000 orbitals, Betzy got scared by the declaration of the huge matrix, even if it was not used. Therefore it is removed.

src/qmfunctions/QMFunction.cpp
gitpeterwind: not used, but should work

src/qmoperators/two_electron/ExchangePotential.cpp
gitpeterwind: yes!

src/qmoperators/two_electron/ExchangePotential.cpp
gitpeterwind: It is allowed to mix anything into phi_opt_vec. The precision condition will work as "logical and" for all functions included.
I have added the  phi_k.hasReal (and phi_j.hasReal()) conditions.

src/qmoperators/two_electron/ExchangePotential.cpp
gitpeterwind: The main precision parameter steering calc_i_Int_jk_P in the HF calculation can now be set by something like:
Precisions {
  exchange_prec = 1.0e-6
}          
in input. No other precision parameters are changed (The "Operator" precision still steers Poisson). 
It involved changing 14 files... so I am not sure all the precision strategies are respected.

src/qmoperators/two_electron/ExchangePotential.cpp
gitpeterwind: Absolute precision for multiply is implemented now

stigrj:
How much is missing for the new algorithms?

gitpeterwind:
> How much is missing for the new algorithms?

You mean the "task" PR? 
It still needs a lot of cleanups. It is anyway building on the new Poisson, so it does not harm to have them after each other.

gitpeterwind:
I have shifted all bank id in the calculation of the D1 exchange integral, so that they do not collide with the D2 orbitals.
This is a quick fix, it can be done better in the future, for example that the bank has several "types" of orbitals: normal, work, D2.

stigrj:
Thanks! I will do a rebase and try to resolve the failing tests

stigrj (approved):

PR #263, opened 2020-02-25
Fix fetching of Eigen3

ilfreddy:


external/upstream/fetch_eigen3.cmake
```diff
@@ -6,19 +6,15 @@ else()
...
+    QUIET
+    URL
+      https://gitlab.com/libeigen/eigen/-/archive/before-git-migration/eigen-before-git-migration.tar.gz
+    )
   FetchContent_GetProperties(eigen3_sources)
 
   if(NOT eigen3_sources_POPULATED)
     FetchContent_Populate(eigen3_sources)
 
-    set(BUILD_TESTING OFF CACHE BOOL "" FORCE)
```
robertodr: This should not have disapperead, but rather moved to before the `if(NOT ...`

robertodr:
Could you also apply the same to MRCPP?

ilfreddy:
> Could you also apply the same to MRCPP?

Yes. As soon as the tests pass on MRChem :-)

PR #265, opened 2020-03-02
Checkpointing and MW guess

stigrj:
- Restructure initial guesses
- Add checkpoint guess to restart from previous calculation
- Amend MW guess to allow re-projection onto different MRA
- Cleanup `mrchem.in`
- Cleanup input parameters and set more practical defaults


stigrj:
I will decompose this into smaller PRs

PR #267, opened 2020-03-04
Modernize energy calculation

bjorgve:
Previously, the kinetic energy could not be calculated accurately, and the total energy was found through massaging equations and sorcery. Now, however, thanks to the ABGV  derivative operator (yes, this should have been fixed ages ago), the kinetic energy can be calculated accurately, in the traditional approach. This should make it allot easier to add energy contributions from other potentials like `V_ff (periodic)`, `V_r (solvation)` and `V_ext (external electric field)`. 

Tests appear to work for all cases apart from for `H_SCF_finite_electric_field`, I'm not sure why, but probably some sorcery, that no longer should be necessary, was done to make it work to begin with.

The Hartree--Fock test also fails, but I suspect this is due to errors in the reference values.

As a sidenote `H_SCF_finite_electric_field` does not appear to test the `E_nex`, it is zero. Is this tested somewhere else?


stigrj:
This snowballed a bit. I completely removed the `EnergyOptimizer` and moved the `OrbitalOptimizer` up into the `GroundStateSolver` base. @bjorgve you'll have to self-review this one

PR #268, opened 2020-03-11
Update compiler flags

stigrj:
Added some compiler flags. The missing `-xHost` Intel flag had a significant effect on performance (30%) and accounts for the discrepancy we saw between Intel and GNU before.

Not sure if all the other should be there (they come from EasyBuild defaults) or if we need more?


robertodr:
Is CircleCI blocked somehow?

PR #269, opened 2020-03-11
Fix debug/release combo fail

stigrj:
See https://github.com/MRChemSoft/mrcpp/pull/133


robertodr:
```
/home/travis/build/MRChemSoft/mrchem/build/_deps/eigen3_sources-src/Eigen/src/Core/arch/AVX512/MathFunctions.h:75:97: error: macro "_mm512_srli_epi32" passed 3 arguments, but takes just 2
 mm512_cvtepi32_ps(_mm512_srli_epi32(preinterpret<Packet16i,Packet16f>(x), 23));
```

stigrj:
It's a bit inconvenient to require a released version of mrcpp. How about we go back to fetching from git repo in the development `master` branch of mrchem, allowing also to point to unreleased commits in mrcpp, but then on the `release` branches we require a properly released URL tarball?

robertodr:
That works for me. Though this seems a fundamental enough change to warrant at least a new alpha version.

stigrj:
Yes, there will be a new mrcpp-alpha as soon as this https://github.com/MRChemSoft/mrcpp/pull/131 is included. But I still think it is better to be more dynamic on the development branch. Also, our fancy git-describe `binary-info` stuff did not work without a git repo...

stigrj:
I don't know how to tackle this error, it's not reproduced on Woolf which is the only `avx512` machine I have access to. Is it not possible to get an ssh to the Travis build, as for Circle?

robertodr:
Never investigated that possibility...

stigrj:
Alright! Eigen 3.3.90 reproduces this on woolf. How can I avoid this particular version?

stigrj:
Not even sure where that version comes from... Seems a bit of a leap fro 3.3.7 to 3.3.90

robertodr:
> Not even sure where that version comes from... Seems a bit of a leap fro 3.3.7 to 3.3.90

I think they're moving towards releasing a 3.4, hence the 3.3.90 version for the tag we're downloading. I think the best is to keep downloading 3.3.7 but then patch the `EigenConfig.cmake` file for our purposes (like this: https://gitlab.com/libeigen/eigen/-/commit/2cbd9dd49806d686a40841b6d888a83c816efccf#618d329ccc83294fb4adc1383903316e95a1236d) What a clusterf**k

stigrj:
What exactly should I do here?

stigrj:
:tada:

PR #270, opened 2020-03-16
Property printing

stigrj:
Initialize properties to NaN, so that un-computed properties are not confused with zero value.

E.g. polarizability computed only in x and y `directions=[1,1,0]` will look like this
```
======================================================================
                            Polarizability
----------------------------------------------------------------------
 Wavelength            :           (nm)                    694.573171
 Frequency             :           (au)                      0.065600
                       :         (cm-1)                  14397.331226
----------------------------------------------------------------------
 Total tensor          :       1.668198       0.000000       0.000000
                       :       0.000000       1.668198       0.000000
                       :            N/A            N/A            N/A
 Isotropic average     :           (au)                           N/A
======================================================================
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -123,3 +123,5 @@ doc/xml/
 *_inst
 *_build
 .ccls-cache/
+
+Pipfile.lock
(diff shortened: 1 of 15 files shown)


stigrj:
Rebased and readt to go

PR #272, opened 2020-03-18
Fix displayed final precision in response

stigrj:
Response solver printed out `current_prec` and `start_prec` instead of `start_prec` and `final_prec`


stigrj:
another small fix coming up

PR #294, opened 2020-05-05
Update external to official MRCPP-1.2.0 and XCFun-2.0.0

stigrj:


stigrj:
We get the following error in `debug`, reproduced on my laptop
```
make[2]: *** No rule to make target 'lib/libxcfun_d.so.2', needed by 'lib/libmrchem.so'.
```
@robertodr do you know where the "_d" comes from?

robertodr:
It's the suffix appended for the debug version. It's not quite clear why it looks for the debug version when it's building it in release...

stigrj:
I'm not able to locate this error. Should I just change back to debug build of XCFun, or do you want to have a go at it?

robertodr:
I can have a look tomorrow.

robertodr:
I cannot reproduce locally using your branch. I am building with `./setup --type=release`

stigrj:
It's when building mrchem in debug while xcfun is release

robertodr:
> It's when building mrchem in debug while xcfun is release

OK, I had completely misunderstood...

robertodr:
I think the quickest is to remove the `DEBUG_POSTFIX` property from the `xcfun` target directly in the XCFun repo. I will release a v2.0.1 once I've tested it with MRChem.

robertodr:
See dftlibs/xcfun#130

stigrj:
Wow, that's a heavy blow on code coverage

robertodr:
> Wow, that's a heavy blow on code coverage

How's that possible? :scream:

stigrj:
Looks like _a lot_ of files were not included in the coverage before, and for some reason they were included now. I think we need a unit test hackathon

PR #300, opened 2020-06-25
Few CMake fixes

robertodr:


stigrj:
These are for the linking order warnings?

robertodr:
> These are for the linking order warnings?

Forgot to describe what I did... Yes, these fix those warnings and also disallow CMake from being "too eager" in searching for dependencies.

stigrj:
I still get these warnings...

robertodr:
Was this on a fresh copy with no installations around?

stigrj:
Yes, fresh copy. First install is fine, then 
```
$ touch src/input/mrchem.in
```
and a new `make` gives (without removing the old install)
```
-- Configuring done
CMake Warning at src/CMakeLists.txt:78 (add_executable):
  Cannot generate a safe runtime search path for target mrchem.x because
  there is a cycle in the constraint graph:

    dir 0 is [/home/stig/src/mrchem-master/build/lib]
      dir 1 must precede it due to runtime library [libxcfun.so.2]
    dir 1 is [/home/stig/src/mrchem-master/install/lib]
      dir 0 must precede it due to runtime library [libmrchem.so]

  Some of these libraries may not be found correctly.
```

stigrj:
Should we still merge this?

robertodr:
I'd merge: they don't solve the problem, but they don't hurt either.

PR #305, opened 2020-08-25
Build libmrchem as static library

robertodr:
This fixes an obscure issue with dynamic linking that has annoyed @stigrj for quite a long time now. I still don't understand the original issue, but we agreed it's fine to have this fix in rather than break our teeth on getting dynamic linking right.


bjorgve (approved):

stigrj:
I also removed the `STATIC_LIBRARY_ONLY` line in `fetch_xcfun.cmake` which is no longer relevant

stigrj (approved):

PR #306, opened 2020-08-25
Make parselglossy a developer-only dependency

robertodr:
The parser is now generated by the developers as a Python package that is checked in the repo. Every time `template.yml` changes, so must the generated Python package. This avoids cascading the dependency on `parselglossy` down to end-users who might not know/care about the subtleties of Python package management.

**Bonus** (@bjorgve @Gabrielgerez) The `mrchem` frontend now tolerates having the `.inp` in the filename of the input.

## How to regenerate the parser upon changing `template.yml`

- `cd src/input`
- Run `parselglossy generate --template template.yml --docfile user_ref.rst --doc-header="User input reference" `
- Run `cp input_parser/docs/user_ref.rst ../../doc/users/user_ref.rst`  to update the reference documentation for keywords and sections. 

This will produce a folder `input_parser` with tree:
```
input_parser/
├── api.py
├── cli.py
├── docs
│   └── user_ref.rst
├── __init__.py
├── plumbing
│   ├── atoms.py
│   ├── exceptions.py
│   ├── getkw.py
│   ├── lexer.py
│   ├── pyparsing.py
│   ├── types.py
│   ├── utils.py
│   ├── validation_plumbing.py
│   ├── validation.py
│   └── views.py
└── README.md
```

This folder will be copied where appropriate by CMake when installing.


robertodr:
To fix the tests I need to fix a bug in `parselglossy`...

robertodr:
All green now!

stigrj:
Hello, 10k lines of code :wave:

robertodr:
LOL, yes. That's because `template.yml` is validated and the validated representation saved as a dictionary in the auto-generated `api.py`. Not much I can do about that, maybe prune the `docstring`-s which are not used during parsing.

stigrj (changes requested):
Can you update `requirements.txt` as well (If we still want to keep this)?

robertodr:
> Can you update `requirements.txt` as well (If we still want to keep this)?

This file will eventually go (when `runtest` is gone too) as it is supposed to specify end-user dependencies. We might add a `requirements-dev.txt` file, but the `Pipfile` should be good enough.

stigrj:
This is probably a `parselglossy` issue, but the autogenerated input reference does not line break very nicely. Idk if anything changed here recently, but the current official [manual](https://mrchem.readthedocs.io/en/latest/users/user_ref.html) looks much better (I had to manually break the long predicate lists like `WaveFunction.method`, though).

robertodr:
It's a `parselglossy` issue, because I didn't spend much time looking at pretty-printing documentation. If you look at the diff for `user_ref.rst` the changes are end-of-line whitespaces (and the line breaks you added by hand, which I can put back)
Maybe open an issue on `parselglossy`? I can't promise it will be prioritized: prettyfying the docs is not among the goals of the project :smile_cat:

stigrj (approved):

PR #308, opened 2020-08-26
Remove runtest dependency

robertodr:
MRChem can output its results in structured JSON format and is thus unnecessary to extract (_via_ regex) results from the more user-friendly output files.

The new framework looks mostly like `runtest`: you still have to write a little `test` script in the folder with the input file.
Notable differences:
1. You have to check in under the `reference` folder the JSON output. You can check in the usual `.out` file too, but this one is not checked against anymore.
2. The floating point comparisons are run _à la_ NumPy. Given an absolute tolerance `atol` and relative tolerance `rtol`:
```python
abs(computed - expected) <= atol + rtol * abs(expected)
```
3. The filters are given as a Python `dict`. The keys are _tuples_, the values a `Tolerance` object (a pair of absolute and relative tolerance values). The tuple is the address in the JSON file we need to probe for the value to check. For example, to check the computed sum of orbital energies to `runtest`'s `rel_tolerance=1.0e-6` you would do:
```python
filters = {("output", "properties", "orbital_energies", "sum_occupied"): Tolerance(rtol=1.0e-6, atol=1.0e-6)}
```

If you wanted to check to `runtest`'s `abs_tolerance=1.0e-6` you'd have to use `Tolerance(rtol=0.0, atol=1.0e-6)`
You can find all the "addresses" in the JSON output in the schema: https://mrchem.readthedocs.io/en/latest/users/program_json.html#program-input-output-file

This is, of course, very tedious, hence I've put in some "aliases" for commonly used "addresses" and functions to mimic `runtest` tolerance set ups. For the above, you can do:
```python
filters = {SUM_OCCUPIED: rel_tolerance(1.0e-6)}
```
and similarly for `abs_tolerance`.

With this change, MRChem is free from external Python packages for end-users: it can be downloaded, compiled, and run without having to create virtual environments.


tests/tester.py
```diff
@@ -0,0 +1,349 @@
...
+        help="working directory [default: %default]",
+    )
+    parser.add_option(
+        "--launch-agent",
+        "-l",
+        action="store",
+        default=None,
+        help=
+        'prepend a launch agent command (e.g. "mpirun -np 8" or "valgrind --leak-check=yes") [default: %default]',
+    )
```
stigrj, line 232: This feature is already available with the `--launcher` option to `mrchem`

robertodr: So it's enough to set it with the `extra_args` arguments to the `run` function?

stigrj: yes

robertodr: Cool. Then this part in the `CMakeLists.txt`:
```cmake
set(_h_el_field_launcher " ")

if(ENABLE_MPI)
    set(_h_el_field_launcher "mpirun -np 1")
endif()
```
is unnecessary too, right?

robertodr: I think the CMake argument is still necessary, but it should be used to set `extra_args` in the macro

robertodr: Think I got it correct now.

bjorgve (approved):

stigrj:
Are any of the tests now run in MPI? Before we had `mpirun -np 2` for `h2o_energy_blyp`, `li_pol_lda` and `li_scf_pbe0`, but I cannot see any `mpirun` in the new setup

robertodr:
> Are any of the tests now run in MPI? Before we had `mpirun -np 2` for `h2o_energy_blyp`, `li_pol_lda` and `li_scf_pbe0`, but I cannot see any `mpirun` in the new setup

We still pass the launch agent option to the `test` script which then set it for the `mrchem` script: https://github.com/MRChemSoft/mrchem/pull/308/files#diff-7ecb0701ffeb2d223889e59042ff9a64R67-R68

PR #309, opened 2020-08-27
Require python 3.6 or better to run setup

bjorgve:
You won't be able to run the code with python 2.

### DIFF
--- a/setup
+++ b/setup
@@ -5,7 +5,7 @@
 
 import os
 import sys
-assert sys.version_info >= (2, 6), 'Python >= 2.6 is required'
+assert sys.version_info >= (3, 6), 'Python >= 3.6 is required'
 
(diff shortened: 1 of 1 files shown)


robertodr (changes requested):
This script is autogenerated and this change will be overwritten next time it's updated. The `setup` script itself _can_ run with Python as early as 2.6 I guess what we want is at the CMake level.

robertodr:
To be clear: you would add something of the sort:

``` cmake
if("${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}" VERSION_LESS 3.6)
  message(FATAL_ERROR "You will need Python >= 3.6 to run MRChem")
endif()
```

in the `cmake/custom/main.cmake`.

bjorgve:
Opened it again under a new name

PR #310, opened 2020-08-27
Clean up after removal of runtest

robertodr:
Irons out some wrinkles left when removing the dependency on ``runtest``, most notably some confusion on what the working directory was supposed to be.

### DIFF
--- a/tests/tester.py
+++ b/tests/tester.py
@@ -78,7 +78,7 @@ def run(options, *, input_file, filters=None, extra_args=None):
     output_prefix = inp_no_suffix
 
     sys.stdout.write(
-        f"\nrunning {' '.join(command)}\ntest with input files {input_file} and args {extra_args}"
+        f"\nrunning {' '.join(command)}\ntest with input files {input_file} and args {extra_args}\n"
     )
 
     if options.skip_run:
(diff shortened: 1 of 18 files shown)


tests/tester.py
```diff
@@ -77,37 +84,49 @@ def run(options, *, input_file, filters=None, extra_args=None):
...
                                              rtol=threshold.rtol,
                                              atol=threshold.atol)
             sys.stdout.write(f"\n{message}")
+            success = success and passed
+
+    # clean up output files on success (if requested!)
+    if success and options.clean_up:
+        computed.unlink()
+        out = Path(options.work_dir) / f"{inp_no_suffix}.out"
+        out.unlink(missing_ok=True)
```
stigrj, line 127: The `missing_ok` argument is not recognized. Why do we want to clean up the outputs after success? Sometimes I like to inspect the output also for successful tests.

robertodr: That argument is in 3.7+ only. And the removal was a workaround for not using the working directory properly.

stigrj:
Arg!  :angry:

robertodr:
> Arg!

Indeed

robertodr:
It's apparently choking on this line: `out_file = f"{name}.out"` But the Docker image we are using has Python 3.6 and that is allowed syntax... :thinking:

robertodr:
A-ha! The shebang in `mrchem.in` should be interpolated by CMake not hard-coded to `/usr/bin/env python`!

robertodr:
I am not sure whether we want to remove `#!/usr/bin/env python` from the little `test` scripts too. That would require executing them in the test folders prepending the correct Python executable, but would avoid these arcane errors to pop up.

stigrj (approved):
I have no strong feelings on the last point. Do whatever you think is best and then merge.

robertodr:
I went for `#!/usr/bin/env python3`

PR #311, opened 2020-08-28
Require python 3.6 to run setup

bjorgve:
Title says it all, had some issues with updating the original PR.


robertodr (approved):

robertodr:
Did you test on woolf or sandel that this indeed catches the problem?

bjorgve:
I tried it on Sandel, not really on woolf (got some cmake requirement issues).

PR #313, opened 2020-08-30
Improve handling of JSON input

robertodr:
- Allow to pass suffixed (`.inp`, `.json`) or unsuffixed input files.
- Recognize format from suffix, if present, else follow `--json` flag.
- JSON output file is named `<inp_name>_out.json`, such that the JSON
  input is not overwritten.

### DIFF
--- a/src/input/mrchem.in
+++ b/src/input/mrchem.in
@@ -53,11 +53,24 @@ def main():
     # Parse command line
     launcher, executable, dryrun, stdout, inp_json, inp_name = parse_cmdline()
 
-    inp_file = Path(inp_name) if inp_name.endswith(".inp") else Path(inp_name + ".inp")
+    suffix = Path(inp_name).suffix
+    if suffix == ".inp":
+       inp_file = Path(inp_name)
+    elif suffix == ".json":
+       inp_file = Path(inp_name)
+       inp_json = True
+    elif suffix == "":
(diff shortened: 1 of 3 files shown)


robertodr:
@stigrj I think I have some confusion in my head as to what the JSON input should really be when fed to the `mrchem` Python front-end.
Is it the `program_dict`? 
```json
{
"input": {...},
"output": {...}
}
```

Or is it the JSON I would get from only running `lex` on the `.inp` file (_i.e._ what is called _intermediate representation_ in `parselglossy`)?
```json
{
"world_prec": ...,
"WaveFunction": {...}
}
```

I think both should be allowed. What's your opinion?

stigrj:
It's the latter. The python front-end accepts the user input, which is the one defined in `template.yml`, but its format can be either getkw or JSON.

stigrj:
Can you elaborate a bit on the need for this change? This is basically how things were two months ago, and I kind of like the current setup better.

robertodr:
I'm trying to figure out what's best to run MRChem through QCEngine. I need to produce a JSON file at some point, but it could also be enough with producing a `dict`. 
So I might throw away this PR.
I think what I actually need is to a) modify `parselglossy` slightly b) separate the `mrchem` script into an `api.py` and a `cli.py`.

PR #314, opened 2020-09-01
Refactor mrchem.in

robertodr:
All Python portions now live in the `python` subfolder of the source tree.

- Upgrade to latest `parselglossy`
- Regenerate the input parser
- Rewrite the CLI to use `argparse` (`optparse` is deprecated)
- Split the Python frontend into `mrchem.in`, a CLI module, and an API module. The latter two live in a `mrchem` Python module, of which the generated input parser is a submodule.
- All Python portions of the code (including `periodictable.py`) are now copied into the `python/mrchem` folder, that is: they are available as a Python module. ~**QUESTION** Should all of this be moved to some other folder in the source tree?~
- The Bohr to meter conversion factor is from CODATA 2018 *WARNING* This gives a different Angstrom to AU conversion factor that could impact results (though the tests are unaffected)
- Remove one Autocmake module in favor of better alternative in CMake>=3.12
- Update minimum CMake requirement to 3.12
- Clean up `.travis.yml` and `.circleci/config.yml`
- Remove outdated `lsdalton-guess` script and `lsdalton.py` module


stigrj:
`periodictable.py` is also only used in the front-end script, so it could perhaps also be moved to `src/input/mrchem`?

stigrj (approved):
Thanks a lot, I've been putting off this refactoring a long time :slightly_smiling_face:

robertodr:
> `periodictable.py` is also only used in the front-end script, so it could perhaps also be moved to `src/input/mrchem`?

I agree with moving, but should we move the whole Python front-end to some other folder?

stigrj:
You mean moving everything in `src/input` out from `src`? I think that's a good idea, then `src` will only involve the `mrchem.x` binary.

robertodr:
Yes. I'll create a `python` folder under the source tree with all these bits.

robertodr:
What should I do with the `lsdalton` folder under `initial-guess`?

stigrj:
Hmm, I think it should be nuked. It doesn't really work anymore, and hopefully we'll have a better solution soon.

robertodr:
> Hmm, I think it should be nuked. It doesn't really work anymore, and hopefully we'll have a better solution soon.

:bomb: awesome!

stigrj:
Please squash when finished. I forgot in the previous PRs that we need to cherry-pick everything to `release/1.0` :man_facepalming:

robertodr:
Yes, I'll squash. I'm still fixing some details.

robertodr:
@stigrj you might want to re-review.

PR #315, opened 2020-09-03
Prepare MRChem for QCEngine

robertodr:
These are some changes necessary to run MRChem through QCEngine.

- [X] The frontend script can print the location of the `mrchem` Python module.
- [X] The input and output schemas have been extended with versioning fields.
- [X] The output schema gained a `provenance` section.
- [X] The global success of a calculation is saved to the `success` field of the output schema.
~- [ ] Add a function to translate a QCElemental molecule specification to the format expected by the MRChem input schema~
~- [ ] Add a function to translate between the MRChem output schema and the [QCSchema output](https://molssi-qc-schema.readthedocs.io/en/latest/spec_components.html#output-components)~

The accompanying PR to QCEngine: MolSSI/QCEngine#269


robertodr:
- I have submitted a PR to QCElemental for handling molecule info. MolSSI/QCElemental#238
- Translating from MRChem schema to QCSchema has been left into the QCEngine harness.

python/template.yml
```diff
@@ -310,12 +310,12 @@ sections:
...
+        default: true
         docstring: |
           Compute dipole moment.
       - name: quadrupole_moment
         type: bool
-        default: false
+        default: true
         docstring: |
           Compute quadrupole moment. Note: Gauge origin dependent, should be
           used with ``translate = true`` in Molecule.
```
stigrj, line 321: Is the quadrupole moment always expected in the qcschema? I think our origin dependent definition is not very nice when the property is computed by default

robertodr:
No, it's not.

python/template.yml
robertodr: I chose to have both of them computed when SCF succeeds to conform with what some GTO codes do. The quadrupole moment is almost always origin-dependent, so I agree it is not the most significant thing to compute. I can change this back to `false`. The "issue" was mostly about the dipole, because that is a property listed in the QCSchema (but not required to be there) 
I have anyway structured the harness to identify which properties were computed among the ones known to MRChem and to move them to the correct spot in the QCSchema.

stigrj: Then I'm in favor of `false` for the quadrupole

stigrj (approved):

PR #316, opened 2020-09-04
Fix RPATH handling

robertodr:
Fixes the issue observed by @bjorgve and @stigrj with linking using a pre-installed MRCPP. Tested on `sandel`.

Note that this is the line that fixed it: https://github.com/MRChemSoft/mrchem/pull/316/files#diff-95e351a3805a1dafa85bf20b81d086e6R101
I have changed also the `fetch_*` files such that upon rebuilding the build tree links against libraries in the build tree (if present), rather than ones previously installed.


bjorgve (approved):
It works wonders on both woolf and sandel

PR #318, opened 2020-09-10
Tasks

gitpeterwind:
Complete reorganization of the parallelization of the loops over orbitals.
The matrices computed to build the Fock matrix ( <Orb_a| O | Orb_b> ) are organized in blocks. The list of blocks is considered as a task list which is distributed among MPI workers. (there is no "ownership" of orbitals in this approach, but the ownership is still used indirectly).
The main new parts are for:
1) rotation among orbitals
2) overlap, which is used for kinetic energy and other parts too
3) Exchange also, although less important.

Some other improvement too. The localization routine is now much faster.

The tasks are distributed by a "task" manager. One MPI process is reserved only to distribute tasks, and keep accounting of what has been done etc. (this process is taken among the "banks")
The banks are now much more used, and a significant amount of the time is used to move orbitals to and from the bank. 
The number of banks must be much larger, maybe 1/3 to 1/2 of all MPI processes. 
It is observed that it is more effective to have banks and workers on the same compute-nodes. (not clear why). This can be achieved with for example:

` mpirun  --rank-by node -map-by numa -bind-to numa mrchem.x  gram.json`

The size of the blocks are assigned dynamically. Large blocks are often beneficial because the orbitals can be reused, diminishing total communication time. If the blocks are too large, there may be a memory limitation. The total memory available is now estimated based on 2GB per thread.


gitpeterwind:
Most parts of this PR have been transferred to other PR

PR #319, opened 2020-09-12
Fix failing ReadTheDocs build

stigrj:


robertodr (approved):
I guess it's my fault that they were failing isn't it?

PR #323, opened 2020-09-22
Clean docs

stigrj:


doc/installation.rst
```diff
@@ -211,3 +203,13 @@ Please refer to the :ref:`User's Manual` for instructions for how to run the pro
...
     dependencies`, so this must be done manually in order to run the code.
+
+
+Installing using Conda
+----------------------
+
+The OpenMP parallel version of MRChem can be installed in a Conda environment
+named ``myenv`` with::
+
+    $ conda create -n myenv mrchem -c conda-forge
```
robertodr, line 214: Maybe add a note that this will install the OpenMP-parallel package. To install an MPI/OpenMP parallel version:
a. With OpenMPI
```bash
conda create -n myenv mrchem=*=*openmpi* -c conda-forge
```
b. With MPICH
```bash
conda create -n myenv mrchem=*=*mpich* -c conda-forge
```

robertodr (approved):
All looks good. I would suggest to mention both Conda and Docker in the `README.md` so that "casual browsers" might get an interest in trying it out in a simple  way.

doc/installation.rst
stigrj: Has the MPI/OpenMP been tested? I'm a bit sceptical :slightly_smiling_face:

robertodr: No, it hasn't. The skepticism is about performance or whether it works at all? I'm also not so sure the former will be amazing, but the build came for free so :man_shrugging: 

The package is still under staging and as such cannot be downloaded yet. I'm too lazy to build it locally and test... What you can try now is installing the hybrid version of the MRCPP package and using it in a sample executable. There might still be sharp edges (for example, having to use the C++ compiler that comes with Conda) This first iteration of the Conda packaging was aimed at obtaining an easy-to-install package for MRChem so that "casual browsers" can try it without much trouble, but also not expecting much performance from it.

robertodr: I guess my point is: the docs should make absolutely clear that Docker and Conda are _NOT_ the state-of-the-art when it comes to what performance one can expect from MRChem.

Unrelated: maybe the docs main page can have some numbers on scaling and timings?

stigrj: Just a general skepticism for anything working out of the box :slightly_smiling_face:

stigrj:
Do you remember the details on using the Docker image?

doc/installation.rst
robertodr: Understandable :)

robertodr:
> Do you remember the details on using the Docker image?

No, not really. Does this still work?

```bash
docker run mrchemsoft/mrchem_ubuntu18.04 </path/to/input/file>
```

robertodr:
The bot complains :facepalm: But you can use your admin privileges to squash-merge

PR #329, opened 2020-10-12
Fix MPI tests

stigrj:


robertodr (approved):
All good, but same comments as the MRCPP PR for `cmake/compiler_flags/CXXFlags.cmake`

stigrj:
still issues with the integration tests on Fram/Saga

stigrj:
So I fixed Saga/Fram but broke Circle :ok_hand:

Issue #330, opened 2020-10-15
Can we make `QMPotential` not virtual?


ilfreddy:
@stigrj and @robertodr 

I think `QMPotential` should be made real (non virtual).

1. Currently only `setup()` and `clear()` are preventing `QMPotential` from being a non-virtual class. In several cases those two routines reduce however to 
`void setup(double prec) override { setApplyPrec(prec); }` and `void clear() override { clearApplyPrec(); }`.
2. It is meaningful to consider a generic potenital for a particle, which does not need more than just the definition of the potential function.

I would then suggest to use the definintions of those two methods as the generic ones and overridde those only when needed.

stigrj:
agreed

ilfreddy:
I will prepare a PR

ilfreddy:
Question: why is `free(NUMBER::Total)` in the destructor for `NuclearPotential` but it is in the `clear()` method for all other classes? Is there a reason for it?

stigrj:
Depends on whether the potential is built in the constructor or in `setup`. Originally, all potentials were built with precision `prec` in `setup(prec)` and deleted in `clear()`, which is the reason these were made virtual in the first place. But then the `NuclearPotential` was so time consuming to build, so it was better to move it's projection to the constructor.

This makes a difference when we're using dynamic precision in the SCF, starting with low `start_prec` and ending with high `final_prec`. So instead of re-building the `NuclearPotential` with `current_prec` in every iteration, it is instead build with `final_prec` in the constructor, and then we only vary the `apply_prec` through `setup`.

So whether the potential _should_ be build in the constructor or in `setup` will depend on the operator, e.g. the `CoulombPotential` must be built in `setup` because it depends on the current orbitals, but if the operator is "static" it should probably be built in the constructor (note that also memory footprint needs to be considered here).

ilfreddy:
Thanks. Then I will leave `NuclearPotential` as it was.

PR #331, opened 2020-10-15
Makes QMPotenital non-virtual again!

ilfreddy:
closing #330

### DIFF
--- a/src/qmoperators/one_electron/QMPotential.h
+++ b/src/qmoperators/one_electron/QMPotential.h
@@ -37,6 +37,9 @@ protected:
     Orbital apply(Orbital inp) override;
     Orbital dagger(Orbital inp) override;
 
+    virtual void setup(double prec) override { setApplyPrec(prec); }
+    virtual void clear() override { free(NUMBER::Total); }
+
     void calcRealPart(Orbital &out, Orbital &inp, bool dagger);
     void calcImagPart(Orbital &out, Orbital &inp, bool dagger);
 };
--- a/src/qmoperators/one_electron/QMPotential.h
+++ b/src/qmoperators/one_electron/QMPotential.h
@@ -37,9 +37,6 @@ protected:
     Orbital apply(Orbital inp) override;
     Orbital dagger(Orbital inp) override;
 
-    virtual void setup(double prec) override { setApplyPrec(prec); }
-    virtual void clear() override { clearApplyPrec(); }
-
     void calcRealPart(Orbital &out, Orbital &inp, bool dagger);
     void calcImagPart(Orbital &out, Orbital &inp, bool dagger);
 };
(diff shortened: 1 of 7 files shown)


src/qmoperators/one_electron/QMPotential.h
```diff
@@ -37,6 +37,12 @@ class QMPotential : public QMFunction, public QMOperator {
     Orbital apply(Orbital inp) override;
     Orbital dagger(Orbital inp) override;
 
+    virtual void setup(double prec) override { setApplyPrec(prec); }
+    virtual void clear() override {
+        free(NUMBER::Total);
```
stigrj, line 42: We should keep the non-free version here for consistency. Now the expected default behavior of `QMOperator` is that the potential is "persistent", i.e. built in constructor and kept through multiple `setup`'s and `clear`'s. If the potential needs to be re-computed every time, then both `setup` and `clear` should be overwritten. It will lead to less code-reuse, but the solution is rather to make more of the other potentials persistent.

stigrj (changes requested):

src/qmoperators/one_electron/QMPotential.h
ilfreddy: I see. I'm fine either way. I'd go for the first solution, which is easier to implement.

src/qmoperators/one_electron/QMPotential.h
```diff
@@ -37,6 +37,9 @@ class QMPotential : public QMFunction, public QMOperator {
     Orbital apply(Orbital inp) override;
     Orbital dagger(Orbital inp) override;
 
+    virtual void setup(double prec) override { setApplyPrec(prec); }
+    virtual void clear() override { free(NUMBER::Total); }
```
stigrj, line 41: Nope, the other way :smiley: This should be
```
virtual void clear() override { clearApplyPrec(); }
```
and then remove `NuclearPotential::clear()`

ilfreddy: Ooops... I just deleted the wrong line of code :-)

stigrj:
Actually, this could be escalated to `QMOperator`, so that `setup/clear` can be removed from classes `QMIdentity`, `QMSpin`, `QMNabla` and `QMMomentum`

stigrj:
Hold on, I will make some extra changes :slightly_smiling_face:

stigrj:
I decided to restrain myself :slightly_smiling_face: My plan was to move all `setup` builds into constructors for simple `QMPotentials`, but changed my mind...

stigrj:
I think the right thing to do in the end is to build the one-electron operators in their constructor and two-electron operators in `setup`, but then add more arguments to the `setup` function, like `J.setup(prec, Phi)` or `J.setup(prec, Rho)` for the `CoulombPotential`, and thus not tie the two-electron operators to particular `OrbitalVectors` at construct-time. But all this will require much more work :slightly_smiling_face:

stigrj (approved):

PR #333, opened 2020-10-23
Add GitHub action

robertodr:
So we test on macOS... This builds in release mode with OpenMP enabled. It gets Eigen, XCFun, and MRCPP from conda-forge.


robertodr:
Needs to be done as a direct commit to the main repo :facepalm:

PR #334, opened 2020-10-23
Add build and test action

robertodr:
So we test on macOS... This builds in release mode with OpenMP enabled and no architecture-specific flags. It gets Eigen, JSON library, XCFun, and MRCPP from conda-forge.


stigrj:
This was way too easy to set up :wink:

robertodr:
> This was way too easy to set up

Indeed, just 3 commits...

stigrj (approved):

robertodr:
I'll add this to the checks required to pass for PRs

PR #336, opened 2020-10-28
Prepare v1.0.0

stigrj:
@robertodr do you think the CHANGELOG entry for v1.0.0 should contain only changes since `1.0.0-alpha3`, or should we combine all important changes since the last proper release?


robertodr (approved):
:1st_place_medal: awesome

robertodr:
The changelog looks good as is, but I would put all the new goodies since the last stable release in the description of the GitHub release.

PR #340, opened 2020-11-06
Fix NMR print

stigrj:


doc/users/user_inp.rst
```diff
@@ -299,7 +299,10 @@ The ``nuclear_specific`` keyword triggers response calculations using the
 nuclear magnetic moment operator instead of the external magnetic field. For
 small molecules this is not recommended since it requires a separate response
 calculation for each nucleus, but it might be beneficial for larger systems if
-you are interested only in a single shielding constant.
+you are interested only in a single shielding constant. Note that the components
+of the *perturbing* operator defines the *row* index in the output tensor, so
+``nuclear_specific = true`` will result in a shielding tensor which is
+the transpose of ``false``.
```
robertodr, line 305: "the transpose of the one obtained by setting `nuclear_specific = false`". More verbose, but less ambiguous. I think.

src/properties/NMRShielding.h
```diff
@@ -53,12 +53,12 @@ class NMRShielding final {
 
     void print(const std::string &id) const {
         auto sigma = getTensor();
-        Eigen::SelfAdjointEigenSolver<DoubleMatrix> es;
+        Eigen::EigenSolver<DoubleMatrix> es;
```
robertodr, line 56: Why is the matrix no longer self-adjoint?

src/qmoperators/one_electron/H_MB_dia.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#include "NuclearGradientOperator.h"
+#include "PositionOperator.h"
+#include "qmoperators/RankTwoTensorOperator.h"
+
+namespace mrchem {
+
```
robertodr, line 8: Can you add a Doxygen string explaining what this operator is? Maybe also for the `H_BM_dia` one, while you're at it :grin:

robertodr (approved):

src/properties/NMRShielding.h
stigrj: It shouldn't be. It was an error before which lead to wrong calculation of anisotropy. I guess it was only tested on symmetric cases?

robertodr (approved):
So much documentation added, I'm awed :smile:

src/properties/NMRShielding.h
robertodr: And what happens with the imaginary part of the eigenvalues?

stigrj: No idea, but now we're consistent with orca :slightly_smiling_face:  It might never become complex?

stigrj: I'm adding a warning in case that happens

PR #347, opened 2020-11-14
Fix precision issue with chk initial guess

stigrj:
Fixes issue reported by Anders that `chk` initial guesses was limited by the chosen `guess_prec`, due to an orbital rotation immediately after loading the orbitals. Now the `chk` guess will get the full `world_prec` for this rotation. All other guesses will still use `guess_prec`, but they are anyway inherently limited by this precision in their construction (unlike `chk`).


robertodr (approved):

stigrj:
Any hope of fixing the bot, or should I just override?

robertodr:
You should override. I've tried few things and none have worked...

stigrj:
There was a conflict with the last PR, so I reverted the Minazo commits

PR #351, opened 2020-11-19
Fix MPI deadlock

stigrj:
Fixes MPI deadlock introduced in #328. The lock is triggered whenever the MPI orbital bank occupies _all but one_ rank on a particular node, _and_ at the same time a function is allocated in shared memory. Then the remaining rank on the node would mark the function as non-shared (because it's not actually sharing with anyone), which lead to subsequent `broadcast_function` to be called with a different communicator, thus deadlock.

### DIFF
--- a/.github/workflows/all-hail-minazo-bot.yml
+++ b/.github/workflows/all-hail-minazo-bot.yml
@@ -8,9 +8,7 @@ jobs:
   danger:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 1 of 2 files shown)


robertodr:
I tacked another attempt at fixing the bot on top of your changes (sorry!) I didn't want to spam with more one-line PRs...

gitpeterwind (approved):
Well done!

PR #352, opened 2020-11-21
Limit Eigen to single thread

stigrj:
Threading lead to highly unpredictable results for large matrix-matrix multiplications, which caused Core and SAD guesses to fail for larger systems.


robertodr (approved):
Maybe this should be considered: http://eigen.tuxfamily.org/dox/TopicUsingBlasLapack.html to squeeze some extra performance.

stigrj:
I tried recently to use standard BLAS instead of Eigen for the core MRCPP routines, but it was actually slightly slower. But everything performance related should definitely be reconsidered, including BLAS and Eigen threading

PR #359, opened 2021-02-01
Superfast rotate

gitpeterwind:
New faster method for rotation and dot product of orbital vectors. Those are the parts that scale as 0(N²)
The idea is to rotate one nodes at a time; this can be done as a matrix multiplication which is extremely fast.
For Valinomycine (300 orbitals), rotation is up to 50 times faster. Overall 5 times faster LDA calculation.
The benefit is largest for large molecules.

Could still relatively easily be further optimized, as the mpi version does not even use omp (!) for example. However other parts are now the bottleneck (XC and Exchange) and should be optimized first.

The functions that have been replaced are "rotate" and "overlap". Unfortunately those have become "monolithic" and heavy, since they are written for efficiency (avoid data copying and use pointer and indices manipulation).

Other: faster localization, and accept U matrix also when it is not converged (that simply mean it is not the optimal localization, but very close to it).

Requires updated mrcpp library.

### DIFF
--- a/.github/mrchem-codecov.yml
+++ b/.github/mrchem-codecov.yml
@@ -6,7 +6,6 @@ dependencies:
   - cxx-compiler
   - eigen
   - lcov
-  - mrcpp
   - ninja
   - nlohmann_json
   - xcfun
--- a/.github/mrchem-gha.yml
+++ b/.github/mrchem-gha.yml
@@ -5,7 +5,6 @@ dependencies:
   - cmake
   - cxx-compiler
   - eigen
-  - mrcpp
   - ninja
   - nlohmann_json
   - xcfun
--- a/doc/conf.py
+++ b/doc/conf.py
@@ -457,7 +457,7 @@ def setup(app):
             '@PROJECT_BINARY_DIR@' : project_root_dir,
             '@PERL_EXECUTABLE@'    : which('perl')
           }
-#    tup = (os.walk(project_src_dir)) 
+#    tup = (os.walk(project_src_dir))
 #    for root, dirs, files in os.walk(project_src_dir):
 #        if 'mwfilters' in dirs:
 #            dirs.remove('mwfilters')
@@ -486,3 +486,6 @@ def setup(app):
         app.connect("builder-inited", generate_doxygen_xml)
     else:
         run_doxygen(project_doc_dir)
+
+# configure sphinxcontrib.bibtex
+bibtex_bibfiles = ["bibliography.bib"]
--- a/doc/requirements.txt
+++ b/doc/requirements.txt
@@ -2,4 +2,4 @@ breathe
 recommonmark
 Sphinx>=2.0
 sphinx_rtd_theme
-sphinxcontrib-bibtex
+sphinxcontrib-bibtex==1.0.0
--- a/doc/users/betzy_example.job
+++ b/doc/users/betzy_example.job
@@ -1,0 +1,9 @@
+#!/bin/bash -l
+#SBATCH --nodes=4
+#SBATCH --tasks-per-node=8
+#SBATCH --cpus-per-task=16
+
+export UCX_LOG_LEVEL=ERROR
+export OMP_NUM_THREADS=16
+
+~/my_path/to/mrchem --launcher='mpirun --rank-by node --map-by numa --bind-to numa' mw6
--- a/doc/users/running.rst
+++ b/doc/users/running.rst
@@ -137,6 +137,35 @@ as it will be literally prepended to the ``mrchem.x`` command when the
     definitely not *more* than one process per orbital).
 
 
+Job example (Betzy)
++++++++++++++++++++
+
+This job will use 4 compute nodes, with 8 MPI processes on each, and each MPI
+process will use 16 OpenMP threads. The flags are optimized for OpenMPI (foss)
+library on Betzy.
+
+.. literalinclude:: betzy_example.job
+
+``--rank-by node``
+  Tells the system to place the first MPI rank on the first node, the second MPI
+  rank on the second node, until the last node, then start at the first node again.
+
+``--map-by numa``
+  Tells the system to map MPI ranks according to NUMA (Non Uniform Memory Access).
+  On Betzy memory configuration groups cores by groups of 16, with cores in the same
+  group having the same access to memory (other cores will have access to that part
+  of the memory too, but slower).
+
+``--bind-to numa``
+  Tells the system to bind cores to one NUMA group. That means that a process will
+  only be allowed to use one of the 16 cores of the group. (The operating system may
+  change the core assigned to a thread/process and, without precautions, it may be
+  assigned to any other core, which would result in much reduced performance). The 16
+  cores of the group may then be used by the threads initiated by that MPI process.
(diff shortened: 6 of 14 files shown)


.github/mrchem-codecov.yml
```diff
@@ -6,7 +6,6 @@ dependencies:
   - cxx-compiler
   - eigen
   - lcov
-  - mrcpp
```
robertodr, line 9: Yep, sorry for the trouble. I added this to make CI faster.

stigrj: still no luck, though :confused:

stigrj:
There seems to be an issue with the Gram-Schmidt orthogonalization `orbital::orthogonalize(double prec, OrbitalVector &Phi)`. The `bin/mrchem-tests [orbital_vector]` test says that the orbitals are not orthogonal after this operation. Quite far from, actually, so this is probably not a precision thing. Complex-valued overlap matrix for four orbitals:
```
                      (1,0)           (0.824794,0.3972)       (-0.209579,-0.262804) (-2.19148e-16,-5.66448e-16)
         (0.824794,-0.3972)                       (1,0)  (5.36696e-15,-1.09495e-15)          (0.23242,0.291446)
       (-0.209579,0.262804)   (5.36696e-15,1.09495e-15)                       (1,0)         (0.822506,0.396098)
 (-2.19148e-16,5.66448e-16)         (0.23242,-0.291446)        (0.822506,-0.396098)                       (1,0)
``` 
This was with a pure OMP build

gitpeterwind:
I only tested pure imaginary  (or real). I did not know how to test fully complex cases. How does one run a single test?

stigrj:
```
bin/mrchem-tests [orbital_vector]
```
This will run only the tests labelled `orbital_vector`, which in this case is everything in the `tests/qmfunctions/orbital_vector.cpp` file.For debugging in the tests, note that the `mrcpp::Printer` is uninitialized, so you have to use `std::cout` here to get output, `println()` will be silent

stigrj:
It is actually still not orthogonalizing correctly with real orbtials

gitpeterwind:
The spin dependency was missing. Now most test passes, but a few give answer almost correct but outside of the error range.
How can I run "H2O_energy_BLYP" myself? `mrchem-tests [h2o_energy_blyp]` did not work

stigrj:
No, the integration tests must be run through ctest
```
ctest -L h2o_energy_blyp --output-on-failure
```
I sometimes get a `segfault` in the `orbital_vector` test if I run with more than a few threads, so I think there's something wrong in the OMP part. Perhaps the flushing thing is not working as expected?

gitpeterwind:
The differences in the 3 failing tests are small, well within world_prec. I have done some extensive testing on h2o_energy_blyp , (including varying the precision and making the scf converge) and I cannot conclude than the old or new version is more accurate. The new version produces more nodes though, so that is an indication that it is actually more conservative. The test uses "numerically_exact = true", and it may be that the new  version uses another interpretation of it.It includes all the nodes of the union grid, but I thought the previous version did too. It would be good to find the reason for the discrepancy. @stigrj , do you have an idea what may cause the differences? Is there a "crop" somewhere in the "linear_combination" procedure for example?

gitpeterwind:
Where is that crop?

gitpeterwind:
I haven't tried too much to get the seg fault. Do you have a recipe ?

gitpeterwind:
I found the crop :)

stigrj:
For the segfault, just a regular build on my laptop: pure OMP, GCC-9.3.0. Then OMP_NUM_THREADS=2 mostly works, OMP=4 sometimes fails, while OMP=8 always fails (sometimes segfault, sometimes it hangs).

gitpeterwind:
> For the segfault, just a regular build on my laptop: pure OMP, GCC-9.3.0. Then OMP_NUM_THREADS=2 mostly works, OMP=4 sometimes fails, while OMP=8 always fails (sometimes segfault, sometimes it hangs).

I just run "CH4_10" 41 orbitals with 128 threads on Betzy using omp version, and no problems (about 90 seconds/ieration).
Will try to install mrchem on my PC, it might well be system dependent.  
Can you give me your .inp file, in case some settings are important?

gitpeterwind:
water runs fine with 8 threads on my PC. You are sure it is not some other problem, like lack of memory?

stigrj:
I think it's just the `orbital_vector` unit test. Try to run it repeatedly

gitpeterwind:
> I think it's just the `orbital_vector` unit test. Try to run it repeatedly

 Now I can reproduce it!

stigrj:
looks like it's the "vector orthogonalize" section

gitpeterwind:
It is the `orbital::calc_overlap_matrix(OrbitalVector &Bra, OrbitalVector &Ket)` which causes the problem.
The omp part is quite simple. It only has one "atomic" line, and the rest are independent threads. If I change the schedule from dynamic to static, the error disappears.

stigrj:
The atomic thing has always confused me. All I know is that it does not do what I like to think it does :slightly_smiling_face:

gitpeterwind:
Ah, found it. A real bug, nothing to do with omp :) , except that it is in a part not used by mpi. The special thing with the test was that the sizes of the orbital vectors where different for Psi and Phi.
Those unit tests are quite useful after all!

gitpeterwind:
I am not sure my change of doc repository has been successfully included the PR: `This commit does not belong to any branch on this repository.`

stigrj:
Where did you see the git error?

gitpeterwind:
> Where did you see the git error?

https://github.com/MRChemSoft/mrchem/pull/359/commits/e5002445f8c29a2a848d94c0fff839fd8257c74a

In the light blue box after "Betzy job example ".

doc/users/Betzy_example.job
```diff
@@ -0,0 +1,11 @@
...
+#SBATCH -A nn4654k
+#SBATCH --nodes=4
+#SBATCH --tasks-per-node=8
+#SBATCH --cpus-per-task=16
+#SBATCH --time=1:00:0
+
+export UCX_LOG_LEVEL=ERROR
+export OMP_NUM_THREADS=16
+
+mpirun --rank-by node -map-by numa -bind-to numa ~/my_path/to/mrchem/dir/mrchem.x mw6.json
```
stigrj, line 11: Should use the `launcher` flag:
```
~/my_path/to/mrchem --launcher='mpirun -rank-by node -map-by numa -bind-to numa' mw6
```

gitpeterwind: ok!

stigrj:
Hmm, it says so on all commits, it seems. I guess it's because the commit currently lives only in your fork, so nothing to worry about

stigrj:
@gitpeterwind do you consider this one ready to go? I will do a rebase to remove the merge commits and squash some of the latest commits. As the banking features continue to grow, we should consider moving it to a separate location.

gitpeterwind:
Yes, It is ready.
I work now with a version that aims to separate the Bank in a new file (under utils).

stigrj (approved):

PR #360, opened 2021-02-17
Bank upgrade

gitpeterwind:
Refactoring of Bank. Main changes:

- Moved all Bank related to own files utils/Bank.cpp
- The usage of the Bank is now through the `BankAccount` class. Once a type `BankAccount` is defined, it can be used similarly to an array (or rather a std::map) to put and get orbitals or data. 
- The `BankAccount` are private, so no interference between different parts of the code. 
- One can open several `BankAccounts` (for Bra and Ket orbitals for example). 
- They are cleaned up automatically (destructor) when getting out of scope.
- Note that `BankAccount` class declaration (creator) is collective, i.e. it must be declared by all MPI at once.  
- Slightly faster

### DIFF
--- a/doc/users/Betzy_example.job
+++ b/doc/users/Betzy_example.job
@@ -1,0 +1,8 @@
+#!/bin/bash -l
+#SBATCH --nodes=4
+#SBATCH --tasks-per-node=8
+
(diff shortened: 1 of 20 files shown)


gitpeterwind:
Trying again

PR #362, opened 2021-02-17
Bank upgrade

gitpeterwind:
Refactoring of Bank. Main changes:

- Moved all Bank related to own files utils/Bank.cpp
- The usage of the Bank is now through the `BankAccount` class. Once a type `BankAccount` is defined, it can be used similarly to an array (or rather a std::map) to put and get orbitals or data. 
- The `BankAccount` are private, so no interference between different parts of the code. 
- One can open several `BankAccounts` (for Bra and Ket orbitals for example). 
- They are cleaned up automatically (destructor) when getting out of scope.
- Note that `BankAccount` class declaration (creator) is collective, i.e. it must be declared by all MPI at once.  
- Slightly faster

### DIFF
--- a/src/utils/Bank.h
+++ b/src/utils/Bank.h
@@ -1,0 +1,95 @@
+#pragma once
+
+#include "MRCPP/Parallel"
+
+#include "mrchem.h"
+#include "parallel.h"
+#include "qmfunctions/qmfunction_fwd.h"
+
+namespace mrchem {
+
+using namespace mpi;
+
+struct deposit {
+    Orbital *orb;
+    double *data; // for pure data arrays
+    bool hasdata;
+    int datasize;
+    int id = -1; // to identify what is deposited
+    int source;  // mpi rank from the source of the data
+};
+
+struct queue_struct {
+    int id;
+    std::vector<int> clients;
+};
+int const CLOSE_BANK = 1;
+int const CLEAR_BANK = 2;
+int const NEW_ACCOUNT = 3;
+int const CLOSE_ACCOUNT = 4;
+int const GET_ORBITAL = 5;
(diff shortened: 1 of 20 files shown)


src/utils/Bank.h
```diff
@@ -0,0 +1,95 @@
...
+int const SET_DATASIZE = 11;
+int const GET_DATA = 12;
+int const SAVE_DATA = 13;
+int const SAVE_NODEDATA = 14;
+int const GET_NODEDATA = 15;
+int const GET_NODEBLOCK = 16;
+int const GET_ORBBLOCK = 17;
+int const CLEAR_BLOCKS = 18;
+int const GETMAXTOTDATA = 19;
+int const GETTOTDATA = 20;
```
robertodr, line 45: Can these be inside an `enum`?

gitpeterwind: I thought of it, but it will make the code only more complicated. I do not see what positive it brings.

stigrj:
This looks like a good step in the right direction. One general problem we have is that the MPI parts are not very well tested, so we should get a few unit tests up and running here. I also had to disable MPI for the integration tests some time ago since they refused to launch correctly after `runtest` was removed. I will try to set up a testing framework for the banking, then I might come back to @gitpeterwind for help with making actual tests.

gitpeterwind:
Yes, a lot of tests including MPI would be useful.  It would however not be a big disadvantage to run those outside github.

gitpeterwind:
I plan also to make new PR relatively soon, which also include some changes on the Bank (I also put the "enum" requested by @robertodr ).
So if there are changes to be made in this PR, it would be better to take it into the new directly.

stigrj (approved):
ok, let's just take this one in

PR #364, opened 2021-02-24
Fast exchange

gitpeterwind:
Exchange (saved, i.e. not on the fly) is twice as fast for large systems. Improvement in the parallelization setup:
- Division into "tasks" and dynamical allocation of task between "workers"
- One task involves several orbital, so that they can be reused (saving communication time)

More upgrade of Bank:
- One of the "bank" process is reserved as "task manager", for dynamical allocation of tasks.
- "id" are no more used as MPI tags, allowing to use any int (it had previously been limited by the mpi's max tag value)
- some refactoring

An example of setup for Betzy, where the bank processes do not "occupy" a full worker process, but only one core instead. This allows to better take advantage of all the cores on a node.


gitpeterwind:
@stigrj , What was missing in the exchangeD1 for full functionality of the not-on-fly setup?

stigrj:
I don't remember. Is it not working?

stigrj:
You said a month ago that the non-diagonal terms were not computed

gitpeterwind:
It is working for normal HF, but in a commit this summer PR you switched it off for some reason. For example what is "dagger"?

stigrj:
Check this discussion: https://github.com/gitpeterwind/mrchem/pull/2

stigrj:
Seems the problem was mixing several OrbitalVectors in the bank at the same time, but this should be seamless now with the new BankAccounts. `dagger()` is the adjoint operator. For response (D2) this means swapping `x_i` and `y_i` terms:
```
K_1.apply(phi_p)  = sum_i   x_i * P[phi_i.dagger() * phi_p] +
                    sum_i phi_i * P[  y_i.dagger() * phi_p]  
K_1.dagger(phi_p) = sum_i phi_i * P[  x_i.dagger() * phi_p] +
                    sum_i   y_i * P[phi_i.dagger() * phi_p]
```

gitpeterwind:
Ah, thanks maybe I start to understand: K.dagger is actually more "K.dagger.apply"?

doc/users/betzy_example.job
```diff
@@ -1,9 +1,8 @@
...
-#SBATCH --tasks-per-node=8
-#SBATCH --cpus-per-task=16
+#SBATCH --tasks-per-node=12
 
 export UCX_LOG_LEVEL=ERROR
-export OMP_NUM_THREADS=16
+export OMP_NUM_THREADS=15
 
-~/my_path/to/mrchem --launcher='mpirun --rank-by node --map-by numa --bind-to numa' mw6
+~/my_path/to/mrchem --launcher='mpirun --rank-by node --map-by numa --bind-to numa --oversubscribe' mw6
```
stigrj, line 8: The launching procedure is getting complicated. We need to make sure it's transferable to other MPI runtimes and machines like Saga where we don't necessarily get the same number of tasks on each node.

gitpeterwind: Yes, this is only for Betzy. 
Note that any "normal" launch will work too, but will not necessarily fully use the resources of the node. ... in this sense this should more be considered as an optimization setting.

gitpeterwind:
The test fails, but the results difference are within world_prec. The new code will compute the exchange terms in a different order, so it is not so surprising. Does this mean I should update the test results?

stigrj:
I can have a look at the tests. Is this otherwise ready?

stigrj:
Will the exchange terms be computed in arbitrary (not reproducible) order?

gitpeterwind:
When run on one MPI, it will be reproducible. (it is very hard to make a code that is not!).
What is there is ready, but I thought having a further look at this dagger.

stigrj:
Is it possible to force it reproducible with `numerically_exact = true`?

gitpeterwind:
You mean that it will be the same, even with many mpi?
(If it is useful I could try, but it means a very different approach: for example forcing one mpi to do all the work, and things like that)

stigrj:
It has proved useful in the past, but it has to make sense also in the code. The changes in the HF energies seems to have been to the better, so I will update the test references.

gitpeterwind:
I was certainly wrong about how complicated it was to include the "numerically exact" flag, one extra line :)

stigrj:
I get this error if I run with the "old" setup `--nodes=1 --ntasks-per-node=8 --cpus-per-task=16` and `mpirun --map-by numa --bind-to numa` on Betzy:
```
Error: /cluster/home/stig/src/mrchem/src/utils/Bank.cpp: openAccount(), line 607: Account id mismatch!
```

gitpeterwind:
Could you give me your .inp file?

stigrj:
`/cluster/projects/nn4654k/stig/benzene.inp`

gitpeterwind:
The id bug was when bank_size=2. In that case the task manager would be run by one of (2) the bank in order not to occupy half of the banks, but that was not tested properly.

gitpeterwind:
For H2O there are no real differences between the precomputed and non-precomputed results. but for Benzene there is a difference is due to different approximation in the integral in the non-symmetric case. I have now forced the two to give the same results.

gitpeterwind:
I think it is ready now

stigrj (approved):

PR #367, opened 2021-03-08
parallel DFT and some small improvements

gitpeterwind:
Parallel DFT , evaluate part (mostly call to XCfun).
Valinomycin PBE is about twice faster per cycle. The memory demand is slighter less, but still too high.
An improvement in memory usage could probably easily be achieved by inserting a crop in the loop that make the final potentials. But there is no "prec" available in DFT (!). 

Requires updated mrcpp (therefore the tests fail for now)

This PR is only superficial: a better approach would be to treat the entire XC potential locally, (see main [idea](https://github.com/MRChemSoft/mrchem/discussions/366) ).
Other changes in this PR: 
- Print out memory and time for the "evaluate" part (was missing)
- make the grid.unify as a tree without coefficient. Nicer... but useless as the memory was anyway allocated but never used and never used memory does not take real place. 
- Small Bank simplification
- More accurate calcExchange_kij
- Longer max line length in .clang_format


stigrj:
Regarding cropping of the final potentials, this is not so easy because of numerical noise. This was the initial reason for using fixed grids for all DFT related functions, the adaptive refinement of the intermediate and final functions did not converge as they were too noisy in the low-density regions, and a post crop would completely remove all useful information in the atomic regions and leave the potentials completely dominated by the far-away noise.

gitpeterwind:
About the same kind of situation (high accuracy everywhere, to get the right precision at some specific parts): The nuclear potential get way to large: 30GB for gramicidin at 1e-7. The potential is very smooth except at the nuclei, so this seems excessive. Even an "on the fly" computation of the potential (just summing the contribution from nuclei at every interpolating point) would be feasible.
How was the prec determined for the nuclear potential? it may be to tight, but I am not sure how it should be set wrt volume of the system, charge of the nuclei, and total charge of the nuclei.

stigrj:
It's been a long time since these parameters were tuned, and they are probably not ideal for large systems. The nuclear correlation factor of Florian that removes the nuclear cusp should help, but possibly at the expense of something else.

gitpeterwind:
How do I change the mrcpp version used by github?

stigrj:
Looks like it is using the MRCPP you are asking for in the latest commit

gitpeterwind:
> Looks like it is using the MRCPP you are asking for in the latest commit

Right! it was the challenging case with one orbital systems which failed.

gitpeterwind:
The hf_grad_lda test fails. It uses `guess_type = CORE_DZ` , and it seems this is making large difference.  Is this numerically stable? I see that (in the correct case also) most orbital energies are positive for example. If I use `guess_type = sad_dz` the results are much closer.

stigrj:
I also get a segmentation fault in the `li_solv` test

src/utils/Bank.cpp
```diff
@@ -590,7 +583,7 @@ void Bank::remove_account(int account) {
 
 int Bank::openAccount(int iclient, MPI_Comm comm) {
     // NB: this is a collective call, since we need all the accounts to be synchronized
-    int account_id = -1;
+    int account_id[1];
```
stigrj, line 586: Compiler complains: uninitialized value in the return if not `MRCHEM_HAS_MPI`

src/qmoperators/one_electron/NuclearOperator.cpp
```diff
@@ -79,16 +80,17 @@ NuclearPotential::NuclearPotential(const Nuclei &nucs, double proj_prec, double
 
     // Scale precision by system size
     int Z_tot = chemistry::get_total_charge(nucs);
-    double abs_prec = proj_prec / Z_tot;
-
+    double abs_prec = proj_prec / std::min(1.0 * Z_tot, sqrt(2 * Z_tot));
+    double myabs_prec = proj_prec / std::max(1.0, mycharge);
```
stigrj, line 84: What is the reasoning behind the `sqrt(2 * Z_tot)`?

stigrj: The square root is theoretical and the factor 2 a safety margin?

src/utils/Bank.cpp
gitpeterwind: Right. Not used if no MPI, but still we do not want any complains.

src/qmoperators/one_electron/NuclearOperator.cpp
gitpeterwind: This is purely heuristic: Trees of 30GB for Nuclear potential are not useful. This reduces the size to more tractable values,  and does not affect the energy significantly. Now we can perform tests at least, but we should come back to this.
Isn't there a sort of plan to use absolute precision at some point?

gitpeterwind: The reduction of abs_prec for large system, can be justified by the fact that for larger volumes (in cubic meters), the total norms will increase.

gitpeterwind:
For the Li test, the crashes comes during read of the initial guess. Somewhat it is not compatible with the new version. However when creating a new initial guess with the new version it will work. Do you have any idea what may have changed for the Li case? (for other systems I used older initial guess without problems)

gitpeterwind:
(I tried to change to older mrcpp, but that did not make a difference)

stigrj:
Oh, it's using a MW guess, maybe the input files were saved at the time the coeff blocks were 8 times too large?

gitpeterwind:
I thought that, but was not able to prove it. (by setting the "large" values back). Could be an error in my tests though. The guess has date 14th feb, so that would be reasonable to think it is in a "large" format

gitpeterwind:
OK, it must be this. The new mrcpp always allocates 64 nodes per chunk, while the old made chunks of 2 MB. But I was not able to correct for this and make it work. Maybe there are other format differences in the "February" version. 
We can update the initial_guess, I guess?

stigrj:
Yes, if we are confident this is the reason, you can just update the input files to the test

gitpeterwind:
Isn't it the `mrchem/tests/li_solv/initial_guess` which should be updated? It works fine when I run `ctest -L li_solv ` myself, but somehow not on github

stigrj:
But only the mac build failed. Strange

gitpeterwind:
wow, it passed! what did you do?

gitpeterwind:
Is it new that one use initial_guess across os? The guess is a binary file, so it is not necessarily compatible

stigrj:
I was not aware that this test used a MW guess, and perhaps it's better to avoid this. I will try to change this

stigrj (approved):

PR #368, opened 2021-03-08
Add option of using atom-number as well as atom-symbol in input

Gabrielgerez:
Add a try-except block on api.py that makes it possible to use both atom-number and atom-symbol as input in the molecule coordinates.


stigrj (commented):
Can you change one of the integration tests to use atomic number rather than symbol in the input file?

Gabrielgerez:
> Can you change one of the integration tests to use atomic number rather than symbol in the input file?

On it

stigrj (approved):
It was not necessary to make a new test out of it, you could have just changed the input file in one of the tests we're already running. Anyway, thanks! :slightly_smiling_face:

PR #370, opened 2021-04-06
Cleanups in one-electron operators

stigrj:
This will step on some ZORA toes, but it should make things simpler and I will help with the rebase.

- Name change: `qmoperators/Rank[Zero/One/Two]TensorOperator` -> `tensor/Rank[Zero/One/Two]Operator`
- Added operator "composition" functionality `P(Q)`. This represents the application of `P` on the operator `Q`, in general not equal to `P*Q`, e.g. if `P` is a `QMDerivative` and `Q` is a `QMPotential`, then `P(Q)` results in the differentiated potential and the derivative does NOT carry through by the chain rule.  If `P` and `Q` are both `QMPotentials`, then they will be combined into a single potential, and the order of multiplication will be different for `P*Q(phi) = P*(Q*phi)` and `P(Q)(phi) = (P*Q)*phi`. If operators cannot be merged, like `P = QMPotential` and `Q = QMSpin`, then it reduces literally to the same `P(Q) == P*Q`.

The following applies only to `one_electron` operators for now, `two_electron` will follow:

- There exists only four kinds of `QMOperator` with no further subclasses:
  - `QMPotential`
  - `QMDerivative`
  - `QMSpin`
  - `QMIdentity`
- This means that all `QMPotentials` are now fundamentally the same, they differ only by how their `real()` and `imag()` parts are computed in the corresponding `TensorOperator` constructor. 
- All `shared_ptr` members in the `TensorOperators` have been removed, e.g. the `MomentumOperator` and `PositionOperator` members of `AngularMomentumOperator`. From now on the subclasses of `RankXxxOperator` should be used _only_ as constructors of their underlying `QMOperators`, there should be no extra features added to these operator specializations. This means that after construction the operator can and should be considered a generic `RankXxxOperator`.
- Following this the kinetic matrix specialization has been moved out of the `KineticOperator` and should be handled outside. The current solution for this is not permanent and should be combined with the ZORA efforts into some general structure for handling the kinetic energy, either in the `FockOperator` or in the `SCFSolver`.


ilfreddy (approved):
A bit difficult for me to evaluate the implications on the current development on ZORA. But I trust you it will not be too much work :-)

PR #371, opened 2021-04-12
ZORA rebase

stigrj:
This reproduces the ZORA energies that @Andersmb reported for He and Ne, i.e. slightly off from ORCA.

Todo:
- [x] Need some entity that can take care of the calculation of the kinetic matrix
- [x] `ZoraPotential` should be reduced to a `QMPotential`
- [ ] Move the operator juggling out of `HelmholtzOperator::apply_zora_v3()`, so that we can use the regular `apply()` instead.


stigrj:
Moving in the right direction, but I broke the Ne ZORA energy...

stigrj:
New improved PR to come

PR #376, opened 2021-06-30
MPI-ify operator composition tests

stigrj:
Minor fixes to make `operator_composition` unit test pass with MPI
```
mpirun -np 4 mrchem-tests [operator_composition]
```


gitpeterwind (approved):
(it would be better to avoid auto)

stigrj:
hmm, not sure why I changed that...

Issue #377, opened 2021-07-07
Bug in open-shell exchange with MPI


stigrj:
Beryllium total energies computed with RHF and UHF, using one and three MPI procs:
```
be_rhf_mpi_1.out- Total energy            :            (au)             -1.457296885087e+01
be_rhf_mpi_3.out- Total energy            :            (au)             -1.457296867897e+01
be_uhf_mpi_1.out- Total energy            :            (au)             -1.457297786710e+01
be_uhf_mpi_3.out- Total energy            :            (au)             -1.590381861487e+01
```

PR #380, opened 2021-07-08
Quick fix: Fall back to on-the-fly exchange for open-shell

stigrj:
The open-shell exchange error #377  is located in `ExchangePotentialD1::setupInternal`, so we circumvent the problem by avoiding `K.setPreCompute()`


gitpeterwind (approved):
ok, did not see what the real problem was, but since it is to be temporary...

stigrj:
I will investigate the problem

PR #382, opened 2021-08-06
Fix wrong size parameter in contracted XC output

### LINKED ISSUE #374: Bug in DFT Hessian calculation with MPI
The new MPI parallelized version of XC Hessian fails with `mpi::orb_size > 1` with error message
```
Error: expand_nodes(), line 173: Size mismatch
```
from `xc_utils.cpp::expand_nodes`. It does not get the expected number of input functions vs data points in the case of XC Hessian.

stigrj:
Another simple bugfix, closes #374


gitpeterwind (approved):
Good work. Bugs are usually difficult to find but easy to correct

PR #383, opened 2021-08-09
Fix uninitialized value in orbital::rotate

### LINKED ISSUE #378: Bug in calculations with more MPI processes than orbitals
Something is not right when using more MPI processes (in the `mpi::comm_orb` communicator) than number of orbitals. Calculations tend to dead lock or crash with error messages like
```
realloc(): invalid next size
[sandel:125065] *** Process received signal ***
[sandel:125065] Signal: Aborted (6)
[sandel:125065] Signal code:  (-6)
[sandel:125065] realloc(): invalid next size
[sandel:125066] *** Process received signal ***
[sandel:125066] Signal: Aborted (6)
[sandel:125066] Signal code:  (-6)
```

I think the error is related to the new and faster `orbital::rotate()` function but I haven't been digging too deep yet.

Similar calculations run smoothly in `v1.0.1`.

stigrj:
Only triggered when size of `mpi::comm_orb` is greater than number of orbitals, i.e. when some MPI ranks does not have any orbitals. Closes #378.

### DIFF
--- a/src/qmfunctions/orbital_utils.cpp
+++ b/src/qmfunctions/orbital_utils.cpp
@@ -260,7 +260,7 @@ OrbitalVector orbital::rotate(OrbitalVector &Phi, const ComplexMatrix &U, double
         }
     }
 
-    IntVector PsihasReIm(2);
+    IntVector PsihasReIm = IntVector::Zero(2);
     for (int i = 0; i < N; i++) {
         if (!mpi::my_orb(Phi[i])) continue;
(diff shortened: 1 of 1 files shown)


gitpeterwind (approved):
Ah, yes. Does the code otherwise run smoothly in the case Norb > NMPI ?
BTW: we should work towards removing fixed own orbitals. It can be replaced by better solutions usually.

stigrj:
Yes, now it seems to run correctly again for all configurations I have tested.

PR #401, opened 2022-03-11
ZORA implementation

Andersmb:
This PR features a ZORA implementation with the possibility of constructing the ZORA potential with `V_nuc`, `J`, and `V_xc`.


stigrj:
The Betzy example managed to creep back in...

Andersmb:
There are two betzy files in that dir differing only in the case. Since my Mac OS is not case sensitive, it messes up my git 😠 What should I do?

stigrj:
How about `rm <the-file-you-dont-want>`? :slightly_smiling_face:

Andersmb:
Some tests fail linux (tested on saga), but passes on my Mac. Perhaps I need to rerun the reference calculations? (but weird that they pass on one system but fail on another?)

src/qmoperators/two_electron/FockBuilder.cpp
```diff
@@ -83,6 +83,7 @@ void FockBuilder::setup(double prec) {
     this->perturbation().setup(prec);
 
     if (isZora()) {
+        println(0, "isZora!!");
```
Andersmb, line 86: Do you want the `println`  in?

Andersmb:
Did the tests fail due to my sloppy code? 😸

src/qmoperators/two_electron/FockBuilder.cpp
stigrj: oops

Andersmb:
Everything works as expected for me. The new relativity string options in `WaveFunction.relativity` reproduce the explicitly set `ZORA.include_` options.

stigrj (approved):

stigrj:
@Andersmb you should continue from a fresh `master` when this is merged

Issue #171, opened 2018-12-15
Proper handling of physical constants


stigrj:
E.g. through https://github.com/MolSSI/QCElemental, but this requires new C++ interface

robertodr:
What about the `mrchem` Python script handling conversions to/from AU at the beginning/end of a calculation? `mrchem.x` could dump the computed data to JSON, re-read it in Python and convert it to whatever units.

Andersmb:
From @stigrj in https://github.com/MRChemSoft/mrchem/pull/404#issuecomment-1078868054 
> I think one nice way to handle the constants is to let Python define them and feed them into the C++ program through the JSON input. Then we also have complete control and can tweak the parameters without recompiling, by intercepting the input parser and edit the JSON file before execution.

Is something like this what you had in mind?


```python
from qcelemental.physical_constants.context import PhysicalConstantsContext
from qcelemental.datum import Datum


# Subclass it here so that the same (non-default) context can easily be used
# when importing the class
# We could also define custom aliases here
class MRChemConstants(PhysicalConstantsContext):
    def __init__(self, context="CODATA2018"):
        super().__init__(context)

        customs = [
            ('pi',      '', 3.1415926535897932384, 'Pi'),
            ('pi_sqrt', '', 1.7724538509055160273, 'Square root of pi')
        ]

        # Add aliases to internally stored constants and make them callable
        for ident, units, value, comment in customs:
            self.pc[ident.lower()] = Datum(ident, units, value, comment=comment)
            self.__setattr__(ident, value)

    def to_json(self):
        """Generate a JSON file storing the constants, which can be passed to the c++ program."""
        return {
            qca.label: float(qca.data) for qca in self.pc.values()
        }
```

And then in `api.py`


```python
import math

from .helpers import *
from .CUBEparser import *
from .periodictable import PeriodicTable as PT, PeriodicTableByZ as PT_Z
from .physical_constants import MRChemConstants

pc = MRChemConstants()


def translate_input(user_dict):
    # get the origin in the desired units of measure
    origin = user_dict["world_origin"]
    if user_dict["world_unit"] == "angstrom":
        origin = [r / pc.bohr2angstroms for r in origin]

    # prepare bits and pieces
    mol_dict = write_molecule(user_dict, origin)
    mpi_dict = write_mpi(user_dict)
    mra_dict = write_mra(user_dict, mol_dict)
    scf_dict = write_scf_calculation(user_dict, origin)
[...]

PR #404, opened 2022-03-25
Input parsing cleaning and PhysConsts from QCelemental

Andersmb:
# Summary of changes

## On the Python side

- Moved the molecule validations out from `python/mrchem/api.py` to a separate class `MoleculeValidator` in `python/mrchem/validators.py`. I think this makes the structure and flow of the input parsing a bit clearer and readable, and cleans the responsibility of `api.py` to converting from user.json to program.json (with some help from `helpers.py`)
- Added support for reading physical constants from `qcelemental`. Now (almost) all constants are derived directly from NIST constants collected in [CODATA2018](https://physics.nist.gov/cuu/Constants/Table/allascii.txt).
- The constants are added to the template file via the utility script `python/mrchem/update_input_parser.py`, which also runs parselglossy and syncs the user ref file. The user can therefore overwrite the constant defaults in the input file for their calculation.
- The constants are passed to `program.json` in the `constants` field, so that the c++ code gains access to these
- Notable differences in how the constants are defined in QCElemental and (old) MRChem:

```
    - electron g factor        : CODATA: -2.00231930436256
                                 MRChem:  2.0023193043618

    - fine structure constant  : CODATA: 7.2973525693e-3
                                 MRChem: 7.2973525664

    - au -> wavenumbers        : CODATA: 219474.6313632
                                 MRChem: 219471.5125976648

    - au -> debye              : CODATA: 2.5417464739297717
                               : MRChem: 2.54174623105
```

## On the c++ side
- Update the interface to physical constants. A new singleton class `chemistry/PhysicalConstants` has been created that stores the constants from `program.json` in a `nlohmann::json` object.

## ToDo
- [x] Add set of constants to c++ so that unit tests pass


stigrj:
I believe this file [python/mrchem/input_parser/plumbing/__init__.py](https://github.com/MRChemSoft/mrchem/pull/404/files#diff-e7dbd0887ae4697a86c6b069f02dc856d381b1709140ce33b87110262e7675eb) should not be committed. It was the one causing problems before (reason still unknown). Are you now able to run with file in source?

Andersmb:
Ah, that one slipped through. Very annoying. Why has this file started causing problems?

Another thing. The `ANG_2_BOHR` conversion factor is present both in `helpers.py` and `CUBEparser.py`. The cube parser cannot import from helpers due to circular import errors. A simple fix solution could be to collect the constants we use during input parsing in its own file, and import from there. Perhaps ideally python and c++ should access the same set of constants (this is perhaps what you mentioned in #171)

Gabrielgerez:
That li_solv has always had weird behavior, I'll check it out

stigrj:
@robertodr do you know why this file suddenly started to appear?

I think one nice way to handle the constants is to let Python define them and feed them into the C++ program through the JSON input. Then we also have complete control and can tweak the parameters without recompiling, by intercepting the input parser and edit the JSON file before execution.

robertodr:
@Andersmb Which version of `pyparsing` is installed locally? `parselglossy` cannot yet deal with v3 of that library, I think that's what you're seeing. I opened a PR sometime ago to fix that https://github.com/dev-cafe/parselglossy/pull/109 Didn't get around to finishing it.

robertodr:
The quick solution is to ignore any change in that `plumbing` when committing.

Andersmb:
> The quick solution is to ignore any change in that `plumbing` when committing.

I have version `3.0.7` installed...

robertodr:
> > The quick solution is to ignore any change in that `plumbing` when committing.
> 
> I have version `3.0.7` installed...

Mystery understood then.

Andersmb:
> @robertodr do you know why this file suddenly started to appear?
> 
> I think one nice way to handle the constants is to let Python define them and feed them into the C++ program through the JSON input. Then we also have complete control and can tweak the parameters without recompiling, by intercepting the input parser and edit the JSON file before execution.

Should we base it on some official set of constants, as `QCElemental` does with the nist files? The nist data files was over 2000 lines, so it contains a lot of stuff we don't need, but gave a lot of neat stuff like descriptions, units, etc. And they don't come with intuitive names, so to actually access common constants we would perhaps need to manually define some type of aliases... I think that is what is done in `QCElemental`...?

Or simply make a python class that holds the set of constants currently present in `mrchem.h`, and define them explicitly, and update this whenever a new constant is needed?

aka
```python
class PhysConstants:
         kJ        = 2625.49962;         // AU -> kJ/mol conversion factor
         kcal      =  627.509469;        // AU -> kcal/mol conversion factor
         eV        =   27.21138505;      // AU -> eV conversion factor
         JT_m2     =   78.9451185;       // AU -> 10^(-30) J/T^2 (SI magnetizability)
         Debye     =    2.54174623105;   // AU -> Debye conversion factor
         cm_m1     =    1.0e7 / 45.5640; // AU -> cm^(-1) frequency -> wavelenght
         alpha     =    7.2973525664;    // Fine structure constant
         g_e       =    2.0023193043618; // Free-electron g-value
         beta_e    =    0.5000000;       // Bohr magneton
         beta_N    =    0.0002723;       // Nuclear magneton
         alpha_inv =  137.035999084;     // Inverse of the fine structure constant alpha
```
(perhaps with more intuitive names and units added in)

Or it should be easy parse the nist file for values and units, and set up a class like above but only with some set of predefined constants that we need..

Andersmb:
@Gabrielgerez I discovered why the `li_solv` was failing. The `Molecule.cavity_coords` section had a duplicated sphere

```json
"cavity_coords": [
        {
          "center": [
            0.0,
            0.0,
            0.0
          ],
          "radius": 4.0
        },
        {
          "center": [
            0.0,
            0.0,
            0.0
          ],
          "radius": 4.0
        }
      ]
```

Just deleting the extra sphere lead to a passing test.

Andersmb:
@stigrj Many unit tests will now fail because they are not being fed the constants generated by the input parser. How best solve this?

stigrj:
Hmm yes, we probably need some defaults on the C++ side. Did you get around the linking issue?

robertodr:
As for the singleton, I would use the pattern given in Alexandrescu's "Modern C++ Design" which uses static local references and avoids spilling around pointers and checking for `NULL`s.

`PhysicalConstants.h`
```c++
namespace mrchem {
class PhysicalConstants final {
  public:
     static PhysicalConstants& getInstance(const nlhomann::json & constants = /* */);
     static auto get(const std::string& key) const {
         return constants_[key];
     }
     PhysicalConstants() = default;
   ~PhysicalConstants() = default;
   PhysicalConstants(const PhysicalConstants&) = delete; 
   PhysicalConstants& operator=(const PhysicalConstants&) = delete;
   PhysicalConstants(PhysicalConstants&&) = delete; 
   PhysicalConstants& operator=(PhysicalConstants&&) = delete;
  private:
     PhysicalConstants(const nlhomann::json & constants) : constants_{constants} {}
     nlohmann::json constants_;
};
}
```

`PhysicalConstants.cpp`
```c++
namespace mrchem {
PhysicalConstants& PhysicalConstants::getInstance(const nlhomann::json & constants)
{
	static PhysicalConstants obj(constants);
	return obj;
}
}
```

The default value for the constants can be declared in an anonymous namespace, I think.

Andersmb:
@robertodr I will update singleton design.

The tests should pass now, but I am not sure if it is the best design. Now some constants are explicitly set in the singleton constructor. I added a method to `MRChemConstants.py` for printing a predefined subset of constants, to help avoid errors (you just need to copy/paste the output into the c++ file).

Andersmb:
@robertodr I am not able to get this to work. After sorting out a few typos, I end up with `Undefined symbols for architecture x86_64:` errors during compilation, ala:
```console
Undefined symbols for architecture x86_64:
  "mrchem::PhysicalConstants::constants_", referenced from:
      mrchem::driver::build_fock_operator(nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer> const&, mrchem::Molecule&, mrchem::FockBuilder&, int) in driver.cpp.o
      mrchem::H_M_pso::H_M_pso(mrchem::MomentumOperator, mrchem::NuclearGradientOperator) in driver.cpp.o
      mrchem::H_MB_dia::H_MB_dia(mrchem::PositionOperator, mrchem::NuclearGradientOperator) in driver.cpp.o
      mrchem::H_BM_dia::H_BM_dia(mrchem::PositionOperator, mrchem::NuclearGradientOperator) in driver.cpp.o
      mrchem::AngularFunction::calcConstant() const in HydrogenFunction.cpp.o
      mrchem::chemistry::compute_nuclear_density(double, mrchem::Nuclei const&, double) in chemistry_utils.cpp.o
      mrchem::SCFEnergy::print(std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> > const&) const in Molecule.cpp.o
      ...
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
```

I currently have in `PhysicalConstants.h`

```c++
#pragma once
#include <nlohmann/json.hpp>

namespace mrchem {
class PhysicalConstants final {
  public:
     static PhysicalConstants& getInstance(const nlohmann::json &constants);
     static double get(const std::string& key) {
         return constants_[key];
     }
     PhysicalConstants() = default;
   ~PhysicalConstants() = default;
   PhysicalConstants(const PhysicalConstants&) = delete; 
   PhysicalConstants& operator=(const PhysicalConstants&) = delete;
   PhysicalConstants(PhysicalConstants&&) = delete; 
   PhysicalConstants& operator=(PhysicalConstants&&) = delete;

  private:
     PhysicalConstants(const nlohmann::json & constants) { constants_ = constants; }
     static nlohmann::json constants_;
};
}
```

and in `PhysicalConstants.cpp`

```c++
#include "PhysicalConstants.h"
#include <nlohmann/json.hpp>

namespace mrchem {
PhysicalConstants& PhysicalConstants::getInstance(const nlohmann::json & constants) {
	static PhysicalConstants obj(constants);
	return obj;
}
}
```

I think it is means that the private static member `constants_` is declared but not defined anywhere. Perhaps I broke the class when trying to get rid of the compiler errors after copy/pasting...

stigrj:
I guess this is not in the latest commit you pushed, because that one work fine for me?

Two comments:
- The `constants` input section turned out a bit massive. Can we limit to include only the constants we need?
- The `light_speed` parameter in the ZORA operators should perhaps be moved into this section? But then how should we treat constants that are derived from others in case we want to change the default? One example would be `light_speed` vs `fine_structure_constant`: if we change the speed of light in a ZORA calculation, should we then change the fine-structure constant accordingly in e.g. the NMR operators?

Andersmb:
@stigrj I updated the python side  so that only the constants we need are added to the program input. The "constants" section now looks like this:

```
"constants": {
      "angstrom2bohrs": 1.8897261246257702,
      "atomic_unit_of_bohr_magneton": 0.5000000000000764,
      "atomic_unit_of_nuclear_magneton": 0.0002723085107443953,
      "c_au": 137.035999084,
      "dipmom_au2debye": 2.5417464739297717,
      "electron_g_factor": -2.00231930436256,
      "fine_structure_constant": 0.0072973525693,
      "hartree2ev": 27.211386245988,
      "hartree2kJmol": 2625.4996394798254,
      "hartree2kcalmol": 627.5094740630558,
      "hartree2simagnetizability": 78.9451185,
      "hartree2wavenumbers": 219474.6313632,
      "pi": 3.141592653589793,
      "pi_sqrt": 1.7724538509055159
    }
```

I see no mention in the ORCA manual that changing the light speed in a relativistic calculation affects any other aspect of that code. But I am not sure what makes sense.

Andersmb:
Also, how do I add `qcelemental` as a python dependency?  As of now, the source of everything needed on the python side (that is not part of the standard library) is explicitly included, right, to avoid having the user install packages themselves? Do we do the same with `qcelemental`?

stigrj:
To fix the tests on GitHub Actions I think you just have to add `qcelemental` to `.github/mrchem-gha.yml`. For the general user I'm not sure how we want to do this, perhaps it's not too much to ask of the user to install a simple python package? @robertodr ?

robertodr:
It's not too onerous to install a Python package. Consider though that we removed the dependency on `runtest` to make development work run smoother. `runtest` was used to only run tests and we could rewrite that functionality in pure Python. `qcelemental` might be used to do more than just supply units of measure so it is a slightly different use case. If there's a way to make the dependency optional, you'd make a few developers happier 😄

stigrj:
We could use `qcelemental` in the same way we use `parselglossy`, e.i. only by developers once in a while to update a JSON file with the parameters we need. Or, it could be used as part of the `parselglossy` workflow of generating the input template, so that default values are fetched directly from `qcelemental` when we run `parselglossy`?

robertodr:
> We could use `qcelemental` in the same way we use `parselglossy`, e.i. only by developers once in a while to update a JSON file with the parameters we need. Or, it could be used as part of the `parselglossy` workflow of generating the input template, so that default values are fetched directly from `qcelemental` when we run `parselglossy`?

sounds reasonable, yes. As long as you don't dump all constants and conversion factors the JSON shouldn't be too large.

Andersmb:
@robertodr 

I have updated the singleton class, but I still get problems during linking:

```make
[ 76%] Linking CXX executable ../bin/mrchem-tests
Undefined symbols for architecture x86_64:
  "mrchem::PhysicalConstants::constants_", referenced from:
      mrchem::PhysicalConstants::PhysicalConstants(nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer> const&) in PhysicalConstants.cpp.o
Undefined symbols for architecture x86_64:
  "mrchem::PhysicalConstants::constants_", referenced from:
      mrchem::PhysicalConstants::PhysicalConstants(nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer> const&) in PhysicalConstants.cpp.o
ld: symbol(s) not found for architecture x86_64
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/mrchem.x] Error 1
make[2]: *** [bin/mrchem-pilot.x] Error 1
make[1]: *** [src/CMakeFiles/mrchem.x.dir/all] Error 2
make[1]: *** Waiting for unfinished jobs....
make[1]: *** [pilot/CMakeFiles/mrchem-pilot.x.dir/all] Error 2
Undefined symbols for architecture x86_64:
  "mrchem::PhysicalConstants::constants_", referenced from:
      mrchem::PhysicalConstants::PhysicalConstants(nlohmann::basic_json<std::__1::map, std::__1::vector, std::__1::basic_string<char, std::__1::char_traits<char>, std::__1::allocator<char> >, bool, long long, unsigned long long, double, std::__1::allocator, nlohmann::adl_serializer> const&) in PhysicalConstants.cpp.o
ld: symbol(s) not found for architecture x86_64
clang: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [bin/mrchem-tests] Error 1
make[1]: *** [tests/CMakeFiles/mrchem-tests.dir/all] Error 2
make: *** [all] Error 2
```

Do you understand why the linker complains? The updated code has been pushed.

Andersmb:
The user can now overwrite the constant defaults in the user input, e.g.

```
Constants {
    pi = 3
}
```

The keywords to the new input section are added automatically by `python/mrchem/update_input_parser.py`, which reads the template and build the `Constants` section based on the constants defined in `MRChemPhysConstants`. It also runs parselglossy and syncs the user ref file.

python/README.md
```diff
@@ -1,12 +1,44 @@
...
+and should be added manually to the template.
+
+Run parselglossy to update the input parser:
 
 ``` bash
 $ cd python
 $ parselglossy generate --template template.yml --docfile user_ref.rst --doc-header="User input reference" --target="mrchem/input_parser"
 ```
-Remember to also update the documentation:
+Copy the user reference file to update the documentation:
```
stigrj, line 41: I guess if we just add a keyword in an input section unrelated to the constants, we can still update the input in the same way as before, by just running `parselglossy` manually like this?

Andersmb: Yes, if the constants are unchanged then one can run parselglossy the same way as before.

doc/users/user_ref.rst
```diff
@@ -863,4 +866,91 @@ User input reference
...
+  
+    **Type** ``float``
+  
+    **Default** ``1.8897261246257702``
+  
+   :atomic_unit_of_light_speed: Speed of light in atomic units (unit: )
+  
+    **Type** ``float``
+  
+    **Default** ``137.035999084``
```
stigrj, line 913: I think it would be good to add to the docstring exactly what parts of the code will be affected by changing any of these parameters. E.g. changing the light speed will only affect the relativistic Hamiltonians (so any NREL calculation is unaffected), `hartree2simagnetizability` will only change the printed value of the magnetizability property, etc.

Andersmb: I have updated the way the constants are defined in `MRChemPhysConstants`. It is more verbose, but I think it is easier to add new constants than the way it was before. I attempted to mention affected code for each constant, but you should probably double check.

stigrj: Excellent! Are we closing in on a finished version?

Andersmb: Yes, I think so. There are just a few small formatting issues in the user ref, I just noticed:

 - The title includes the quotes: `"User input reference"`
 - I wanted to add multiline docstrings in `Constants` by using the literal `|` character, but it did not work properly. How should I format the string in Python to ensure that it gets converted to the proper yaml literal string format?

This:

```
docstring = f"""|\nConversion factor from Hartree to kcal/mol  (unit: {unit}).
        Affected code: Printed value of energies."""
```

did not work as I had hoped..

stigrj: Similar discussion here: https://github.com/dev-cafe/parselglossy/pull/52

So line breaks are explicitly replaced with white-spaces to avoid errors in the rendering of the docs?

Andersmb: Also, for some reason the units are repeated. They occur at the position specified in the string and also at the end of every string...

Andersmb: I see. I just made the docstrings into normal strings with a `| `in front. From an online ReStructured viewer it looks like they are rendered okay.

Andersmb:
Still some bugs floating around.

- [ ]

Andersmb:
Is it expected that the total energy is identical for one calculation with normal `pi` and one with `pi = 1000000.0`?

stigrj:
That sounds strange. Maybe it's not used? There's also a `pi` in MRCPP, so maybe that is the one that is actually used?

stigrj:
Come to think of it, maybe `pi` should be fixed and provided by MRCPP? Then we leave only physical constants to MRChem

Andersmb:
`HelmholtVector::apply` uses pi, but this method is not currently being called during a normal SCF (we do a `H()` instead). 
pi is also referenced in `Cavity.cpp` and `SCRF.cpp`, so it might be relevant there.
And s unit tests fail when drastically changing the value of pi :)

stigrj:
I see MRCPP's pi is used once in `tests/qmfunctions/orbital_vector.cpp`. Can you just replace all the pi's with `mrcpp::pi` and remove it from MRChem's constants?

python/template.yml
```diff
@@ -1,877 +1,894 @@
...
+    default: [0.0, 0.0, 1.0]
+    predicates:
+    - len(value) == 3
+    docstring: |
+      Third boundary vector for plot.
+- name: MPI
+  docstring: |
+    Define MPI related parameters.
+  keywords:
+  - name: numerically_exact
```
stigrj, line 164: I see the indentation has changed in the entire `template.yml` file, is this intentional?

src/mrchem.cpp
```diff
@@ -55,6 +56,12 @@ int main(int argc, char **argv) {
     const auto &mol_inp = json_inp["molecule"];
     const auto &scf_inp = json_inp["scf_calculation"];
     const auto &rsp_inp = json_inp["rsp_calculations"];
+    const auto &con_inp = json_inp["constants"];
+
+    // Instantiate the physical constants singleton
+    PhysicalConstants::Initialize(con_inp);
+    int print_prec = json_inp["printer"]["print_prec"];
+    if (json_inp["printer"]["print_constants"]) PhysicalConstants::Print(print_prec);
```
stigrj, line 64: I'm not sure about the `print_prec` for this output. If you want to print it (`print_constants == true`), then you probably want the exact value that is used? There's also plenty of space on the line here, so I suggest we just fix it to `print_prec = 14` decimal places here.

src/chemistry/PhysicalConstants.cpp
```diff
@@ -0,0 +1,76 @@
...
+        if (item.key().length() > w) w = item.key().length() + 1;
+    }
+
+    mrcpp::print::header(0, "Physical Constants");
+    for (const auto &item : PhysicalConstants::constants_.items()) {
+        std::string key = item.key();
+        double val = item.value().get<double>();
+        print_utils::scalar(0, key, val, "", pprec, true);
+    }
+    mrcpp::print::separator(0, '=');
```
stigrj, line 73: Can you add some more space between this output and the next:
```
mrcpp::print::separator(0, '=', 2);
```

src/utils/print_utils.cpp
```diff
@@ -122,7 +122,7 @@ void print_utils::coord(int level, const std::string &txt, const mrcpp::Coord<3>
 void print_utils::scalar(int level, const std::string &txt, double val, const std::string &unit, int p, bool s) {
     if (p < 0) p = Printer::getPrecision();
     int w0 = Printer::getWidth() - 2;
-    int w1 = w0 * 2 / 9;
+    int w1 = w0 * 2 / 10;
```
stigrj, line 125: This misaligned some of the `:`'s in the output, but with the old value the name `hartree2simagnetizability` becomes too long... Perhaps abbreviate magnetizability?

python/mrchem/update_input_parser.py
```diff
@@ -0,0 +1,116 @@
...
+# You should have received a copy of the GNU Lesser General Public License
+# along with MRChem.  If not, see <https://www.gnu.org/licenses/>.
+#
+# For information on the complete list of contributors to MRChem, see:
+# <https://mrchem.readthedocs.io/>
+#
+
+
+from pathlib import Path
+from ruamel.yaml import YAML
```
stigrj, line 30: `ruamel.yaml` not listed in dev-packages

stigrj (changes requested):

python/template.yml
Andersmb: hmm, this must be an effect of the `yaml->dict->yaml` roundtrip with `ruamel.yaml`. There are some options for setting indentation - I can see if I can reproduce the original indentation scheme.

Andersmb: I think the following gets the old indentation (source: https://yaml.readthedocs.io/en/latest/detail.html#:~:text=yaml.indent(mapping%3D2%2C%20sequence%3D4%2C%20offset%3D2))

```python
yaml.indent(sequence=4, mapping=2, offset=2)
```

Excerpt:

```yaml
keywords:
  - name: world_prec
    type: float
    predicates:
      - 1.0e-10 < value < 1.0
    docstring: |
      Overall relative precision in the calculation.
  - name: world_size
    type: int
    default: -1
    predicates:
      - value <= 10
    docstring: |
      Total size of computational domain given as 2**(``world_size``). Always cubic
      and symmetric around the origin. Negative value means it will be computed
      from the molecular geometry.
  - name: world_unit
    type: str
    default: bohr
    predicates:
      - value.lower() in ["bohr", "angstrom"]
    docstring: |
      Length unit for *all* coordinates given in user input. Everything will be
      converted to atomic units (bohr) before the main executable is launched,
      so the JSON input is *always* given in bohrs.
  - name: world_origin
    type: List[float]
    default: [0.0, 0.0, 0.0]
    predicates:
      - len(value) == 3
    docstring: |
      Global gauge origin of the calculation.
sections:
  - name: Precisions
    docstring: |
      Define specific precision parameters.
    keywords:
      - name: nuclear_prec
        type: float
        default: user['world_prec']
        predicates:
          - 1.0e-10 < value < 1.0
        docstring: |
          Precision parameter used in smoothing and projection of nuclear potential.
      - name: poisson_prec
        type: float
        default: user['world_prec']
        predicates:
          - 1.0e-10 < value < 1.0
        docstring: |
          Precision parameter used in construction of Poisson operators.
      - name: exchange_prec
        type: float
        default: -1.0
        docstring: |
          Precision parameter used in construction of Exchange operators.
          Negative value means it will follow the dynamic precision in SCF.
      - name: helmholtz_prec
        type: float
        default: -1.0
        docstring: |
          Precision parameter used in construction of Helmholtz operators.
          Negative value means it will follow the dynamic precision in SCF.
```

PR #413, opened 2022-05-20
Add periodic methods to MRCHEM

bjorgve:
Periodic MRCHEM capabilities. For this to work we need to wait for https://github.com/MRChemSoft/mrcpp/pull/197

We follow Pask et. al. 2018 for the implementation.

@Andersmb can you look over my changes in `python/mrchem`


python/mrchem/validators.py
```diff
@@ -211,6 +216,17 @@ def validate_atomic_coordinates(self):
 
         return labels, coords
 
+    def calculate_max_rc(self):
```
Andersmb, line 219: The idea with `MoleculeValidator` was to keep all sanity checks on the molecular geometry and cavity inputs in a single place. Since `calculate_max_rc` is not used in any type of input validation, perhaps the function belongs in `helpers.py`?

Ideally just a single instance of the validator is needed. Perhaps an instance could be created in `api.py.translate_input`, and then this instance gets passed to `write_scf_calculation` and subsequently `write_scf_fock` where you need access to `MoleculeValidator.atomic_coords` for computing the max RCs?

Although I guess there is no real problem with doing it the way you are doing it, by creating new instances where needed :) But the duplication makes it feel like something is not optimal.

Andersmb, line 219: The idea with `MoleculeValidator` was to keep all sanity checks on the molecular geometry and cavity inputs in a single place. Since `calculate_max_rc` is not used in any type of input validation, perhaps the function belongs in `helpers.py`?

Ideally just a single instance of the validator is needed. Perhaps an instance could be created in `api.py.translate_input`, and then this instance gets passed to `write_scf_calculation` and subsequently `write_scf_fock` where you need access to `MoleculeValidator.atomic_coords` for computing the max RCs?

Although I guess there is no real problem with doing it the way you are doing it, by creating new instances where needed :) But the duplication makes it feel like something is not optimal.

python/template.yml
```diff
@@ -31,6 +36,55 @@ keywords:
...
+          - len(value) == 3
+        docstring: |
+          primitive 3 vector.
+      - name: operator_root
+        type: int
+        default: 0
+        predicates:
+          - value <= 0
+        docstring: |
+          operator_root
```
Andersmb, line 75: Some of these docstrings are not that helpful 😛

Andersmb, line 75: Some of these docstrings are not that helpful 😛

ilfreddy:
Work on PBC is suspended for now

PR #421, opened 2022-08-17
3 small bug fixes

gitpeterwind:
MPI_CXX_DOUBLE_COMPLEX is not supported on some machines

3 "asserts" were stopping the code unnecessarily (only in debug mode).

Typo in history rotation (not used by default)

### DIFF
--- a/src/parallel.cpp
+++ b/src/parallel.cpp
@@ -273,7 +273,7 @@ void mpi::allreduce_vector(DoubleVector &vec, MPI_Comm comm) {
 void mpi::allreduce_vector(ComplexVector &vec, MPI_Comm comm) {
 #ifdef MRCHEM_HAS_MPI
     int N = vec.size();
-    MPI_Allreduce(MPI_IN_PLACE, vec.data(), N, MPI_CXX_DOUBLE_COMPLEX, MPI_SUM, comm);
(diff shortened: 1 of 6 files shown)


src/parallel.cpp
```diff
@@ -273,7 +273,7 @@ void mpi::allreduce_vector(DoubleVector &vec, MPI_Comm comm) {
 void mpi::allreduce_vector(ComplexVector &vec, MPI_Comm comm) {
 #ifdef MRCHEM_HAS_MPI
     int N = vec.size();
-    MPI_Allreduce(MPI_IN_PLACE, vec.data(), N, MPI_CXX_DOUBLE_COMPLEX, MPI_SUM, comm);
```
stigrj, line 276: Not sure about this one. The correct option for `std::complex<double>` should be `MPI_CXX_DOUBLE_COMPLEX`, while `MPI_C_DOUBLE_COMPLEX` refers to C99's `_Complex double`, as discussed here: https://github.com/cyclops-community/ctf/issues/10

But then again (as also mentioned in the thread above), it's perhaps better to have something that works but is not according to the standard, rather than something that is _supposed_ to work but doesn't.

gitpeterwind: We use C for MPI. The MPI_C_DOUBLE_COMPLEX is the official (table 3.2 in MPI 3.1) name. MPI_CXX_DOUBLE_COMPLEX is for C++ (table 3.4).
.. and as you say, the argument that gets over all other arguments, it works :)

stigrj (approved):

PR #422, opened 2022-08-27
Add alternative way of running MRChem in Betzy to docs

Gabrielgerez:
As @gitpeterwind has mentioned we could run in Betzy in a more general way, which is what i tried to show here in the docs.

### DIFF
--- a/doc/users/betzy_example.job
+++ b/doc/users/betzy_example.job
@@ -5,4 +5,4 @@
 export UCX_LOG_LEVEL=ERROR
 export OMP_NUM_THREADS=15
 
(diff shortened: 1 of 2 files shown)


Gabrielgerez:
Note to everyone, i am not entirely sure of the options in the new command

gitpeterwind:
Actually, the second "cyclic" , does not work exactly as I expected, so I still try to improve

stigrj:
@gitpeterwind is this tested and ready?

gitpeterwind:
> @gitpeterwind is this tested and ready?

The constant is tested (though not for h2o, which is a bit small), but not how it looks in the docs.

stigrj (approved):

PR #423, opened 2022-08-30
add reaction energy to GroundStateSolver printout

Gabrielgerez:


src/scf_solver/GroundStateSolver.cpp
```diff
@@ -93,6 +95,7 @@ void GroundStateSolver::printProperty() const {
     printUpdate(2, " Coulomb energy  ", J_1, J_1 - J_0, this->propThrs);
     printUpdate(2, " Exchange energy ", K_1, K_1 - K_0, this->propThrs);
     printUpdate(2, " X-C energy      ", XC_1, XC_1 - XC_0, this->propThrs);
+    printUpdate(2, "Reaction energy      ", Er_1, Er_1 - Er_0, this->propThrs);
```
stigrj, line 98: Please be kind to my ocd and add a space at the beginning of the line

stigrj (changes requested):

stigrj:
Hmm, what's up with the failing tests?

Gabrielgerez:
> Hmm, what's up with the failing tests?
Whatever this might be?
```
terminate called after throwing an instance of 'std::length_error'
  what():  basic_string::_M_create
```

src/scf_solver/GroundStateSolver.cpp
Gabrielgerez: Will push a change with a better formatted version

stigrj:
I get the same errors locally. Investigating :mag_right:

src/scf_solver/GroundStateSolver.cpp
```diff
@@ -93,6 +100,10 @@ void GroundStateSolver::printProperty() const {
     printUpdate(2, " Coulomb energy  ", J_1, J_1 - J_0, this->propThrs);
     printUpdate(2, " Exchange energy ", K_1, K_1 - K_0, this->propThrs);
     printUpdate(2, " X-C energy      ", XC_1, XC_1 - XC_0, this->propThrs);
+    mrcpp::print::separator(0, '-');
+    printUpdate(2, " Reaction energy (el)  ", Er_1, Er_1 - Er_0, this->propThrs);
+    printUpdate(2, " Reaction energy (nuc) ", Er_el_1, Er_el_1 - Er_el_0, this->propThrs);
+    printUpdate(2, " Reaction energy (tot) ", Er_nuc_1, Er_nuc_1 - Er_nuc_0, this->propThrs);
```
stigrj, line 106: The ordering is messed up, `(el) -> (tot)` `(nuc) -> (el)` `(tot) -> (nuc)`. Perhaps enclose it in an `if > zero` to avoid printing a bunch of zeros in case we are not doing PCM, similar to what is done in the `SCFEnergy` print. I see now that also the external field energy contribution is missing, could you add that one as well?

stigrj: The test error comes from this string being too long, currently limited to 20 chars. Fix coming up.

stigrj (changes requested):

src/scf_solver/GroundStateSolver.cpp
```diff
@@ -93,6 +100,10 @@ void GroundStateSolver::printProperty() const {
     printUpdate(2, " Coulomb energy  ", J_1, J_1 - J_0, this->propThrs);
     printUpdate(2, " Exchange energy ", K_1, K_1 - K_0, this->propThrs);
     printUpdate(2, " X-C energy      ", XC_1, XC_1 - XC_0, this->propThrs);
+    mrcpp::print::separator(0, '-');
```
stigrj, line 103: Should only be visible from `print_level = 2`, so first argument `2`.

src/scf_solver/GroundStateSolver.cpp
Gabrielgerez: > The ordering is messed up, `(el) -> (tot)` `(nuc) -> (el)` `(tot) -> (nuc)`. Perhaps enclose it in an `if > zero` to avoid printing a bunch of zeros in case we are not doing PCM, similar to what is done in the `SCFEnergy` print. I see now that also the external field energy contribution is missing, could you add that one as well?

woops, fixed

PR #424, opened 2022-09-26
Prepare MRChem-v1.1

stigrj:
`release/1.1` has been branched off the current `master`, and this PR basically just *removes* all ZORA-related features which will instead appear in the next `v1.2` version.

See `CHANGELOG` for added features relative to `v1.0`

The "Update docs" commit will be cherry-picked into `master` after the merge, everything else will remain local to the `release/1.1` branch.

The new solvent feature could have been described explicitly in the User's Manual, but that would require some work. At least it is reasonably well documented in the input reference.

Also the use of MRChem through QCEngine could have been documented somewhere.


doc/users/user_inp.rst
```diff
@@ -419,14 +420,23 @@ respectively. The matrix is constructed in a small AO basis with a given
...
+
 The ``core`` and ``sad`` guesses are fully specified with the following keywords
 (defaults shown):
 
 .. code-block:: bash
 
     SCF {
       guess_prec = 1.0e-3                   # Numerical precision used in guess
-      guess_type = sad_dz                   # Type of inital guess (chk, mw, gto, core, sad)
+      guess_type = sad_gto                  # Type of inital guess (chk, mw, gto, core, sad)
```
gitpeterwind, line 438: It is unclear (for me), how the "sad" and "gto" combines. Are the valid option: sad, sad_dz, sad_tz,sad_qz, sad_gto?
are "gto", "sad", "sad_gto" valid? equal?

gitpeterwind: I read once more, and maybe I understand, the valid options are:
sad, sad_dz, sad_tz,sad_qz, sad_gto, gto
is that correct? then the comment could be changed to 
`# Type of inital guess (chk, mw, gto, core, sad, sad_XX) `

gitpeterwind (approved):
Great that is ready!

robertodr (approved):

doc/users/user_inp.rst
stigrj: Yes, this is a bit confusing. `sad_gto` is a version of SAD (SAD in a GTO AO basis), and is not related to the `gto` guess which reads pre-computed GTO MOs. SAD requires the specification of an AO basis, and valid options are `sad_X`, with `X={sz, dz, tz, qz, gto}`, where the first four refers to a hydrogenic basis set with the given zeta quality, while the latter uses 3-21G.

stigrj: ACtually, the pure `gto` guess could be removed from the manual to avoid confusion. This guess is not really supported because there's currently no way of generating such input files...

PR #432, opened 2022-10-06
Unspaghettify CUBEparser.py

Gabrielgerez:
Small refactoring of CUBEparser.py. Using loops instead of explicit repeated lines.


python/mrchem/CUBEparser.py
```diff
@@ -29,42 +29,51 @@
...
 global pc
+global world_unit
+
+def make_cube_vector(file_key, file_val, vector_dir):
+    data_type = "_".join(file_key.split("_")[2:])
+    path_list = sort_paths(file_val)
+    cube_list = []
+    
+    if not os.path.isdir(vector_dir):
+        os.mkdir(vector_dir)
```
robertodr, line 40: The standard `Pathlib` module should be used instead of string manipulations and `os` to handle interactions with the filesystem. https://docs.python.org/3/library/pathlib.html For example:

```suggestion
    if not Path(vector_dir).is_dir():
        Path(vector_dir).mkdir()
```

python/mrchem/CUBEparser.py
```diff
@@ -23,61 +23,54 @@
...
 
 from .input_parser.plumbing import pyparsing as pp
 
 global pc
+global world_unit
 
 
-def write_cube_dict(user_dict):
+def write_cube_vectors(user_dict):
+    global pc
```
Andersmb, line 36: Do you need `global pc` in the function definition when it is made global outside?

python/mrchem/CUBEparser.py
```diff
@@ -23,61 +23,54 @@
...
+            cube_list = []
+            
+            if not vector_dir.is_dir():
+                vector_dir.mkdir()
+
+            if len(path_list) != 0:
+                for path in path_list:
+                    cube_list.append(parse_cube_file(path, world_unit))
+            
+            for d in cube_list:
```
Andersmb, line 56: Is this a debugging print statement?

python/mrchem/CUBEparser.py
Gabrielgerez: I have replaced all `os` uses with path.

python/mrchem/CUBEparser.py
Gabrielgerez: yeah, I introduced some bugs last commit, fixed them and removed the print statements now

python/mrchem/CUBEparser.py
Gabrielgerez: Making it global outside doesn't seem to do anything, for `pc` so we need to declare it global on the inside. Though, it does do something for `world_unit`, which confuses me. Do you have any idea of what is going on with these variables?

python/mrchem/CUBEparser.py
```diff
@@ -23,68 +23,54 @@
 # <https://mrchem.readthedocs.io/>
 #
 
-import os
-from json import dump
+from json import dumps
+from pathlib import Path
 
 from .input_parser.plumbing import pyparsing as pp
```
Gabrielgerez, line 29: I have an issue here, by itself this line works when running the code normally, but not if I'm running the `cube_parser_test`. If i instead remove the dot, it works in the test but not normally.

tests/cube_parser_test/test
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+
+from operator import is_
+import sys
+import numpy as np
+from pathlib import Path
+
+sys.path.append(str(Path(__file__).resolve().parents[2] / "python" / "mrchem" ))
+print(str(Path(__file__).resolve().parents[2] / "python" / "mrchem" ))
+from CUBEparser import parse_cube_file
```
robertodr, line 10: Try like this:

```suggestion
sys.path.append(str(Path(__file__).resolve().parents[2] / "python" ))
from mrchem.CUBEparser import parse_cube_file
```

tests/cube_parser_test/test
```diff
@@ -0,0 +1,47 @@
...
+    'electron_g_factor': -2.00231930436256,
+    'dipmom_au2debye': 2.5417464739297717}
+
+    world_unit = "bohr"
+    test_files = ["phi_a_test_idx_0_re.cube" ,"phi_a_test_idx_1_re.cube", "phi_a_test_idx_2_re.cube", "phi_a_test_idx_3_re.cube"]
+    parsed_cubes = [parse_cube_file(Path(test_path), world_unit, pc) for test_path in test_files]
+
+
+    for elem in parsed_cubes:
+        is_same = np.array_equal(voxel_reference, np.array(elem["CUBE_data"][0]))
```
robertodr, line 37:
```suggestion
    ierr = 0
    for elem in parsed_cubes:
        passed, message = compare_values(elem["CUBE_data"][0], voxel_reference, "Parsed data from cube file vs. reference data", rtol=1.0e-6, atol=1.0e-6,)
        sys.stdout.write(f"{message}\n")
        ierr += 0 if passed else 137v
```

robertodr (commented):
I've left some suggestions in the test script, but I think they need some additional work.
You don't need to define a `main` function: you save the return status of the comparison in a variable `ierr` and then call `sys.exit(ierr)` at the end. Non zero exit status means a test failure.
Look at the HF gradient test script for inspiration 😄

tests/cube_parser_test/test
Gabrielgerez: what does the `# isort:skip` do?

robertodr: `isort` reorders import statements according to some (semi)standard rules. Lines with `# isort:skip` comments are ignored by `isort`. In this case it's needed because the rules are that import statements should come before anything else in a Python script, but here that would lead to "module not found" errors.

robertodr:
Also, you should run `black` (for formatting) and `isort` (for reordering imports) in the Python files that are modified in this PR.

tests/cube_parser_test/test
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+
+from operator import is_
+import sys
+import numpy as np
```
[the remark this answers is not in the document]
robertodr: And I can't see `is_` being used anywhere, so maybe it's a leftover?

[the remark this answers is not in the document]
Gabrielgerez: yeah, I haven't set my text editor that well yet, sometimes it autocompletes some weird import statements. I think it is gone now

Gabrielgerez:
> Looks good. Do we already have a test for a calculation starting from a cube guess? If not, I suggest to add it, but in a separate PR.

that was the plan :)

PR #434, opened 2022-10-25
Add cube guess test

Gabrielgerez:
also add a line which reduces clutter


robertodr (approved):
Looks good to me. But do we need to have the `cube_vectors/CUBE_p_vector.json` in the Git repository? Isn't it enough that the cube files are there?

Gabrielgerez:
> Looks good to me. But do we need to have the `cube_vectors/CUBE_p_vector.json` in the Git repository? Isn't it enough that the cube files are there?
Yeah it should be created when running ctest, so no need for it, let me remove it and push the changes

PR #435, opened 2022-10-26
Cube guess rsp

Gabrielgerez:
adding cubes as a guess to response calculations


doc/users/user_ref.rst
```diff
@@ -566,6 +566,42 @@ User input reference
   
     **Default** ``initial_guess/phi_b``
   
+   :guess_cube_x_p: File name for paired response orbitals, used with ``cube`` guess. Expected path is ``<path_orbitals>/x_p_rsp_<0,1,2>_<0...Np>_<re/im>.cube
```
robertodr, line 569: You should explain what `<0, 1, 2>` mean (the perturbation components?) and `<0...Np>` mean (the number of occupied orbitals?)

doc/users/user_ref.rst
```diff
@@ -642,7 +678,7 @@ User input reference
   
     **Type** ``str``
   
-    **Default** ``sad_dz``
+    **Default** ``sad_gto``
```
robertodr, line 681: Is this change intended here?

doc/users/user_ref.rst
```diff
@@ -731,14 +767,14 @@ User input reference
     **Predicates**
       - ``1.0e-10 < value < 1.0``
   
-   :guess_type: Type of initial guess for response. ``none`` will start from a zero guess for the response functions. ``chk`` restarts a previous calculation which was dumped using the ``write_checkpoint`` keyword. ``mw`` will start from final orbitals in a previous calculation written using the ``write_orbitals`` keyword. The orbitals will be re-projected into the new computational setup. 
+   :guess_type: Type of initial guess for response. ``none`` will start from a zero guess for the response functions. ``chk`` restarts a previous calculation which was dumped using the ``write_checkpoint`` keyword. ``mw`` will start from final orbitals in a previous calculation written ``cube`` will start from cubefile initial guess using the ``write_orbitals`` keyword. The orbitals will be re-projected into the new computational setup.
```
robertodr, line 770:
```suggestion
   :guess_type: Type of initial guess for response. ``none`` will start from a zero guess for the response functions. ``chk`` restarts a previous calculation which was dumped using the ``write_checkpoint`` keyword. ``mw`` will start from final orbitals in a previous calculation written using the ``write_orbitals`` keyword. ``cube`` will start from cubefile initial guess. The orbitals will be re-projected into the new computational setup. 
```

python/template.yml
```diff
@@ -519,12 +519,49 @@ sections:
         docstring: |
           File name for beta orbitals, used with ``cube`` guess.
           Expected path is ``<path_orbitals>/phi_b_scf_idx_<0...Nb>_<re/im>.cube
+      - name: guess_cube_x_p
+        type: str
+        default: initial_guess/x_p
+        docstring: |
+          File name for paired response orbitals, used with ``cube`` guess.
```
robertodr, line 526: You should explain what `<0, 1, 2>` mean (the perturbation components?) and `<0...Np>` mean (the number of occupied orbitals?)

python/template.yml
```diff
@@ -703,13 +740,14 @@ sections:
...
-          - value.lower() in ['none', 'chk', 'mw']
+          - value.lower() in ['none', 'chk', 'mw', 'cube']
         docstring: |
           Type of initial guess for response.
           ``none`` will start from a zero guess for the response functions.
           ``chk`` restarts a previous calculation which was dumped using the
           ``write_checkpoint`` keyword.
           ``mw`` will start from final orbitals in a previous calculation written
+          ``cube`` will start from cubefile initial guess
           using the ``write_orbitals`` keyword. The orbitals will be re-projected
```
robertodr, line 751:
```suggestion
          ``mw`` will start from final orbitals in a previous calculation written
          using the ``write_orbitals`` keyword. ``cube`` will start from cubefile initial guess. The orbitals will be re-projected
```

doc/users/user_ref.rst
Gabrielgerez: No, it seems it got automatically changed by something

robertodr: Are you sure you started this branch from the most up-to-date version of `master`?

Gabrielgerez: well, I was rebasing it against the two previous PRs while working on them. I assumed they were all on the same version

robertodr: The rebasing should be against `origin/master`

Gabrielgerez: I did do that before I started pushing to upstream, I was working on this one parallel while making the tests, i guess I must have fumbled a conflict. I'm fixing the history now.

Gabrielgerez: ok, its was not the rebasing, I don't know why, but running  `update_input_parser.py` adds this line like this.

src/driver.cpp
```diff
@@ -758,18 +758,18 @@ json driver::rsp::run(const json &json_rsp, Molecule &mol) {
         const auto &json_comp = json_rsp["components"][d];
         F_1.perturbation() = h_1[d];
 
-        ///////////////////////////////////////////////////////////
```
robertodr, line 761: @stigrj can you check these changes? We're trying to test the calculation of just the zz component of the polarizability. We only keep in the repo the starting guess cube files for this component of the response vector (one for X, one for Y) because they're pretty hefty. The parser generates a program JSON which contains also the location of the X/Y guesses for the other components, but these files do not exist and the C++ JSON reader, correctly, crashes. To move forward with testing this feature, I have rearragned the set up of the initial guess and the calculation of the actual response tensor within the `if (json_comp.contains("rsp_solver"))` but I don't think it's the correct thing to do, because now only the diagonal components of the tensor will be computed.

Is it OK to make the cube parser dump empty JSON files (just containing `{}`) for the components we don't want to compute?

python/template.yml
Gabrielgerez: done :)

src/driver.cpp
Gabrielgerez: If @stigrj says its ok, then the fix for this is as easy as just removing the last `if` condition in `write_cube_vectors`

stigrj: Sorry for my lateness.

I think the correct behavior would be that a CUBE guess is treated as a zero guess if no file is found for a particular response component. This should be the case regardless of whether the respective components are optimized (`run = [true, true, true]`) or not.

One way to achieve this would be to have the input parser check the given paths up front, e.g.
```
"file_x_a": "initial_guess/X_a_rsp_1",
"file_x_b": "initial_guess/X_b_rsp_1",
"file_x_p": "initial_guess/X_p_rsp_1",
"file_y_a": "initial_guess/Y_a_rsp_1",
"file_y_b": "initial_guess/Y_b_rsp_1",
"file_y_p": "initial_guess/Y_p_rsp_1",
```
If there is no match for `*rsp_1*` in this folder, then the y-component response guess should be changed to `type = none` in the program input without the cube parser being invoked at all, possibly with a printed warning like: `No cube guess found in $path, falling back to zero initial guess`.

Actually, the same goes for a MW guess. It should be possible to provide guess files only for one or two components and have the remaining treated as zero.

stigrj: Btw, is it necessary to run both `frequency = [0.0, 0.001]`? Running only the dynamic case shaves off 1/3 of the run time.

robertodr: That sound reasonable. I'll implement it as you suggest. Regarding the test at two frequencies: it's there to make sure that we can converge both static and dynamic response with a cube guess.

robertodr: Please have another look: I think your suggested logic is now implemented.

stigrj (approved):
Ok, looks good. I get some failing tests, though, on `ctest -L cube_parser` and `ctest -L h2_scf_cube`.

PR #437, opened 2022-11-10
Excited states solver

Gabrielgerez:
Implementing the Excited states solver. This is a re-make of #395 with better structure and is rebased on the changes introduced in #435, which needs to be merged before this one.
I have Created an additional class called `ExcitedStatesSolver` to solve for the excited states.
The functionality of this class is very similar to `LinearResponseSolver` so we might consider adding the excited states solver in the `LinearResponseSolver` in the future, or creating an entirely new class to hold different response solvers, or the *future*, `HigherOrderResponseSolver`. 

**TODO:**
- [ ] Have a working TDA single state implementation.
- [ ] Extend to multiple states and consider no coupling between states.
- [ ] Update frequencies through the integral equation, still only TDA.
- [ ] Extend to RPA through forming and solving the generalized eigenvalue problem.
- [ ] Clean-up and refactoring

First step is almost ready as I create this draft PR.


Gabrielgerez:
the latest commit introduces a bug, but the one previous to that will compute single excited states given a proper initial guess

PR #439, opened 2022-11-20
Nuclear potential

Christian48596:
Introduced 3 models for finite charge nucleus and modified the actual finite charge nucleus model to consider many atom system.

It is early to be merged to master. I am working on reading of parameters by json file.


stigrj:
Made some small fixes, and now it's able to run the three different models and produces reasonable results. The model can be chosen by the input parameter
```
WaveFunction {
  nuclear_model = 3     # 1=HFYGB, 2=Sphere, 3=Gaussian
}
```

stigrj:
Superseeded by #453

PR #441, opened 2022-12-01
Add documetation on use with QCEngine

robertodr:
At long last...

### DIFF
--- a/doc/users/manual.rst
+++ b/doc/users/manual.rst
@@ -40,4 +40,5 @@ in more detail in the sections below.
    running
    user_inp
    user_ref
+   qcengine
    program_json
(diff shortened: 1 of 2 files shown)


stigrj (approved):
Excellent! Is it possible to launch the full parallel version through the engine? Assuming you have it installed in your PATH it would mean just adding a string option to the launch command:
```
mrchem --launcher="mpirun <mpirun-options>" molecule.inp
```
Is it possible to customize this command from within the engine?

stigrj:
Please `squash`-merge if you beat me to the button. Going to :sleeping_bed:

robertodr:
> Excellent! Is it possible to launch the full parallel version through the engine? Assuming you have it installed in your PATH it would mean just adding a string option to the launch command:
> 
> ```
> mrchem --launcher="mpirun <mpirun-options>" molecule.inp
> ```
> 
> Is it possible to customize this command from within the engine?

You can actually set up the MPI call in the Python script. I was going to add that to the docs 😅 I'll do it in the next half an hour, just need to do  a quick test.

PR #443, opened 2022-12-03
Response initial guess by random dipole perturbation

robertodr:
The response orbitals are guessed applying a linear combination, with random coefficients, of monomials in x, y, z.
Essentially, we're trying to introduce some dipole components into the ground state.
Seems to perform marginally better than the zero guess for static/dynamic polarizabilities. It should be more useful for excitation energies, where the zero guess would not work.


stigrj:
Seems to be missing a CMakeLists in the new test folder

robertodr:
Ooops! Now the test is actually in there. I'm not quite sure why for non-zero frequency it's giving me a full polarizability tensor though. That's why the test fails...

ilfreddy:
DITTO

PR #451, opened 2023-04-19
Feature/relax

moritzgubler:
This is a draft of a geometry optimization implementation using the [SQNM](https://aip.scitation.org/doi/10.1063/1.4905665) method. [Variable cell shape optimization](https://arxiv.org/abs/2206.07339) would also possible.

At the moment the geometry optimization input parameters are not parsed. Can someone help me with that?

I changed the main executable to always do a geometry optimization for testing purposes. To test the program, just compile it and run mrchem. A few parameters need to be set manualy:
```
world_prec = 1.0e-6
Properties {
  geometric_derivative = true          # Compute geometric derivative
}
SCF {
  path_orbitals = initial_guess              # Path to orbital files
  write_orbitals = true                # Save converged orbitals to file
  orbital_thrs = 1.0e-6
}
```
A tighter orbital_thrs reduces the noise in the forces significantly. The "path_orbitals" and "write_orbitals" need to be set like that because after the first geometry optimization iteration, I recycle the previous wave function as the initial guess for the next scf cycle to speed up convergence.

TODOLIST:
- [x] Parsing of input parameters
- [ ] Final summary in the mrchem output file


stigrj:
Thanks a lot @moritzgubler, this looks really interesting! I will have a closer look later this week.

moritzgubler:
That sounds great @stigrj.

stigrj:
@moritzgubler I made a pull request on your fork with an updated input parser.

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,139 @@
...
+ * @param forces: forces will be stored here. Dimension of matrix is (3, num_atoms).
+ * 
+ * @return: json summary of scf results.
+*/
+json energyAndForces(json mol_inp, json scf_inp, double &energy, Eigen::MatrixXd &forces) {
+    Molecule mol;
+    driver::init_molecule(mol_inp, mol);
+    auto scf_out = driver::scf::run(scf_inp, mol);
+    // TODO: Ask someone if this mpi barrier is needed!
+    mrcpp::mpi::barrier(mrcpp::mpi::comm_wrk);
```
moritzgubler, line 61: @stigrj Is it save to iterate over the driver that way? I think the mpi barrier also belongs here but I am not sure. What do you think?

stigrj:
I will try to figure out a smart way to handle the output asap

moritzgubler:
Do you have time to look at it soon?

ilfreddy:
Vacation time right now in Norway, but I guess Stig will be able to have a look at it once he is back, or at least delegate it to some of us! 😃

src/vc_sqnm/README.md
```diff
@@ -0,0 +1 @@
+Taken from: https://github.com/moritzgubler/vc-sqnm
```
robertodr, line 1: Could you also specify what the license for the `vq-sqnm` project is? Ideally, it should be included in the `vc_sqnm` folder.

src/vc_sqnm/historylist.hpp
```diff
@@ -0,0 +1,66 @@
...
+    {
+      nhistx = nhistx_;
+      ndim = ndim_;
+      hlist.resize(ndim, nhistx);
+      difflist.resize(ndim, nhistx);
+      normalized_difflist.resize(ndim, nhistx);
+      i_count = 0;
+    }
+
+    int add(Eigen::VectorXd &x_in)
```
robertodr, line 38: I don't see that you're modifying `x_in`, so you should pass by `const` reference.
```suggestion
    int add(const Eigen::VectorXd &x_in)
```

src/vc_sqnm/historylist.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    HistoryList(int &ndim_, int &nhistx_)
+    {
+      nhistx = nhistx_;
+      ndim = ndim_;
+      hlist.resize(ndim, nhistx);
+      difflist.resize(ndim, nhistx);
+      normalized_difflist.resize(ndim, nhistx);
+      i_count = 0;
+    }
```
robertodr, line 36: I don't see a reason to pass these integers by reference, you're not modifying them internally. Also you should use [member initializer lists](https://en.cppreference.com/w/cpp/language/constructor). `i_count` is default-initialized in the list of `private` data members.
```suggestion
    HistoryList(int ndim_, int nhistx_) : nhistx(nhistx_), ndim(ndim_)
    {
      hlist.resize(ndim, nhistx);
      difflist.resize(ndim, nhistx);
      normalized_difflist.resize(ndim, nhistx);
    }
```

src/vc_sqnm/historylist.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+    int add(Eigen::VectorXd &x_in)
+    {
+      if (i_count < nhistx){
+        hlist.col(i_count) = x_in;
+        i_count += 1;
+        for (int i = 1; i < i_count; i++){
+          difflist.col(i-1) = hlist.col(i) - hlist.col(i - 1);
+          normalized_difflist.col(i-1) = difflist.col(i-1) / difflist.col(i-1).norm();
```
robertodr, line 45: I think you can use Eigen's block operations and slicing syntax here: https://eigen.tuxfamily.org/dox/group__TutorialSlicingIndexing.html It should be cleaner to read and also perform a bit better.

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,136 @@
...
+#include "chemistry/Molecule.h"
+// #include "chemistry/PhysicalConstants.h"
+#include "vc_sqnm/periodic_optimizer.hpp"
+
+#include <Eigen/Dense>
+
+using json = nlohmann::json;
+using namespace mrchem;
+
+Eigen::MatrixXd getPositions(json mol_inp) {
```
robertodr, line 21: pass by const-reference because you're not modifying `mol_inp` and you avoid to take a, possibly expensive, copy
```suggestion
Eigen::MatrixXd getPositions(const json &mol_inp) {
```

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,136 @@
...
+        xyz[1] = coord["xyz"][1];
+        xyz[2] = coord["xyz"][2];
+
+        pos.col(i) = Eigen::Vector3d(xyz);
+        i++;
+    }
+    return pos;
+}
+
+void setPositions(json &mol_inp, Eigen::MatrixXd &pos) {
```
robertodr, line 39: pass by const-reference when not modifying the argument.
```suggestion
void setPositions(json &mol_inp, const Eigen::MatrixXd &pos) {
```

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,136 @@
...
+    Eigen::MatrixXd pos(3, mol.getNNuclei());
+
+    int i = 0;
+    for (const auto &coord : mol_inp["coords"]) {
+        double xyz[3];
+        xyz[0] = coord["xyz"][0];
+        xyz[1] = coord["xyz"][1];
+        xyz[2] = coord["xyz"][2];
+
+        pos.col(i) = Eigen::Vector3d(xyz);
```
robertodr, line 33: I think this does it:
```suggestion
        pos.col(i) << coord["xyz"][0], coord["xyz"][1], coord["xyz"][2];
```

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,136 @@
...
+ * @brief Calculates energy and forces of a molecule.
+ * 
+ * @param mol_inp: json that describes the molecule.
+ * @param scf_inp: scf settings.
+ * @param energy: energy will be stored in this variable.
+ * @param forces: forces will be stored here. Dimension of matrix is (3, num_atoms).
+ * 
+ * @return: json summary of scf results.
+*/
+json energyAndForces(json mol_inp, json scf_inp, double &energy, Eigen::MatrixXd &forces) {
```
robertodr, line 56: If you have to return multiple things from a function, either use a data structure _or_ return a tuple. Pass-by-reference (or pass-by-pointer) to mimic multiple return values is confusing. Also, I believe that `mol_inp` and `scf_inp` can be passed by const-reference.

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+    PES_optimizer::periodic_optimizer optimizer(num_atoms, init_step_size, max_history_length, minimal_step_size, subspace_tolerance);
+    
+    int i = 0;
+    Eigen::MatrixXd forces(3, num_atoms);
+    double energy;
+    
+    json results = energyAndForces(mol_inp, scf_inp, energy, forces);
+    // write progress to stderr for earier debugging.
+    std::cerr << i << " " << energy << " " << forces.cwiseAbs().maxCoeff() << "\n";
```
robertodr, line 105: You should use the various printing facilities in MRChem to do this.

src/vc_sqnm/periodic_optimizer.hpp
robertodr, line 1: is this file used currently? If not I would say it should not be part of this PR.

src/vc_sqnm/README.md
moritzgubler: the code is licensed under gplv3. In [this commit](https://github.com/MRChemSoft/mrchem/pull/451/commits/3e81c1a4fd43ec5a00d04fd9e7b31e57edd49aa4) I added this information to the README and added a copy of the gplv3 license in the vc-sqnm directory.

src/vc_sqnm/historylist.hpp
moritzgubler: Ok, I did that [here](https://github.com/MRChemSoft/mrchem/pull/451/commits/c822f09c2bd99c8a81b532dfed58820fa97beec8)

src/vc_sqnm/historylist.hpp
moritzgubler: Ok, I did that [here](https://github.com/MRChemSoft/mrchem/pull/451/commits/c822f09c2bd99c8a81b532dfed58820fa97beec8)

src/vc_sqnm/historylist.hpp
moritzgubler: Ok, I did that [here](https://github.com/MRChemSoft/mrchem/pull/451/commits/c822f09c2bd99c8a81b532dfed58820fa97beec8)

src/vc_sqnm/periodic_optimizer.hpp
moritzgubler: No, it is not. I have seen that periodic boundary conditions are a planned feature. Because of that, I included it. It would only require a few changes in the mrchem_optimizer.hpp file for it to be able to do variable cell shape relaxations.

If this feature is not needed in the near future, we can delete the file.

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: I don't really now what to print/ how to format the output. Do you have some printing guidelines?

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: I improved this mechanism [here](https://github.com/MRChemSoft/mrchem/pull/451/commits/8389499e273071a821e75e4359562c38d5e4217d)

src/vc_sqnm/README.md
```diff
@@ -0,0 +1,3 @@
...
+
+The vc-sqnm code is licensed under the GNU General Public License v3.0
```
robertodr, line 3: According to `LICENSE` it's LGPL:
```suggestion
The vc-sqnm code is licensed under the GNU Lesser General Public License v3.0
```

src/vc_sqnm/sqnm.hpp
```diff
@@ -0,0 +1,270 @@
...
+        }
+
+        if (this->estimate_step_size)
+        {
+          double prev_df_squared = std::pow(prev_df_dx.norm(), 2);
+          double l1 = (f_of_x - prev_f + alpha * prev_df_squared) / (0.5 * alpha * alpha * prev_df_squared);
+          double l2 = (df_dx - prev_df_dx).norm() / (alpha * prev_df_dx.norm());
+          alpha = 1.0 / std::max(l1, l2);
+          std::cout << "Automatic initial step size guess for geometry optimization: " << alpha << '\n';
+          std::cout << "Should be between 0.4 and 8. If there are problems in the geometry optimization, consider setting the initial step size manually."
```
moritzgubler, line 142: This may also be useful for printing

src/vc_sqnm/README.md
moritzgubler: Sorry, I meant to upload the General Public License v3.0 not the Lesser General Public License. I have chosen theGeneral Public License v3.0 for the vc-sqnm repository. Is that compatible with mrchem? If not we can change my license to the Lesser GPL3.

moritzgubler:
I implemented all your suggestions and tested the code.

I have a question about the units. The parameters that will be set in my code are in Hartree/ Bohr and my code expects Hartree/ Bohr. 

Some input parameters would change if the units were angstrom. For example the initial step size. Do you think it makes sense to change them?

src/vc_sqnm/README.md
robertodr: MRChem is LGPLv3 or later. Since we're copying the source code, the combination has to be re-licensed under GPLv3 or later, which means MRChem needs to relicense, as seen from the license compatibility matrix that the FSF publishes. There might be, but I doubt it, some nuance to this. See here: https://www.gnu.org/licenses/gpl-faq.en.html#AllCompatibility 

Long story short, we'd appreciate it if you relicense under LGPLv3 or later.

moritzgubler: Ok, lets use LGPLv3 then.

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: These numbers are quite important to see if the geometry optimization converges. If they would just be printed, it is easy to overlook them because of all the other prints. Do you have an idea on how we could highlight this?

robertodr:
> I implemented all your suggestions except for the printing and tested the code.

Thank you very much Moritz, especially for your patience! I can have a second review round in the weekend/Monday at the latest.

> I have a question about the units. The parameters that will be set in my code are in Hartree/ Bohr and my code expects Hartree/ Bohr.
> 
> Some input parameters would change if the units were angstrom. For example the initial step size. Do you think it makes sense to change them?

I'm pretty sure that, at the level where you've interfaced your code, everything is guaranteed to be in atomic units already.

robertodr:
I see that your branch is outdated with `master` and has accumulated some conflicts. Can you bring the branch up to date and fix the conflicts?

src/vc_sqnm/mrchem_optimizer.hpp
robertodr: There is a `Printer` class that comes from MRCPP, which tries to handle formatting and such.[^1] You can `#include <MRCPP/Printer>` and see here for some usage documentation https://mrcpp.readthedocs.io/en/latest/mrcpp_api/printer.html 

I would say that this data should be part of both the JSON _and_ the human-readable output files. When it comes to format for the latter, I don't have any great suggestions to offer, unfortunately. Maybe at the end of each optimization step a summary like printed in Psi4?

```
  ==> Convergence Check <==

  Measures of convergence in internal coordinates in au.
  Criteria marked as inactive (o), active & met (*), and active & unmet ( ).
  --------------------------------------------------------------------------------------------- ~
   Step     Total Energy     Delta E     MAX Force     RMS Force      MAX Disp      RMS Disp    ~
  --------------------------------------------------------------------------------------------- ~
    Convergence Criteria    1.00e-06 *    3.00e-04 *             o    1.20e-03 *             o  ~
  --------------------------------------------------------------------------------------------- ~
      1     -74.96466254   -7.50e+01      2.37e-02      1.83e-02 o    1.44e-01      8.71e-02 o  ~
  ---------------------------------------------------------------------------------------------
  ```

[^1]: I'm not the expert on this, usually Stig Rune intervenes to cleansup the output...

File: src/vc_sqnm/mrchem_optimizer.hpp [hunk 13]
```diff
@@ -0,0 +1,162 @@
...
+
+    json summary;
+    summary[i] = {
+        {"results", results},
+        {"molecule", mol_inp}
+    };
+
+    Eigen::MatrixXd pos = getPositions(mol_inp);
+
+    while (i < max_iter && forces.cwiseAbs().maxCoeff() > max_force_component) {
```
robertodr, line 139: I see no you're evaluating convergence only on the maximum of the gradient. Would it be possible to allow different convergence criteria (even a union of them) like done in Gaussian and Psi4?

moritzgubler: Yes, this is of course possible. What criteria do you have in mind?

Personally, I think the maximum absolute value of the gradient is the best convergence parameter. It is not dependant on the system size (meaning that if you double the system, one does not have to change the threshold for an equally tight geometry optimization) and it actually guarantees that the geometry optimization has converged. 

A lot of other codes look at energy differences between iterations. These differences depend on the system size and they don't guarantee that convergence has been achieved. They just show that the algorithm struggles to go on.

src/vc_sqnm/periodic_optimizer.hpp
moritzgubler: @robertodr  Sorry, I was wrong. This file is used at the moment. It is the main geometry optimization file. The constructor is overloaded and if no unit cell is provided in the periodic_optimizer object is created, it will optimize molecules with free boundary conditions.

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: Hi @robertodr 

Now, all the relevant prints are replaced by the mrchem printing api.

moritzgubler: and I format an output json that gets written after the simulation. (see changes in mrchem.cpp)

moritzgubler:
Hi @robertodr 
I just brought the branch up to date with master again. Do you have time for another round of review in the next days?

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,188 @@
...
+ * @param mol_inp: json that describes the molecule.
+ * @param scf_inp: scf settings.
+ * 
+ * @return: json summary of scf results.
+*/
+json getSCFResults(const json mol_inp, const json scf_inp) {
+    Molecule mol;
+    driver::init_molecule(mol_inp, mol);
+    auto scf_out = driver::scf::run(scf_inp, mol);
+    // TODO: Ask someone if this mpi barrier is needed!
```
robertodr, line 65:
```suggestion
    // keeping the mpi barrier to be on the safe side, but not sure if it is needed
```

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,188 @@
...
+ * @param scf_results json containing the results of an scf calculations. Must be obtained using the function getSCFResults.
+ * @param forces: Provide empty matrix of dimension (3, num_atoms).
+ * 
+ * @return Forces acting on nuclei
+*/
+Eigen::MatrixXd extractForcesInPlace(const json scf_results, Eigen::MatrixXd &forces){
+    for (int i = 0; i < forces.cols(); i++)
+    {
+        for (int j = 0; j < 3; j++) {
+            // is this safe? will the forces always be stored in ["geometric_derivative"]["geom-1"]?
```
robertodr, line 81:
```suggestion
```
yes, it is safe to assume `|geom-1"` is always used in a geometry optimization. We could also save it under a different keys, just to be extra sure.

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,188 @@
...
+}
+
+/**
+ * @brief Extracts forces from json containing scf results.
+ * @param scf_results json containing the results of an scf calculations. Must be obtained using the function getSCFResults.
+ * @param forces: Provide empty matrix of dimension (3, num_atoms).
+ * 
+ * @return Forces acting on nuclei
+*/
+Eigen::MatrixXd extractForcesInPlace(const json scf_results, Eigen::MatrixXd &forces){
```
robertodr, line 77:
```suggestion
Eigen::MatrixXd extractForcesInPlace(const json &scf_results, Eigen::MatrixXd &forces){
```

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,188 @@
...
+    return -forces;
+}
+
+/**
+ * @brief Extracts energy from json containing scf results.
+ * @param scf_results json containing the results of an scf calculations. Must be obtained using the function getSCFResults.
+ * 
+ * @return Total energy
+*/
+double extractEnergy(const json scf_results){
```
robertodr, line 95:
```suggestion
double extractEnergy(const json &scf_results){
```

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+/**
+ * @brief Optimizes positions of nuclei.
+ * 
+ * @param scf_inp: scf settings.
+ * @param mol_inp: json that contains the molecule.
+ * @param geopt_inp: json that contains the geometry optization settings.
+ * @return: A summary of the geometry optimization trajectory.
+*/
+json optimize_positions(json scf_inp, json mol_inp, json geopt_inp) {
```
robertodr, line 107:
```suggestion
json optimize_positions(const json &scf_inp, const json &mol_inp, const json &geopt_inp) {
```

src/vc_sqnm/mrchem_optimizer.hpp
robertodr: I'm confused why `forces` is passed from outside, but also returned? Does the function have to have this signature?

src/vc_sqnm/mrchem_optimizer.hpp
robertodr: Thanks for the explanation. We might revisit in the future to combine multiple criteria, but for now it's OK to go with the default you chose.

moritzgubler:
Do you think of unit tests or integration tests? An integration test would be quite easy to set up but it would probably need a couple of minutes of runtime. A good test system there would be H_2O. Molecules with only two atoms are less optimal for testing because of all the symmetries (its just a one dimensional optimization problem).

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: I think this is fine then.

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: The signature was indeed weird. I have fixed this now.

src/vc_sqnm/mrchem_optimizer.hpp
moritzgubler: For the first two arguments, I need copies of the input variables

moritzgubler:
This converges in 20 minutes and 4 geometry optimization iterations on my machine. The tight world_prec is needed for accurate forces. To check if the geometry optimization worked, we could check the energy, the two bond lengths and the angle of the molecule. Would something like this be useful for testing?

```world_prec = 1.0e-6
world_unit = angstrom

WaveFunction {
  method = lda
}

Properties {
  geometric_derivative = true          # Compute geometric derivative
}

GeometryOptimizer{
        run = true
        use_previous_guess = True
        max_force_component = 1e-3
}

Molecule {
$coords
O          0.00000        0.00000        0.11779
H          0.00000        0.75545       -0.47116
H          0.00000       -0.75545       -0.47116
$end
}
```

robertodr:
That is an expensive test to run 😞 In some of the integration tests, we don't run calculations to full precision, exactly to make sure that they run in a reasonable amount of time: for example, we only do one SCF cycle. Would it be possible to only do one optimization cycle and check the result of that? Would it be a reliable test?

Also, can you add a short page in the documentation showing how to run a geometry optimization? A new page is a restructuredtext (rst) plain-text document here: https://github.com/MRChemSoft/mrchem/tree/master/doc/users that you then reference here: https://github.com/MRChemSoft/mrchem/blob/master/doc/users/manual.rst

robertodr:
Also, do you need to have:
```
Properties {
  geometric_derivative = true          # Compute geometric derivative
}
```
in the input? From reading the parser it seems you set it to true whenever `GeometryOptimizer` is active?

moritzgubler:
This test is not as good, but it should also be ok:
```
world_prec = 1.0e-4
world_unit = angstrom

WaveFunction {
  method = lda
}

GeometryOptimizer{
        run = true
        use_previous_guess = True
        max_force_component = 1e-3
}

Molecule {
$coords
H 0.0 0.0 0.0
H 0.9 0.0 0.0
$end
}
```
This only takes 20 seconds. Is that still too long?
When only doing 1 scf cycle, the forces will be really bad. I don't think, it really makes sense to test this. Only testing the first geometry optimization step is also not ideal, since the first iteration is much simpler than the second one. Only in the second step, I start estimating curvatures.

We can also think of using a force field just for testing the geometry optimizer.

moritzgubler:
I have written the rst doc file, do you think it is OK like that?

robertodr:
> This test is not as good, but it should also be ok:
> 
> ```
> world_prec = 1.0e-4
> world_unit = angstrom
> 
> WaveFunction {
>   method = lda
> }
> 
> GeometryOptimizer{
>         run = true
>         use_previous_guess = True
>         max_force_component = 1e-3
> }
> 
> Molecule {
> $coords
> H 0.0 0.0 0.0
> H 0.9 0.0 0.0
> $end
> }
> ```
> 
> This only takes 20 seconds. Is that still too long? When only doing 1 scf cycle, the forces will be really bad. I don't think, it really makes sense to test this. Only testing the first geometry optimization step is also not ideal, since the first iteration is much simpler than the second one. Only in the second step, I start estimating curvatures.
> 
> We can also think of using a force field just for testing the geometry optimizer.

this would be acceptable :) Can you try with beryllium hydride (BeH2) _not_ in a linear geometry? It's 6 electrons, so more expensive, but possibly still reasonable? The documentation looks good, you can check it here: https://mrchem--451.org.readthedocs.build/en/451/users/geometry_optimization.html (the link to the generated documentation for each PR appears in the list of checks) I'll edit it a bit because the TOC is picking up all section headers.

moritzgubler:
This converges in 3 iterations and one minute. Do you think this is fine?
```
world_prec = 1.0e-4

WaveFunction {
  method = lda
}

GeometryOptimizer{
        run = true
        max_force_component = 1e-2
}

Molecule {
$coords
 Be       0.047274        0.032162       -0.275106
  H      -0.000377        2.549743       -0.704835
  H      -0.000807       -2.442007       -0.685253
$end
}
```

robertodr:
Very fine! 🤩 The longest test we have, `H2_polarizability_CUBE`, takes ~120 seconds on GitHub, so something that runs in 1 minutes is quite good!

moritzgubler:
Will you implement the test or should I do it?

robertodr:
I can do it. Ping me Monday if it hasn't happened though :)

robertodr:
I have fleshed out the test locally, but I can't seem to push to your repo...

Apart from this, I would like @stigrj to chime in when it comes to the output (both JSON and not) At least for the JSON, the format is significantly different with or without geometry optimization and we need to figure out a way to not break backwards compatibility.

moritzgubler:
@stigrj @robertodr One possibility is that the json from the last SCF iteration is returned which will have the traditional scf format and write the json I return at the moment into a file to make sure that the information of the optimization trajectory is not lost. What do you think of that?

stigrj:
@moritzgubler sorry for my ridiculous response time on this. I actually thought there would be much more for me to consider here, but it turns out you have sorted out most things already :slightly_smiling_face:  Great work!

I think your last suggestion with a separate JSON file for the geometry iterations is quite acceptable, so if you can just make that final change we can get this PR merged. Then the regular JSON output file should contain the **original** input but the output of the **final** geometry iteration. In this way the file should be backwards compatible.

I have resolved the conflicts in the input parser with a rebase and added some minor changes to the printed output. I pushed the fixes to the `feature/relax` branch on my fork : https://github.com/stigrj/mrchem/tree/feature/relax, so you can continue from there if you want.

moritzgubler:
@stigrj No worries:)

I may have an even better idea: I adjust the mrchem output json the way you mentioned to make sure it is backwards compatible. The geometry optimization trajectory is written as a .xyz file. That way it is is easy to visualize the converged geometry. If you also agree with this, I will implement it like that.

stigrj:
Is there any standardized format for such an xyz-file that other programs use? Will it contain only the molecular geometry? I think it will be good to keep a full record of all the properties (energy contributions, gradients, etc), as is currently done in the non-backward-compatible json we have now, only dump it into a separate file. But if there is a standard xyz-format that allows for immediate visualization of the trajectory, you can of course add that as well :slightly_smiling_face:

moritzgubler:
The [.xyz](https://en.wikipedia.org/wiki/XYZ_file_format) format is pretty standardized and supported by standard visualization software. In that case, I will do both the json summary and the xyz output file.

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -130,8 +134,13 @@ json optimize_positions(json scf_inp, json mol_inp, const json &geopt_inp) {
     Eigen::MatrixXd forces(3, num_atoms);
     double energy;
     
-    json results = getSCFResults(mol_inp, scf_inp);
+    std::tuple<json, json> results_tuple = getSCFResults(mol_inp, scf_inp);
+    std::cout<< "unpacking\n";
```
stigrj, line 174: There are a few debug prints lingering that should be removed

moritzgubler:
Hi @stigrj 
I polished the xyz output, removed the debug prints, renamed the xyz file and the json summary file such that they share the base name of the original output.

src/vc_sqnm/mrchem_optimizer.hpp
```diff
@@ -0,0 +1,267 @@
...
+    }
+
+    int numAtoms = atomicPositions.rows();
+
+    xyzFile << numAtoms << "\n";
+    xyzFile << comment << std::endl;
+
+    for (int i = 0; i < numAtoms; ++i) {
+        xyzFile << atomicLabels[i] << " " << atomicPositions(0, i) * 0.529177 << " " << atomicPositions(1, i) * 0.529177
+                << " " << atomicPositions(2, i) * 0.529177 << "\n";
```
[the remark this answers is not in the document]
stigrj: Do these need to be in Angstrom for the xyz-file? Should perhaps be given in the same unit as the initial coordinates in the input file?

Anyways, we have a conversion factor in the `PhysicalConstants` class:

```
#include "chemistry/PhysicalConstants.h"

double a2b = PhysicalConstants::get("angstrom2bohrs");
```

This conversion factor can be tuned by the user through the input file.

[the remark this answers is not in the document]
moritzgubler: The standard unit in xyz files is angstrom. I used mrchems utility to get the conversion factor.

moritzgubler:
@stigrj I think I adapted all the changes we recently talked about. I want to test one final thing regarding the xyz file output. I will contact you when I think the branch is ready to be merged.

moritzgubler:
@stigrj The xyz writer is now also tested well and works. I have done and tested everything I wanted. If you also agree, we can merge it now.

stigrj:
Great! Going in once the CI checks pass

PR #453, opened 2023-05-21
Finite nucleus, take 2

stigrj:
Supersedes #439

Rebase on master and cleanup in nuclear models:
- Combined the new finite nucleus models with Peter's collection of point-nucleus models
- No more `param_V.json` file for nuclear parameters. Visschers's RMS values are now written directly in the python periodic table, and passed to `mrchem.x` through the JSON input. RMS values can thus only be changed by interrupting the input parser and manually alter the intermediate file. 
- Input keyword changed from `int` to `string`

Should be properly tested against Exciting before merging.

- [x] Ready to go


src/qmoperators/one_electron/NuclearOperator.cpp
```diff
@@ -45,23 +57,41 @@ namespace mrchem {
...
     // Setup local analytic function
     Timer t_loc;
-    NuclearFunction f_loc;
-    setupLocalPotential(f_loc, nucs, smooth_prec, false);
-    Nuc_func = NuclearFunction(nucs, smooth_prec);
+    NuclearFunction *f_loc = nullptr;
+    if (model == "point_like") {
+        mrcpp::print::header(1, "Projecting nuclear potential (point-like HFYGB)");
+        f_loc = new PointNucleusHFYGB();
+        setupLocalPotential(*f_loc, nucs, smooth_prec);
```
robertodr, line 71: I guess `setupLocalPotential(*f_loc, nucs, smooth_prec);` can go right after the long `if`-`else` chain? Less code duplication.

robertodr (approved):
LGTM. 

Is the ambiguity between `point_parabola` and `finite_gaussian` been resolved? I recall some messages on Zulip implying that they are the same?

stigrj:
It was actually the `finite_sphere` that was claimed to be the same as `point_parabola`, but they are in fact quite different. All the `point_*` models are different numerical smoothing strategies operating within the numerical precision, while the `finite_*` models are for physically extended nuclei. So the mathematical form of the potentials are parabolic in both cases, but the extent of the smoothed region is several orders of magnitude different.

Christian48596 (commented):
ok

PR #454, opened 2023-05-23
Fix mpi error on SAD_GTO guess, closes #452

### LINKED ISSUE #452: SAD_GTO guess with MPI
The `SAD_GTO` initial guess seems to bug out when running on many MPI processes

stigrj:
@gitpeterwind , not exactly sure why the rank should be set incrementally by the number of orbitals rather than the number of MPI ranks, but this is the way it's done in the `core::project_ao` equivalent, and it seems to work :shrug:


gitpeterwind (approved):
Yes, the rank of an orbital is now its position in the orbitalvector. The long term idea is that all "collective" or parallel information should be handled through orbitalvector operations, while orbitals operations are serial.

PR #458, opened 2023-07-24
cavity function used directly instead of projected first

gitpeterwind:
Avoids memory consuming function trees, by using directly the analytical function.
Note: requires the latest version of MRCPP
Gabriel will test/polish before merge into master


robertodr:
Thanks Peter!

gitpeterwind:
How was it that we force it to use the latest mrcpp?

src/driver.cpp
```diff
@@ -1056,12 +1057,14 @@ void driver::build_fock_operator(const json &json_fock, Molecule &mol, FockBuild
         auto eps_i = json_fock["reaction_operator"]["epsilon_in"];
         auto eps_o = json_fock["reaction_operator"]["epsilon_out"];
         auto formulation = json_fock["reaction_operator"]["formulation"];
-        auto accelerate_pot = (optimizer == "potential") ? true : false;
+        auto accelerate_pot = (optimizer == "potential") ? true : false; // this has to go
```
robertodr, line 1060: no need for the ternary `if`, but I guess that this whole statement is useless, given the comment? If so, just remove it.
```suggestion
        auto accelerate_pot = (optimizer == "potential"); // this has to go
```

src/environment/SCRF.cpp
```diff
@@ -48,37 +46,30 @@ using OrbitalVector_p = std::shared_ptr<mrchem::OrbitalVector>;
 
 namespace mrchem {
 
-SCRF::SCRF(Permittivity e, const Nuclei &N, PoissonOperator_p P, DerivativeOperator_p D, double orb_prec, int kain_hist, int max_iter, bool acc_pot, bool dyn_thrs, std::string density_type)
-        : accelerate_Vr(acc_pot)
-        , dynamic_thrs(dyn_thrs)
+SCRF::SCRF(Permittivity &e, const Density &rho_nuc, PoissonOperator_p P, DerivativeOperator_p D, int kain_hist, int max_iter, bool dyn_thrs, std::string density_type)
```
robertodr, line 49: The `Permittivity` should be passed by `const`-reference. As far as I can see, the constructor just copies it to `epsilon`?

src/environment/SCRF.cpp
```diff
@@ -48,37 +46,24 @@ using OrbitalVector_p = std::shared_ptr<mrchem::OrbitalVector>;
...
         , history(kain_hist)
-        , apply_prec(orb_prec)
-        , conv_thrs(1.0)
-        , mo_residual(1.0)
         , epsilon(e)
-        , rho_nuc(false)
-        , rho_ext(false)
-        , rho_tot(false)
+        , rho_nuc(rho_nuc)
         , Vr_n(false)
```
robertodr, line 56: I see this was here from before, but it doesn't make sense to me. `Vr_n` is of type `mrcpp::ComplexFunction` how can initializing it with `false` be correct?

robertodr: @Gabrielgerez

src/environment/SCRF.cpp
```diff
@@ -88,59 +73,65 @@ double SCRF::setConvergenceThreshold(double prec) {
...
     mrcpp::ComplexFunction rho_eff;
     mrcpp::ComplexFunction first_term;
-    mrcpp::ComplexFunction Vr;
-    mrcpp::ComplexFunction eps_inv;
-    eps_inv.alloc(NUMBER::Real);
-    Vr.alloc(NUMBER::Real);
-
+    mrcpp::ComplexFunction Vr_np1;
+    Vr_np1.alloc(NUMBER::Real);
+    // TODO want to avoid using the flipFunction method
```
robertodr, line 120: Is this TODO still relevant?

src/environment/SCRF.cpp
```diff
@@ -88,59 +73,65 @@ double SCRF::setConvergenceThreshold(double prec) {
...
     this->epsilon.flipFunction(true);
-    mrcpp::cplxfunc::project(eps_inv, this->epsilon, NUMBER::Real, this->apply_prec / 100);
+
+    Density rho_tot(false);
+    computeDensities(Phi, rho_tot);
+
+    mrcpp::cplxfunc::multiply(first_term, rho_tot, this->epsilon, this->apply_prec);
```
robertodr, line 126: Maybe it's enough to use `1.0 / this->epsilon` to avoid using `flipFunction`?

src/environment/SCRF.cpp
```diff
@@ -161,7 +155,7 @@ void SCRF::accelerateConvergence(mrcpp::ComplexFunction &dfunc, mrcpp::ComplexFu
     dPhi_n.clear();
 }
 
-void SCRF::nestedSCRF(mrcpp::ComplexFunction V_vac) {
+void SCRF::nestedSCRF(mrcpp::ComplexFunction &V_vac, const std::shared_ptr<mrchem::OrbitalVector> &Phi_p) {
```
robertodr, line 158: I think this is the appropriate signature here:
1. You're not going to modify the vacuum potential `V_vac`, and
2. You only want to keep the shared pointer alive as long as you use it, but you're not going to change its contents (see screenshot[^1]), and
3. You're not modifying any of the class data, hence the `const` (but this I might be wrong about)
```suggestion
void SCRF::nestedSCRF(const mrcpp::ComplexFunction &V_vac, std::shared_ptr<mrchem::OrbitalVector> Phi_p) const {
```



[^1]: https://www.modernescpp.com/index.php/c-core-guidelines-passing-smart-pointer/

src/environment/SCRF.cpp
```diff
@@ -172,47 +166,38 @@ void SCRF::nestedSCRF(mrcpp::ComplexFunction V_vac) {
...
-            gamma_np1.free(NUMBER::Real);
-            mrcpp::cplxfunc::add(gamma_np1, 1.0, gamma_n, 1.0, dgamma_n, -1.0);
-        }
-
-        updateCurrentGamma(gamma_np1);
         printConvergenceRow(iter, norm, update, t_iter.elapsed());
         iter++;
     }
     if (iter > max_iter) println(0, "Reaction potential failed to converge after " << iter - 1 << " iterations, residual " << update);
+    // computeEnergies(*Phi_p, this->rho_nuc); // TODO might not really need this, test it.
```
robertodr, line 198: Seems this is not needed after all? I suggest to remove it altogether.

src/environment/SCRF.cpp
```diff
@@ -242,63 +227,41 @@ void SCRF::printConvergenceRow(int i, double norm, double update, double time) c
...
-double SCRF::getElectronicEnergy() {
-    mrcpp::ComplexFunction rho_el;
-    rho_el.alloc(NUMBER::Real);
-    mrcpp::cplxfunc::add(rho_el, 1.0, this->rho_tot, -1.0, this->rho_nuc, -1.0);
-    return mrcpp::cplxfunc::dot(rho_el, this->Vr_n).real();
-}
-
-double SCRF::getTotalEnergy() {
-    return mrcpp::cplxfunc::dot(this->rho_tot, this->Vr_n).real();
+void SCRF::computeEnergies(mrchem::OrbitalVector &Phi) {
```
robertodr, line 257: I think the orbitals can be passed by `const`-reference.
```suggestion
void SCRF::computeEnergies(const mrchem::OrbitalVector &Phi) {
```

src/environment/SCRF.h
```diff
@@ -30,93 +30,78 @@
...
-    mrcpp::ComplexFunction gamma_n;
-    mrcpp::ComplexFunction dgamma_n;
-    mrcpp::ComplexFunction gamma_nm1;
 
-    mrcpp::FunctionTreeVector<3> d_cavity; //!< Vector containing the 3 partial derivatives of the cavity function
     std::shared_ptr<mrcpp::DerivativeOperator<3>> derivative;
     std::shared_ptr<mrcpp::PoissonOperator> poisson;
 
-    void setDCavity();
+    // void setDCavity();
```
robertodr, line 86:
```suggestion
```

src/environment/SCRF.h
```diff
@@ -30,93 +30,78 @@
...
+
+    double const &getNuclearEnergy() { return this->Er_nuc; };
+    double const &getElectronicEnergy() { return this->Er_el; };
+    double const &getTotalEnergy() { return this->Er_tot; };
```
robertodr, line 90: Do we really want to allow modifying these data from outside the class? I think they should be read-only and hence they should return simply `double`.

src/environment/SCRF.h
```diff
@@ -30,93 +30,78 @@
...
-    void computeDensities(OrbitalVector &Phi);
+    void computeDensities(OrbitalVector &Phi, Density &rho_out);
     void computeGamma(mrcpp::ComplexFunction &potential, mrcpp::ComplexFunction &out_gamma);
 
-    mrcpp::ComplexFunction solvePoissonEquation(const mrcpp::ComplexFunction &ingamma);
+    mrcpp::ComplexFunction solvePoissonEquation(const mrcpp::ComplexFunction &ingamma, mrchem::OrbitalVector &Phi);
 
     void accelerateConvergence(mrcpp::ComplexFunction &dfunc, mrcpp::ComplexFunction &func, KAIN &kain);
 
     // TODO    void variationalSCRF(mrcpp::ComplexFunction V_vac);
```
robertodr, line 99:
```suggestion
```

src/environment/SCRF.h
```diff
@@ -30,93 +30,78 @@
...
-    mrcpp::ComplexFunction &getCurrentGamma() { return this->gamma_n; }
-    mrcpp::ComplexFunction &getPreviousGamma() { return this->gamma_nm1; }
-    mrcpp::ComplexFunction &getCurrentDifferenceGamma() { return this->dgamma_n; }
 
     Permittivity &getPermittivity() { return this->epsilon; }
 
     void updateMOResidual(double const err_t) { this->mo_residual = err_t; }
 
     friend class ReactionPotential;
+    void computeEnergies(mrchem::OrbitalVector &Phi);
```
robertodr, line 58: I suggest to change the implementation of `computeEnergies` to make it return the values directly, either just electronic and nuclear or electronic, nuclear, and total. I don't see much need to save the values in the class, as they are quite easily recomputable.
```suggestion
    auto computeEnergies(mrchem::OrbitalVector &Phi) -> std::tuple<double, double>;
```

robertodr: And this would possibly need to become a public method.

src/qmoperators/two_electron/FockBuilder.cpp
```diff
@@ -169,6 +169,7 @@ SCFEnergy FockBuilder::trace(OrbitalVector &Phi, const Nuclei &nucs) {
 
     // Reaction potential part
     if (this->Ro != nullptr) {
+        this->Ro->getHelper()->computeEnergies(Phi);
```
robertodr, line 172: With the above suggestion, here you would simply do:
```suggestion
        auto [Er_nuc, Er_el] = this->Ro->getHelper()->computeEnergies(Phi);
        auto Er_tot = Er_nuc + Er_el;
```
note I'm assuming the 0.5 is _included_ in the return value.

robertodr (commented):
This looks great @gitpeterwind! You made a lot of simplifications that I've been also making in #456 😄 For which system(s) have you been testing the memory footprint reductions? And what is the before and after consumption?

I've left some comments, especially about argument passing. Please have a look.

robertodr:
Note that the solvation tests fail now:

```
         49 - reaction_operator (Failed)
         60 - Li_solvent_effect (Failed)
```
The former is a unit test, the latter an integration test. For the latter it might be enough to regenerate the reference output.

src/driver.cpp
Gabrielgerez: Yes, will get to fix the input file

src/environment/SCRF.cpp
Gabrielgerez: I am not sure of this myself, but removing that line seems to reduce memory usage

src/environment/SCRF.cpp
Gabrielgerez: Well, I still want to avoid it. Though it is not really a big bottleneck anyways

src/qmoperators/two_electron/FockBuilder.cpp
Gabrielgerez: it isn't now, but I can just add it :)

src/environment/SCRF.cpp
gitpeterwind: This is my "fault". I changed the constructor of complex functions, and made a mistake (thought that the type of arguments was determining what was default and what was set, as in fortran, while it is the order which matters in C++).
So now it is not backwards compatible. Previously a single argument would determine if the function was using "MPI shared memory", but now it is the last argument. We will revise the complex functions at some point, and then we can improve this too.
Note that if these functions are using lots of memory (they probably do) it might be useful to use shared memory in the future. But that should use a steering from input file

src/environment/SCRF.cpp
Gabrielgerez: Done

src/environment/SCRF.cpp
Gabrielgerez: Im going to leave it as it is now, running more calculations seem to give different mem. usage, but the end result of the calculation is the same

src/environment/SCRF.cpp
Gabrielgerez: It does not work off the bat. I could do an operator overload of the `/` operator for the permittivity function, but that would just have the same function as the `flipFunction` method and would not necessarily work here.

src/environment/SCRF.cpp
Gabrielgerez: I am modifying class data in this one. As for the pointer, I think it makes sense to just keep it as copy (it is a pointer afterall).

src/environment/SCRF.cpp
Gabrielgerez: the `density::compute` method discards `const` qualifier, according to the compile error. I think the `cplxfunc::add` and `multiply` also do this  for the input `ComplexFunction`.

Gabrielgerez: maybe pass the electronic density directly?

src/environment/SCRF.cpp
robertodr: can't see that it's passed by `const` reference? Did you push?

src/environment/SCRF.cpp
```diff
@@ -48,37 +46,24 @@ using OrbitalVector_p = std::shared_ptr<mrchem::OrbitalVector>;
 
 namespace mrchem {
 
-SCRF::SCRF(Permittivity e, const Nuclei &N, PoissonOperator_p P, DerivativeOperator_p D, double orb_prec, int kain_hist, int max_iter, bool acc_pot, bool dyn_thrs, std::string density_type)
-        : accelerate_Vr(acc_pot)
-        , dynamic_thrs(dyn_thrs)
+SCRF::SCRF(Permittivity &e, const Density &rho_nuc, PoissonOperator_p P, DerivativeOperator_p D, int kain_hist, int max_iter, bool dyn_thrs, std::string density_type)
```
robertodr, line 49:
```suggestion
SCRF::SCRF(const Permittivity &e, const Density &rho_nuc, PoissonOperator_p P, DerivativeOperator_p D, int kain_hist, int max_iter, bool dyn_thrs, const std::string &density_type)
```

src/environment/SCRF.cpp
robertodr: I agree it would serve the same purpose, but it's at least clearer what is happening. I'm not the only one who's been confused by `flipFunction(true)` and `flipFunction(false)` :)

src/environment/SCRF.cpp
robertodr: Yes, I see that adding the `const` qualifier to the function is not possible. Still, you can pass `V_vac` by `const` reference. As for passing the `shared_ptr`: it is a pointer, but not all operations on it have the same cost as with a raw pointer. There are different overheads to the different ways of passing it among functions.

src/environment/SCRF.cpp
robertodr: > maybe pass the electronic density directly?

that's indeed what I ended up doing in #456 😅

src/environment/SCRF.h
Gabrielgerez: removed them now

src/qmoperators/two_electron/FockBuilder.cpp
Gabrielgerez: I think this adds a bug in fockbuilder and SCFEnergy because of the c++ versioning (this wasnt available before c++17) I have added a small fix, but we should probably work on this in the future

Gabrielgerez:
@robertodr @gitpeterwind Now I have fixed what was needed and updated the input parsing (since i removed a keyword) and tests. All tests should be passing now.
Sorry for the wait as well, attempted a big refactor to try and shave off the last few gb and it took longer than expected, so i had to branch it off

src/environment/SCRF.cpp
Gabrielgerez: I pushed it yesterday, i think it should be there now

src/environment/SCRF.cpp
Gabrielgerez: removed this now

src/environment/SCRF.cpp
Gabrielgerez: Ok, I thin it is correct now

src/environment/SCRF.cpp
Gabrielgerez: Done

src/environment/SCRF.cpp
Gabrielgerez: Ok, I am using a const reference to the density now

src/environment/SCRF.cpp
```diff
@@ -48,37 +46,24 @@ using OrbitalVector_p = std::shared_ptr<mrchem::OrbitalVector>;
 
 namespace mrchem {
 
-SCRF::SCRF(Permittivity e, const Nuclei &N, PoissonOperator_p P, DerivativeOperator_p D, double orb_prec, int kain_hist, int max_iter, bool acc_pot, bool dyn_thrs, std::string density_type)
-        : accelerate_Vr(acc_pot)
-        , dynamic_thrs(dyn_thrs)
+SCRF::SCRF(const Permittivity &e, const Density &rho_nuc, PoissonOperator_p P, DerivativeOperator_p D, int kain_hist, int max_iter, bool dyn_thrs, std::string density_type)
```
robertodr, line 49:
```suggestion
SCRF::SCRF(const Permittivity &e, const Density &rho_nuc, PoissonOperator_p P, DerivativeOperator_p D, int kain_hist, int max_iter, bool dyn_thrs, const std::string &density_type)
```

src/environment/SCRF.h
```diff
@@ -30,93 +30,68 @@
...
+    SCRF(const Permittivity &e,
+         const Density &rho_nuc,
          std::shared_ptr<mrcpp::PoissonOperator> P,
          std::shared_ptr<mrcpp::DerivativeOperator<3>> D,
-         double orb_prec,
          int kain_hist,
          int max_iter,
-         bool acc_pot,
          bool dyn_thrs,
          std::string density_type);
```
robertodr, line 46:
```suggestion
         const std::string &density_type);
```

src/environment/SCRF.h
```diff
@@ -30,93 +30,68 @@
...
+    double conv_thrs{1.0};
+    double mo_residual{1.0};
 
     Permittivity epsilon;
 
-    Density rho_nuc;
-    Density rho_ext;
-    Density rho_tot;
+    Density rho_nuc; // As of right now, this is the biggest memory hog.
+    // Alternative could be to precompute its contributions, as a potential is not as heavy as a density (maybe)
```
robertodr, line 76: I think potentials are heavier than densities, as they are less local (@gitpeterwind correct me if I'm wrong) Also, the nuclei are not point charges, so the analytical form of the potential either doesn't exist in closed form (e.g. Gaussian charge densities will have the error function as potential) or it's not trivial to represent. It could be possible to represent the density analytically though... Other options: 
* _not_ store `rho_nuc` inside the class and have it passed from the SCF object at each iteration,
* compute `Vr_n` once (during initialization or at the first iteration) and only store that in this class. Memory-wise though this is probably more expensive...

For linear response this part of the density (and corresponding reaction potential) is not needed at all, so if we figure out a nice way to not store it, those kinds of calculations would be feasible for larger systems.

src/environment/SCRF.cpp
robertodr: What's the conclusion here?

robertodr (approved):
LGTM but there's two comments @Gabrielgerez should chime in on before this can be merged. Also:
* One curiosity: How are you measuring the memory consumption? 
* What are the final before-and-after ballpark numbers on how much memory is needed for a solvent calculation?

src/environment/SCRF.h
gitpeterwind: > I think potentials are heavier than densities, as they are less local

It is more the smoothness than the extension that matters. I think in practice potentials are smaller (less MWnodes) than density normally.

Gabrielgerez:
> LGTM but there's two comments @Gabrielgerez should chime in on before this can be merged. Also:
> 
> * One curiosity: How are you measuring the memory consumption?
> * What are the final before-and-after ballpark numbers on how much memory is needed for a solvent calculation?

Im using the log file output from slurm in betzy an example for a HF water calculation before and after:
before
```
Memory statistics, in GiB:
ID             Alloc   Usage
732516         968.0        
732516.batch   242.0    20.0
732516.orted   726.0    57.5
```

after:
```
Memory statistics, in GiB:
ID             Alloc   Usage
732519         968.0        
732519.batch   242.0     9.1
732519.orted   726.0    22.7
```

When testing we were using a PBE0 calculation, where the gain showed a bit more, but now I am doing some proper scaling calculations with different sizes.

Gabrielgerez:
> LGTM but there's two comments @Gabrielgerez should chime in on before this can be merged. Also:
> 
> * One curiosity: How are you measuring the memory consumption?
> * What are the final before-and-after ballpark numbers on how much memory is needed for a solvent calculation?

For the memory necessary for a solvent calculation, again an example of HF water
gas phase
```
Memory statistics, in GiB:
ID             Alloc   Usage
732514         968.0        
732514.batch   242.0     6.9
732514.orted   726.0    12.9
```
solvent (eps=2.0, default for everything else)
```
Memory statistics, in GiB:
ID             Alloc   Usage
732519         968.0        
732519.batch   242.0     9.1
732519.orted   726.0    22.7
```

robertodr:
Looks very good! What is the difference between batch and orted rows?

Gabrielgerez:
> Looks very good! What is the difference between batch and orted rows?

I was hoping @gitpeterwind  could answer, I am assuming one is the memory used by the main process while the other is the shared one. The machine docs do not have much documentation on this, that I could easily find

gitpeterwind:
> Looks very good! What is the difference between batch and orted rows?

"Batch" is the master compute node, while "orted" (Open Runtime Environment Daemon) is the sum of all others compute nodes. Your run on 4 compute node , each of them has 242 GB available (3*242=726).

robertodr:
So the sum of the two is the estimate of the memory consumption, right? If yes, the cost of the solvent went from ~4x to ~1.6x the cost of the vacuum calculation. Impressive!

src/environment/SCRF.h
robertodr: So storing `Vr_nuc` only could lead to further savings?

src/environment/SCRF.cpp
Gabrielgerez: oh yeah, I think since this wont save any memory (as far as I can see) and is meant more for ease of understanding the code flow, it can probably be incorporated in another PR.

gitpeterwind:
> So the sum of the two is the estimate of the memory consumption, right? 
Yes, but the most relevant is the memory usage of the most memory consuming compute-node (normally master): it does not help to have compute nodes that consume little memory, if one of the compute nodes needs more memory than available. (compute nodes can not use memory from other compute nodes)

PR #461, opened 2023-08-24
bug in disjoin and adjoin when spin is not paired

gitpeterwind:
There was a big bug when running parallel jobs with unpaired electrons.
When splitting the orbitalvector (or joining them) into alpha beta, the ownership was getting wrong, since ownership is determined by the rank in the vector.
Now the orbital::adjoin and orbital::disjoin explicitly send the orbitals if needed.


src/initial_guess/core.cpp
```diff
@@ -129,8 +129,8 @@ bool initial_guess::core::setup(OrbitalVector &Phi, double prec, const Nuclei &n
     initial_guess::core::rotate_orbitals(Phi, prec, U, Psi);
     initial_guess::core::rotate_orbitals(Phi_a, prec, U, Psi);
     initial_guess::core::rotate_orbitals(Phi_b, prec, U, Psi);
-    for (auto &phi_a : Phi_a) Phi.push_back(phi_a);
-    for (auto &phi_b : Phi_b) Phi.push_back(phi_b);
+    Phi = orbital::adjoin(Phi_a, Phi_b);
+
```
stigrj, line 133: In _principle_, there could be orbitals in all three `OrbitalVectors` at this point, `Phi -> Paired`, `Phi_a -> Alpha` and `Phi_b -> Beta`. That was the reason for using `push_back` instead of `adjoin` here. You can just append to `Phi` in two steps:
```
Phi = orbital::adjoin(Phi, Phi_a);
Phi = orbital::adjoin(Phi, Phi_b);
``` 
There shouldn't be much overhead for appending to an empty vector, since the MPI call inside `adjoin` only happens if the rank changes.

stigrj: Same comment applies to all initial guesses...

stigrj (changes requested):

src/initial_guess/core.cpp
gitpeterwind: That was actually also required for the direct paired case!

gitpeterwind: For n=2 there is only one worker, that was trying to send orbitals to itself. 
Not very elegant, but it should be safe.

stigrj (approved):

Issue #419, opened 2022-08-17
Questionable radii in periodictable.py


Gabrielgerez:
There is no specification on where the radii here come from and what units they are

robertodr:
Can we change them to the Bondi-Mantina set?

stigrj:
Can this be fetched from QCElemental, like @Andersmb did for the physical constants?

robertodr:
Yes, that would work: they use Bondi-Mantina values for van der Waals radii. However, in PCM one needs to use 1.2 for hydrogen  instead of 1.1 in the Bondi-Mantina table. I guess it's easy to fix in `validate_cavity` though.

PR #463, opened 2023-09-20
Radii fix

Gabrielgerez:
Finally I implement a simple fix for the default radii for the solvent. This has been an open issue since #419. 
Its not the prettiest fix, but I think it does its job.


stigrj:
I would really like to avoid introducing `qcelemental` as a runtime dependency.

If I understand the `PhysicalConstants` analogue correctly, `qcelemental` is used only as a dev dependency when updating the input `template.yml` file. Then, in `update_input_parser.py`, the parameters are fetched from `qcelemental` and put as default values in a dedicated `Constants` section in the regular input. Not sure if we want to pass all these radii through the standard input file, though?

How about having `update_input_parser.py` write directly to the `periodictable.py` file?

Any thoughts @robertodr ?

stigrj:
Actually, perhaps it's not such a bad idea to pass the entire periodic table through the regular input file? In general we would of course fall back to default values, so the regular user would never have to bother with it. But then the advanced user could easily change a single parameter directly in the input file for use in a particular run, without hard coding a change in a static `periodictable.py` file.

robertodr:
I agree with Stig that we should *not* add Python packages as required production dependencies. They should all stay as development dependencies and the data needed from them should be saved in the `mrchem` Python module, which only needs a working Python interpreter to do its job.

I think it's OK to update `periodictable.py` using `qcelemental` and I think passing it through the parser as done with the constants is a good idea, since it's exposed to the advanced user.

Gabrielgerez:
I have now added the periodic table to the template through the update_input_parser. There are some issues, for example there are elements for which there are no valid vdw-radii in the mantina set. Another one is that the element `No` is confusing the parser since `no` is a reserved yaml keyword for a boolean false (also the `yes` keyword for true)

Gabrielgerez:
Tests are failing, have added a lot of the files that are under `python/mrchem/input_parser/plumbing/pyparsing` as it seems there is a bit of version issues. Will most likely revert these changes later

Gabrielgerez:
Everything should be working now. Have some issues with Nobellium and elements without radius entry in the mantina set, but otherwise it should be ok.

python/mrchem/update_input_parser.py
```diff
@@ -34,6 +34,8 @@
 import argparse
```
robertodr, line 34: Did you run this file through the Black formatter?

python/mrchem/update_input_parser.py
```diff
@@ -71,6 +73,73 @@ def update_constants():
...
+            symbol = '"No"'
+        else:
+            name = str(ele)
+            symbol = vals[4]
+        
+        context = qcel.VanderWaalsRadii("MANTINA2009")
+        aa2bohr = qcel.constants.conversion_factor("angstrom", "bohr")
+        
+        if (ele.lower() == "h"):
+            radius = 1.2*aa2bohr
```
robertodr, line 106: Why convert to atomic units here? The `periodictable.py` is human-readable, so I think we should keep things in Angstrom.

Gabrielgerez: well, all the radii are in atomic units already, so I would be converting one way or the other. I can do it the other way around if it is better

python/mrchem/update_input_parser.py
```diff
@@ -71,6 +73,73 @@ def update_constants():
...
+        else:   
+            try:
+                radius = qcel.vdwradii.get(ele)
+            except:
+                #FIXME this is a hack for the elements which have no radius data in the mantina set
+                radius = vals[0]
```
robertodr, line 112: I think you can set `radius = None` when there is not available radius in the table. You can check on `None` in the parser helper for the solvent and throw an error at that point.

python/mrchem/validators.py
```diff
@@ -319,19 +320,23 @@ def validate_cavity(self):
...
         if self.cavity_mode == "atoms":
             coords = deepcopy(self.atomic_coords)
-            radii = [
-                PeriodicTable[label.lower()].radius for label in self.atomic_symbols
-            ]
+            radii = []
+            for label in self.atomic_symbols:
+                radius = self.user_dict["Elements"][label.lower()]["vdw-radius"]
+                radii.append(radius)
+
```
robertodr, line 327: less lines of code
```suggestion
            radii = [self.user_dict["Elements"][x.lower()]["vdw-radius"] for x in self.atomic_symbols]
```

robertodr (changes requested):
I think you need to run Black and isort on `periodictable.py` and `validators.py`

python/mrchem/update_input_parser.py
Gabrielgerez: I was trying this, but I just cannot get the typing for the yaml entry right. Do you know what we are using for `NoneType` there?

python/mrchem/update_input_parser.py
Gabrielgerez: no

python/mrchem/update_input_parser.py
robertodr: When dumping the YAML, I'd leave them in Angstrom.

python/mrchem/update_input_parser.py
robertodr: Checked and it's not possible, because we never implemented `Union` types (such as `Optional[float]`) in the parser. So the check can be done by using a sentinel value instead, _e.g._ `-1`.

Gabrielgerez: Ok, I'll set it to -1.0 for now then.

Gabrielgerez:
> I think you need to run Black and isort on `periodictable.py` and `validators.py`

Do we have a special configuration file for this? I ran `isort --profile black` on both files, but it barely changed anything

python/mrchem/update_input_parser.py
Gabrielgerez: Using angstrom for everything now

python/mrchem/update_input_parser.py
Gabrielgerez: now it is, but barely any change

robertodr:
So for formatting it's:
`black <files-to-format>`
and for sorting the imports it's:
`isort <files-to-sort>`

python/mrchem/update_input_parser.py
```diff
@@ -71,6 +71,139 @@ def update_constants():
...
+            # FIXME this is a hack to get around the fact that "no" is a reserved keyword in YAML
+            name = '"no"'
+            symbol = '"No"'
+        else:
+            name = str(ele)
+            symbol = vals[4]
+
+        context = qcel.VanderWaalsRadii("MANTINA2009")
+
+        if ele.lower() == "h":
```
robertodr, line 98: Adding a comment as to why we patch the value for Hydrogen might be useful for who reads this code in the future.
```suggestion
        if ele.lower() == "h":
           # these vdw radii are used with a continuum solvation model and 1.2 AA for hydrogen is more appropriate than the default
```

python/mrchem/validators.py
```diff
@@ -318,13 +319,27 @@ def validate_cavity(self):
 
         # the centers of the spheres are the same as the atoms
         if self.cavity_mode == "atoms":
+
+            def radiusNotFound(x):
+                # This raises an exception in the list comprehension if the radius is not valid.
+                raise ValueError(
+                    "The vdw-radius of element {x} is not defined in the mantina set".format(x)
```
robertodr, line 326: f-strings should be preferred:
```suggestion
                    f"The vdw-radius of element {x} is not defined in the mantina set"
```

robertodr (approved):
Two more small comments, but otherwise looks good!

Gabrielgerez:
> Two more small comments, but otherwise looks good!

Alright, i have now added all suggestions :)
We can finally fix this bug

PR #464, opened 2023-09-27
add readthedocs.yml, same as for MRCPP

Gabrielgerez:
Hopefully fixes rtd-builds for mrchem


stigrj (approved):
Seems to work, thanks!

PR #465, opened 2023-10-02
Refactor the Permittivity function

Gabrielgerez:
Create a Class "StepFunction" which sets up the base structure of any Permittivity-like function. This is in preparation to the pb-solvation and other work, so we can avoid repeating code

### DIFF
--- a/src/environment/Permittivity.cpp
+++ b/src/environment/Permittivity.cpp
@@ -38,11 +38,4 @@ double Permittivity::evalf(const mrcpp::Coord<3> &r) const {
     return this->in * std::exp(std::log(this->out / this->in) * (1 - c_pin->evalf(r)));
 }
 
-void Permittivity::printHeader() const {
-    mrcpp::print::header(0, "Solvation Cavity");
-    print_utils::text(0, "Formulation", this->formulation, true);
-    print_utils::scalar(0, "Dielectric constant", getValueIn(), "(in)", 6);
-    print_utils::scalar(0, "", getValueOut(), "(out)", 6);
-}
-
 } // namespace mrchem
--- a/src/environment/Permittivity.cpp
+++ b/src/environment/Permittivity.cpp
@@ -30,7 +30,7 @@
 namespace mrchem {
 
 Permittivity::Permittivity(const mrchem::Cavity cavity, double epsilon_in, double epsilon_out, std::string formulation)
-        : ShiftFunction(cavity, epsilon_in, epsilon_out) {}
+        : StepFunction(cavity, epsilon_in, epsilon_out) {}
 
 double Permittivity::evalf(const mrcpp::Coord<3> &r) const {
     auto epsilon = this->in * std::exp(std::log(this->out / this->in) * (1 - this->cavity.evalf(r)));
--- a/src/environment/Permittivity.cpp
+++ b/src/environment/Permittivity.cpp
@@ -35,8 +35,7 @@ Permittivity::Permittivity(std::shared_ptr<mrchem::Cavity> cavity, double epsilo
 
 double Permittivity::evalf(const mrcpp::Coord<3> &r) const {
     auto c_pin = this->cavity;
-    auto epsilon = this->in * std::exp(std::log(this->out / this->in) * (1 - c_pin->evalf(r)));
-    return (this->inverse) ? 1.0 / epsilon : epsilon;
+    return this->in * std::exp(std::log(this->out / this->in) * (1 - c_pin->evalf(r)));
 }
 
 void Permittivity::printHeader() const {
--- a/src/environment/Permittivity.cpp
+++ b/src/environment/Permittivity.cpp
@@ -29,17 +29,19 @@
 
 namespace mrchem {
 
-Permittivity::Permittivity(const mrchem::Cavity cavity, double epsilon_in, double epsilon_out, std::string formulation)
-        : StepFunction(cavity, epsilon_in, epsilon_out) {}
+Permittivity::Permittivity(std::shared_ptr<mrchem::Cavity> cavity, double epsilon_in, double epsilon_out, std::string formulation)
+        : StepFunction(cavity, epsilon_in, epsilon_out)
+        , formulation(formulation) {}
 
 double Permittivity::evalf(const mrcpp::Coord<3> &r) const {
-    auto epsilon = this->in * std::exp(std::log(this->out / this->in) * (1 - this->cavity.evalf(r)));
-    return (this->inverse) ? 1.0 / epsilon : epsilon; 
+    auto c_pin = this->cavity;
+    auto epsilon = this->in * std::exp(std::log(this->out / this->in) * (1 - c_pin->evalf(r)));
+    return (this->inverse) ? 1.0 / epsilon : epsilon;
 }
 
 void Permittivity::printHeader() const {
     mrcpp::print::header(0, "Solvation Cavity");
-    print_utils::text(0, "Formulation", getFormulation(), true);
+    print_utils::text(0, "Formulation", this->formulation, true);
     print_utils::scalar(0, "Dielectric constant", getValueIn(), "(in)", 6);
     print_utils::scalar(0, "", getValueOut(), "(out)", 6);
 }
--- a/src/environment/Permittivity.cpp
+++ b/src/environment/Permittivity.cpp
@@ -30,13 +30,10 @@
 namespace mrchem {
 
 Permittivity::Permittivity(const mrchem::Cavity cavity, double epsilon_in, double epsilon_out, std::string formulation)
-        : epsilon_in(epsilon_in)
-        , epsilon_out(epsilon_out)
-        , formulation(formulation)
-        , cavity(cavity) {}
+        : ShiftFunction(cavity, epsilon_in, epsilon_out) {}
 
 double Permittivity::evalf(const mrcpp::Coord<3> &r) const {
-    auto epsilon = epsilon_in * std::exp(std::log(epsilon_out / epsilon_in) * (1 - this->cavity.evalf(r)));
+    auto epsilon = this->in * std::exp(std::log(this->out / this->in) * (1 - this->cavity.evalf(r)));
     if (inverse) {
         return 1 / epsilon;
     } else {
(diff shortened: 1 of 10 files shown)


robertodr (commented):
Before I get into actually reviewing, I believe what you've implemented is a [step function](https://en.wikipedia.org/wiki/Step_function)

Gabrielgerez:
> Before I get into actually reviewing, I believe what you've implemented is a [step function](https://en.wikipedia.org/wiki/Step_function)

ah, yes that is the name. I'll rename it properly

src/environment/Permittivity.cpp
```diff
@@ -30,81 +30,22 @@
 namespace mrchem {
 
 Permittivity::Permittivity(const mrchem::Cavity cavity, double epsilon_in, double epsilon_out, std::string formulation)
-        : epsilon_in(epsilon_in)
-        , epsilon_out(epsilon_out)
-        , formulation(formulation)
-        , cavity(cavity) {}
+        : StepFunction(cavity, epsilon_in, epsilon_out) {}
 
 double Permittivity::evalf(const mrcpp::Coord<3> &r) const {
```
robertodr, line 35: Maybe instead of changing the state of the class (`this->inverse`) before calling `evalf`, you can add a boolean to the function?
```suggestion
double Permittivity::evalf(const mrcpp::Coord<3> &r, bool inverse) const {
```

robertodr: Maybe it violates inheritance though 😞

src/environment/Permittivity.cpp
```diff
@@ -30,81 +30,22 @@
...
-    auto epsilon = epsilon_in * std::exp(std::log(epsilon_out / epsilon_in) * (1 - this->cavity.evalf(r)));
+    auto epsilon = this->in * std::exp(std::log(this->out / this->in) * (1 - this->cavity.evalf(r)));
     if (inverse) {
         return 1 / epsilon;
     } else {
         return epsilon;
     }
```
robertodr, line 41: I like ternary `if`s, but I leave it up to you to decide what's preferable here?
```suggestion
    return (this->inverse) : 1.0 / epsilon ? epsilon; 
```

src/environment/StepFunction.cpp
```diff
@@ -0,0 +1,104 @@
...
+ * <https://mrchem.readthedocs.io/>
+ */
+
+#include "StepFunction.h"
+#include "Cavity.h"
+#include <MRCPP/MWFunctions>
+
+namespace mrchem {
+
+StepFunction::StepFunction(const mrchem::Cavity cavity, double val_in, double val_out)
```
robertodr, line 32:
```suggestion
StepFunction::StepFunction(const mrchem::Cavity &cavity, double val_in, double val_out)
```

robertodr: Please, have a read at this to understand the overhead of passing arguments to functions and what the best practices are for the desired semantics: https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#fcall-parameter-passing

src/environment/StepFunction.cpp
```diff
@@ -0,0 +1,104 @@
...
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with MRChem.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * For information on the complete list of contributors to MRChem, see:
+ * <https://mrchem.readthedocs.io/>
+ */
+
+#include "StepFunction.h"
+#include "Cavity.h"
+#include <MRCPP/MWFunctions>
```
robertodr, line 28:
```suggestion
#include <MRCPP/MWFunctions>

#include "StepFunction.h"
#include "Cavity.h"
```
surprised that `clang-format` didn't reorder the headers...

robertodr: for reference, headers inclusion with angle brackets should be done before that with quotes.

src/environment/StepFunction.cpp
```diff
@@ -0,0 +1,104 @@
...
+        o_coord << std::setw(w4) << std::setprecision(6) << std::fixed << x;
+        o_coord << std::setw(w4) << std::setprecision(6) << std::fixed << y;
+        o_coord << std::setw(w4) << std::setprecision(6) << std::fixed << z;
+        println(0, o_coord.str());
+    }
+    mrcpp::print::separator(0, '=', 2);
+}
+
+void StepFunction::printHeader() const {
+    mrcpp::print::header(0, "Shift Function of Cavity values");
```
robertodr, line 98:
```suggestion
    mrcpp::print::header(0, "Step Function of Cavity values");
```

src/environment/StepFunction.h
```diff
@@ -0,0 +1,79 @@
...
+ * For information on the complete list of contributors to MRChem, see:
+ * <https://mrchem.readthedocs.io/>
+ */
+
+#pragma once
+
+#include "Cavity.h"
+#include "utils/print_utils.h"
+#include <MRCPP/MWFunctions>
+#include <MRCPP/Printer>
```
robertodr, line 31:
```suggestion
#include <MRCPP/MWFunctions>
#include <MRCPP/Printer>

#include "Cavity.h"
#include "utils/print_utils.h"
```

src/environment/StepFunction.h
```diff
@@ -0,0 +1,79 @@
...
+
+class StepFunction : public mrcpp::RepresentableFunction<3> {
+public:
+    /** @brief Standard constructor. Initializes the #cavity, #in and #out with the input parameters.
+     *  @param cavity interlocking spheres of Cavity class.
+     *  @param epsilon_in permittivity inside the #cavity.
+     *  @param epsilon_out permittivity outside the #cavity.
+     * available as of now.
+     */
+    StepFunction(const Cavity cavity, double val_in, double val_out);
```
robertodr, line 57:
```suggestion
    StepFunction(const Cavity &cavity, double val_in, double val_out);
```

src/environment/StepFunction.h
```diff
@@ -0,0 +1,79 @@
...
+
+    Cavity getCavity() const { return this->cavity; }
+
+    void printParameters() const;
+    virtual void printHeader() const;
+
+protected:
+    double in;                            //!< Value of the function inside the #cavity.
+    double out;                           //!< value of the function outside the #cavity.
+    std::string formulation = "standard"; //!< Formula/equation/algorithm used for shifting values.
```
robertodr, line 75: Where is this used? I cannot see that you can set it from the CTOR or after initialization...

robertodr (commented):
Potentially we can have many step functions in a calculation. The way it's written now, each will make their own copy of the `Cavity` function, which can possibly become a memory bottleneck. Maybe instead of storing a `Cavity` the classes should store a `std::shared_ptr<Cavity>`?

src/environment/StepFunction.h
Gabrielgerez: I added it to fill in for the virtual `printParameters`, that way the `Permittivity` and `DHScreening` need only change its value without rewriting the `printParameters` function.

Gabrielgerez:
> Potentially we can have many step functions in a calculation. The way it's written now, each will make their own copy of the `Cavity` function, which can possibly become a memory bottleneck. Maybe instead of storing a `Cavity` the classes should store a `std::shared_ptr<Cavity>`?

This is a good idea, but the `DHScreening` function might potentially need a different cavity than the one for the `Permittivity` (Different radii due to Ion accessible surface and such)

Gabrielgerez:
> > Potentially we can have many step functions in a calculation. The way it's written now, each will make their own copy of the `Cavity` function, which can possibly become a memory bottleneck. Maybe instead of storing a `Cavity` the classes should store a `std::shared_ptr<Cavity>`?
> 
> This is a good idea, but the `DHScreening` function might potentially need a different cavity than the one for the `Permittivity` (Different radii due to Ion accessible surface and such)

This was really a fix waiting to happen (cavity was passed as a `shared_ptr` from `molecule` in `driver.cpp`), we can let whoever (us) wants to do anything in the future deal with cavity ownership

src/environment/StepFunction.h
Gabrielgerez: I removed it now, it really wasn't needed

src/environment/Permittivity.cpp
Gabrielgerez: This seems like a nightmare to implement. I think I need to make overloads for the apply and multiply methods of mrcpp for that to work

Gabrielgerez: Turns out this was actually an easy fix with lambda functions :/

robertodr: Hail the lambda functions! 🎊 $\lambda$!

src/environment/SCRF.cpp
```diff
@@ -118,13 +119,11 @@ mrcpp::ComplexFunction SCRF::solvePoissonEquation(const mrcpp::ComplexFunction &
     mrcpp::ComplexFunction Vr_np1;
     Vr_np1.alloc(NUMBER::Real);
 
-    this->epsilon.flipFunction(true);
-
+    auto eps_inv_func = mrcpp::AnalyticFunction<3>([this](const mrcpp::Coord<3> &r) { return 1.0 / this->epsilon.evalf(r); });
```
robertodr, line 122: very good 👍🏻

src/environment/StepFunction.h
```diff
@@ -0,0 +1,74 @@
...
+ *
+ * You should have received a copy of the GNU Lesser General Public License
+ * along with MRChem.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * For information on the complete list of contributors to MRChem, see:
+ * <https://mrchem.readthedocs.io/>
+ */
+
+#pragma once
+
```
robertodr, line 27:
```suggestion
#include <memory>

```

src/environment/StepFunction.h
```diff
@@ -0,0 +1,74 @@
...
+     */
+    StepFunction(std::shared_ptr<Cavity> cavity, double val_in, double val_out);
+
+    auto getValueIn() const { return this->in; }
+    auto getValueOut() const { return this->out; }
+
+    std::shared_ptr<Cavity> getCavity() const { return this->cavity; }
+
+    void printParameters() const;
+    virtual void printHeader() const;
```
robertodr, line 66: there is some duplication between base and derived class. 

We can add a free function (outside the class) to generate headers:
```c++
void print_header(const std::string &header, const std::string &formulation, double in_value, double out_value) {
    mrcpp::print::header(0, header);
    print_utils::text(0, "Formulation", formulation, true);
    print_utils::scalar(0, "Value inside Cavity", in_value, "(in)", 6);
    print_utils::scalar(0, "Value outside Cavity", out_value, "(out)", 6);
}
```
then in this class:
```c++
virtual void printHeader() const { print_header("Step function of Cavity values", "Standard", getValueIn(), getValueOut()); }
```
while in `Permittivity`:
```c++
void printHeader() const override { print_header("Solvation Cavity", this->formulation, getValueIn(), getValueOut()); }
```

src/environment/StepFunction.cpp
```diff
@@ -0,0 +1,107 @@
...
+ * along with MRChem.  If not, see <https://www.gnu.org/licenses/>.
+ *
+ * For information on the complete list of contributors to MRChem, see:
+ * <https://mrchem.readthedocs.io/>
+ */
+
+#include <MRCPP/MWFunctions>
+
+#include "Cavity.h"
+#include "StepFunction.h"
```
robertodr, line 29: in the implementation file the order of inclusion is:
```suggestion
#include "StepFunction.h"

#include <MRCPP/MWFunctions>

#include "Cavity.h"
```

src/environment/StepFunction.h
robertodr: you're using `std::shared_ptr` The compiler is not complaining that it cannot find it because it's been included by some other of the headers.

src/environment/StepFunction.h
Gabrielgerez: Slightly harder than expected, ended up adding it to the `detail` namespace

robertodr (approved):

PR #466, opened 2023-10-02
Fix incoming build error in rtd

Gabrielgerez:
the key `build.os` will start being a requirement soon (see [link](https://blog.readthedocs.com/use-build-os-config/) )
Hopefully this fixes any issues

### DIFF
--- a/.readthedocs.yml
+++ b/.readthedocs.yml
@@ -1,5 +1,10 @@
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.6"
+
 sphinx:
   configuration: doc/conf.py
 
@@ -8,6 +13,5 @@ formats:
   - pdf
 
 python:
-  version: 3.6
   install:
     - requirements: doc/requirements.txt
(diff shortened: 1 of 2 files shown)


.readthedocs.yml
```diff
@@ -1,5 +1,10 @@
 version: 2
 
+build:
+  os: ubuntu-22.04
+  tools:
+    python: "3.6"
```
robertodr, line 6: I think we can use 3.9 or later, also because Python 3.6 is unmaintained.

robertodr (approved):
LGTM

.readthedocs.yml
robertodr, line 6:
```suggestion
    python: "3.9"
```

robertodr: well, using 3.9 broke RTD 🤔

robertodr: It's some bug in `sphinx_rtd_theme` with the latest `sphinx`. Pinned the latter to `<7` and now things work.

Gabrielgerez:
I'll then merge this, @robertodr could you check the PR on MRCPP. it is the same changes

PR #467, opened 2023-10-11
Molecules in high pressure environments

hugoastrom:
- [ ] Confinement potential class
- [ ]

### DIFF
--- a/src/analyticfunctions/ConfinementFunction.cpp
+++ b/src/analyticfunctions/ConfinementFunction.cpp
@@ -1,0 +1,21 @@
+#include "ConfinementFunction.h"
+#include "utils/math_utils.h"
+
+namespace mrchem {
+
+  ConfinementFunction::ConfinementFunction(const double r_0, const int N)
+    : radius(r_0)
+    , stiffness(N) {}
+
+  double ConfinementFunction::evalf(const mrcpp::Coord<3> &r) const {
+    double f = 0.0;
+    const mrcpp::Coord<3> center = {0.0, 0.0, 0.0};
+    const int N = this->stiffness;
+    auto r_0 = this->radius;
+
+    f += std::pow(math_utils::calc_distance(r, center) / r_0, N);
+
+    return f;
+  }
+
+} // namespace mrchem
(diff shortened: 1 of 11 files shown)


src/analyticfunctions/ConfinementFunction.cpp
```diff
@@ -0,0 +1,21 @@
...
+
+namespace mrchem {
+
+  ConfinementFunction::ConfinementFunction(const double r_0, const int N)
+    : radius(r_0)
+    , stiffness(N) {}
+
+  double ConfinementFunction::evalf(const mrcpp::Coord<3> &r) const {
+    double f = 0.0;
+    const mrcpp::Coord<3> center = {0.0, 0.0, 0.0};
```
ilfreddy, line 12: Is the center hard-wired to the origin? Maybe it would be wise to be able to pass this from input.

src/driver.cpp
```diff
@@ -268,6 +269,7 @@ json driver::scf::run(const json &json_scf, Molecule &mol) {
         auto relativity = json_scf["scf_solver"]["relativity"];
         auto environment = json_scf["scf_solver"]["environment"];
         auto external_field = json_scf["scf_solver"]["external_field"];
+	auto confinement_potential = json_scf["scf_solver"]["confinement_potential"];
```
ilfreddy, line 272: Seems you have some indentation issues which should be fixed

src/driver.cpp
```diff
@@ -1123,6 +1128,15 @@ void driver::build_fock_operator(const json &json_fock, Molecule &mol, FockBuild
         auto V_ext = std::make_shared<ElectricFieldOperator>(field, r_O);
         F.getExtOperator() = V_ext;
     }
+    ///////////////////////////////////////////////////////////
+    ///////////////   Confinement Potential   /////////////////
+    ///////////////////////////////////////////////////////////
+    if (json_fock.contains("confinement_operator")) {
+      auto r_0 = json_fock["confinement_operator"]["r_0"];
```
ilfreddy, line 1135: indentation should be 4 spaces :-)

ilfreddy (commented):
I left a few comments for you to consider.

src/analyticfunctions/ConfinementFunction.cpp
hugoastrom: The plan is to obtain the centers as it is implemented in the Cavity function but it is under construction

Gabrielgerez:
About the indentation, I think you can set git so that it can automatically format your files before committing changes

PR #471, opened 2023-11-10
Linear response with solvent

robertodr:
Rehash of #456


doc/users/user_ref.rst
```diff
@@ -920,6 +914,28 @@ User input reference
         **Predicates**
           - ``value.lower() in ['exponential']``
       
+      :red:`Sections`
+       :outside: Parameters for the continuum solvent outside the cavity.
```
Gabrielgerez, line 918: Not very fond of the naming `outside` maybe we can call it `continuum` or `dielectric_continuum`.

robertodr: good idea.

python/mrchem/input_parser/docs/user_ref.rst
```diff
@@ -920,6 +914,28 @@ User input reference
         **Predicates**
           - ``value.lower() in ['exponential']``
       
+      :red:`Sections`
+       :outside: Parameters for the continuum solvent outside the cavity.
```
Gabrielgerez, line 918: Same comment as above, we should use something more memorable

python/template.yml
```diff
@@ -936,20 +936,40 @@ sections:
...
             type: str
             default: exponential
             predicates:
               - value.lower() in ['exponential']
             docstring: |
               Formulation of the Permittivity function. Currently only the
-              exponential is used.
+              exponential is available.
+        sections:
+          - name: outside
```
Gabrielgerez, line 948: Naming here as well

src/driver.cpp
```diff
@@ -1043,26 +1043,62 @@ void driver::build_fock_operator(const json &json_fock, Molecule &mol, FockBuild
...
 
-        Permittivity dielectric_func(mol.getCavity_p(), eps_i, eps_o, formulation);
-        dielectric_func.printParameters();
+        // compute nuclear charge density
         Density rho_nuc(false);
         rho_nuc = chemistry::compute_nuclear_density(poisson_prec, nuclei, 100);
 
+        // decide which permittivity to use outside of the cavity
+        auto eps_o = [order, is_dynamic, noneq, eps_s, eps_d] {
+            if (order == 1 && noneq && is_dynamic) {
```
Gabrielgerez, line 1074: why do we need both `is_dynamic` and `noneq` it seems that both are needed for the response solvent.

src/environment/SCRF.h
```diff
@@ -43,26 +54,29 @@ class SCRF final {
...
     double setConvergenceThreshold(double prec);
 
-    mrcpp::ComplexFunction &getCurrentReactionPotential() { return this->Vr_n; }
-
     Permittivity &getPermittivity() { return this->epsilon; }
 
     void updateMOResidual(double const err_t) { this->mo_residual = err_t; }
 
-    friend class ReactionPotential;
     auto computeEnergies(const Density &rho_el) -> std::tuple<double, double>;
```
Gabrielgerez, line 66: Since you use `std::tie` as the return statement in `FockBuilder`, should we also just write this as 
```suggestion
    auto computeEnergies(const Density &rho_el) -> std::tie<double, double>;
```
to keep it clear what we are working with?

Gabrielgerez (changes requested):
I left some questions and comments. I really like how the `ReactionPotential` looks now. I mentioned the naming of the `outside` section, but it is not that essential besides aesthetics.

src/driver.cpp
robertodr: you can decide to use the equilibrium formulation even when $\omega \neq 0$

src/environment/SCRF.h
robertodr: No, `tie` is to bind elements of a tuple to existing variables: it's a function not a data container. https://en.cppreference.com/w/cpp/utility/tuple/tie

Gabrielgerez: alright

.gitignore
```diff
@@ -134,3 +134,4 @@ Pipfile.lock
 .vscode
 
 cube_vectors/
+tests/**/*.out
```
Gabrielgerez, line 137: Is this something we want to keep for the master branch?

Gabrielgerez (commented):
I like the new naming of the `outside` section of the input. besides the comment on the `.gitignore` everything looks good to me

.gitignore
robertodr: Perhaps not. I've only added the JSON under `reference` for the new tests though. That's what counts anyway.

PR #475, opened 2023-11-30
Node based DFT

gitpeterwind:
The loop over node is set as the outer loop in MRDFT. For large systems using MPI, that removes the memory intensive intermediate Functions (mostly derivatives) , and is also much faster as a by-product.
The code is also much simpler (only one extra method in Functional, instead of the 4 subclasses for each case).

Also the rotation of the sad initial_guess was very slow and a bottleneck. A new rotation is implemented and the time went down from 200 s to 3 s!

With all the "node_xc" changes in mrcpp and mrchem, the code is much more user friendly. It runs smoothly with 1000 orbitals on Betzy. No need to make special settings at the start to "save" memory.  For even larger systems, the O(N^3) terms (diagonalization of Fock matrix, orthonormalization, localization)  become a bottleneck and should be addressed (using ELPA for example). 

Test valinomycine (300 orbitals) betzy, 4 nodes (can also run on 1 nodes now):
old:
```
                           Building XC operator
 Precision                                  (rel)              1.00000e-05
 Compute rho                     46152 nds         1.41 GB        5.37 sec
 Preprocess input               184672 nds         5.64 GB        0.79 sec
 Evaluate functional            230840 nds         7.04 GB       16.43 sec
 Postprocess potential           92336 nds         2.82 GB        1.72 sec
                         Wall time: 2.46008e+01 sec

Memory statistics, in GiB: 190.0
```
new:
```
                           Building XC operator
 Precision                                  (rel)              1.00000e-05
 Compute rho                     46152 nds         1.41 GB        5.39 sec
 Make potential                  46168 nds         1.41 GB        3.03 sec
                         Wall time: 8.79213e+00 sec

Memory statistics, in GiB: 64.9  (128.6 after only XC upgrade, 81.8 after GenNodes agressive cleaning, the rest because of Bank upgrade and large memory chunks)
```


external/upstream/fetch_mrcpp.cmake
```diff
@@ -39,7 +39,7 @@ else()
     GIT_REPOSITORY
       https://github.com/MRChemSoft/mrcpp.git
     GIT_TAG
-      f8def0a086da6410e5dd8e078de4f6b6305b6ea3
+      83df62a6b2bd2dec8b94064089ebb8641704b2f8
```
ilfreddy, line 42: This must be updated before approval

ilfreddy (approved):

doc/users/user_inp.rst
```diff
@@ -131,6 +132,11 @@ it is likely more efficient to set `bank_size = 0`, otherwise it's recommended
 to use the default. If a particular calculation runs out of memory, it might
 help to increase the number of bank processes from the default value.
 
+The number of threads to use in OpenMP can be forced using the omp_threads flag.
+For MPI runs, it is strongly advized to leave the default, as the optimal value
```
robertodr, line 136:
```suggestion
For MPI runs, it is strongly advised to leave the default, as the optimal value
```

doc/users/schema_input.json
```diff
@@ -24,6 +24,7 @@
   },
   "mpi": {                                   # Section for MPI specification
     "bank_size": int,                        # Number of MPI ranks in memory bank
+    "omp_threads": int,                      # Number of omp threads
```
robertodr, line 27: I don't understand why the number of OpenMP threads has to appear in the input, to be honest.

gitpeterwind: It is to keep a way to force the number of threads, as it is set automatically otherwise. For testing performance, for example, one may want to use less threads than the maximum.

robertodr: And the OMP_NUM_THREADS environment variable isn't enough?

gitpeterwind: For the MPI case the OMP_NUM_THREADS variable is not used. This is because it is often not set automatically by the system and even if set, it will not have the right value. Asking the user to set it, they will most probably not choose the optimal value. (the optimal value is larger than the number of cores divided by the number of MPI processes, because not all the MPI processes are threaded).

gitpeterwind: I understand your question: in an earlier version I used OMP_NUM_THREADS, but then I realized that the only cases were this was useful is in the rare case you do not want to use all the cores. In the very majority of practical situations, the risk of taking a non-optimal value was large.
Good you put that remark, because I had forgotten to update the docs :)

ilfreddy (approved):

PR #479, opened 2024-02-06
Only validate cavity when doing PCM

### LINKED ISSUE #476: Lack of radii in Mantina set should raise a warning, not an error
This line of code stops the program from running if the radius of any species is not in the mantina set regardless if we are running solvent or not
https://github.com/MRChemSoft/mrchem/blame/9de2280c94bf3cd2576778d4c79a7cc3a3d1e5bd/python/mrchem/validators.py#L355
 
I think this should not stop the execution of the parser, but actually raise a warning such that the advanced user can fix the radii after the parser is finished. 
In addition, it should not stop execution of the code if one is not running solvent, which it does now.

What is your opinion on this @robertodr ?

robertodr:
Fix #476

### DIFF
--- a/python/mrchem/validators.py
+++ b/python/mrchem/validators.py
@@ -354,7 +354,7 @@ class MoleculeValidator:
 
             if radii_not_found:
                 raise ValueError(
-                    f"vdw-radius for {', '.join(radii_not_found)} not defined in the Mantina set."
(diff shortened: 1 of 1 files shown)


Gabrielgerez (approved):
thanks for the fix roberto!

PR #480, opened 2024-04-16
Remove dead link to old slack channel

Gabrielgerez:
Removed user support link to the old slack channel. We can probably replace it with something else, but I think it is better to stop users from hoping too much

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,6 @@ chemistry (Hartree-Fock and Density Functional Theory).
 The code is being developed at the Hylleraas Centre for Quantum Molecular
 Sciences at UiT - The Arctic University of Norway.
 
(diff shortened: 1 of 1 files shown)


ilfreddy (commented):
Maybe we should point at zulip instead?

Gabrielgerez:
I wonder if support should be centered around posting issues in the github page. This way we can add necessary bugfixes and tag related issues for context. If it is necessary we could link the user to the zulip chat manually.

ilfreddy:
> I wonder if support should be centered around posting issues in the github page. This way we can add necessary bugfixes and tag related issues for context. If it is necessary we could link the user to the zulip chat manually.

Might be better todiscuss coding issues directly on github

ilfreddy:
I believe this can go once the tests pass.

stigrj (approved):

PR #486, opened 2024-05-14
Option to disable rotation in initial guess

msnik1999:
Added an input option to disable localization / diagonalization of the orbitals before calculating the initial guess energy.

### DIFF
--- a/python/mrchem/update_input_parser.py
+++ b/python/mrchem/update_input_parser.py
@@ -220,20 +220,24 @@ def update_periodic_table():
 
 def run_parselglossy():
     os.chdir(root)
-    cmd = [
-        "parselglossy",
-        "generate",
-        "--template",
-        "template.yml",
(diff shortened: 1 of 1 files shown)


stigrj:
Excellent, can you also add a corresponding `rotate` line in the input schema doc?  https://github.com/MRChemSoft/mrchem/blob/37066bab2cc71dfbcf2fba209c5c0cb21fe357b3/doc/users/schema_input.json#L103

I need to double check why the MPI tests are failing :confused:

stigrj (approved):
Thanks! And looks like the failing MPI test resolved itself :slightly_smiling_face:

stigrj (approved):

stigrj:
I get some random segfaults in "Projecting occupied space" in the response SCF test, but it seems to be an issue also before this PR...

PR #488, opened 2024-06-12
Surface forces

moritzgubler:
This is my implementation of the idea I had of [calculating forces with surface integrals](https://arxiv.org/abs/2410.02299).

It works quite well and the forces are more accurate. Also, mrchem struggles to compute forces when the world precision is 1e-7 or smaller. Then it requires a lot of memory (more than 32 GB for a H2 molecule). This is not the case with my approach . Geometry optimizations seem to be reliable if the stopping criterion is chosen 10 * world_prec.

I have already put all the parameters into the parser, so it is quite simple to test and run my code (look for the section "- name: Forces" in the template.yaml input parser file).

It works with lda and gga functionals and for both closed and open shell systems.
At the moment, there is one thing that might be improved in the future:
- The exact exchange stress is not computed. Therefore forces can not be computed for exact exchange. It shouldn't be too complicated to implement, but in a quick search I did not find a reference that derives it.


robertodr:
hei Moritz, could you add a quick description of how this differs from the Hellmann-Feynman calculation of the forces? Also, I have C++ code that:
* generates (at compile-time) the Lebedev quadrature points _without_ reading them from a file, 
* generates equal-weight grids on the sphere (Leopardi partitioning)

I can share both if it's of interest.

robertodr:
> I find i kind of usefule to be able to exchange the lebvedev file with something else and not have to recompile the code to test things. But if you like it more to avoid reading the file, I can also use your method.

Yes, of course. The approach is the same, since the grids are known beforehand. If you tabulate all of them, there's not much you can swap them with, as their structure is not shared (that I know of) with other grids on the sphere. I dislike reading files at runtime. It's Lebedev, not lebvedev, by the way.

> What would be the purpose of the Leopardi points? are they more accurate than the lebvedev grids?

In terms of accuracy they're similar. They're easier to generate though, so you're not limited to a pre-tabulated set of points and weights.

> I sent you a small description of my method on zulip. I don't want to publish this yet.

Thanks!

moritzgubler:
Can you send me the lebedev cpp library please you were talking about? I can also integrate that to avoid reading the file.

robertodr:
here you go https://github.com/robertodr/balls

moritzgubler:
Thanks, I integrated your library into my code.

moritzgubler:
Open shell calculations now also work, I still need to implement the GGA stress density (see my question on zulip) and make sure, that the forces are calculated using the hellmann-feynman theorem in an exact exchange calculation.

moritzgubler:
Hi @robertodr @ilfreddy 
I implemented now everything I wanted to, cleaned up to code and added a bit of documentation. The new method to calculate the forces now works for LDA and GGA functionals for both restricted and unrestricted calculations. The pulll request is now ready for review.
Best, Moritz

moritzgubler:
I previously tested GGA calculations only for H2 where the forces where the forces were correct. In more complicated systems this is not the case. (LDA is always correct). I will look into it and let you know once I have solved the problem.

moritzgubler:
The GGA bug is fixed now, I forgot to add the divergence terms in the xc potential for the GGA case. They have been added now.
MPI does not work anymore, it segfaults in the calculation of the kinetic stress. I suspect, that I try to access orbitals in an orbitalvector that are stored on another rank. @gitpeterwind Do you know what could be the problem?

src/surface_forces/SurfaceForce.cpp
```diff
@@ -0,0 +1,389 @@
...
+    for (int iOrb = 0; iOrb < Phi.size(); iOrb++) {
+        occ = Phi[iOrb].occ();
+    
+        for (int i = 0; i < nGrid; i++) {
+            pos[0] = gridPos(i, 0);
+            pos[1] = gridPos(i, 1);
+            pos[2] = gridPos(i, 2);
+            n1 = nablaPhi[iOrb][0].real().evalf(pos);
+            n2 = nablaPhi[iOrb][1].real().evalf(pos);
+            n3 = nablaPhi[iOrb][2].real().evalf(pos);
```
moritzgubler, line 194: @gitpeterwind These are the lines that cause the segfault

src/surface_forces/SurfaceForce.cpp
```diff
@@ -0,0 +1,389 @@
...
+    auto D2 = std::make_shared<mrcpp::BSOperator<3>>(*mrchem::MRA, derivOrder2);
+    mrchem::HessianOperator hess(D1, D2, prec);
+    hess.setup(prec);
+
+    std::vector<mrchem::OrbitalVector> nablaPhi;
+    mrchem::OrbitalVector hessRho = hess(rho);
+    for (int i = 0; i < Phi.size(); i++) {
+        nablaPhi.push_back(nabla(Phi[i]));
+    }
+
```
moritzgubler, line 291: Here I create the vector that contains the gradient of all orbitals

moritzgubler:
I can recreate the error with this input:
```
world_prec = 1.0e-4
world_unit = bohr

WaveFunction {
  method = blyp
  restricted = false
}

Properties {
  geometric_derivative = true
}
Forces {
    method = "surface_integrals"
    #method = 'hellmann_feynman'
    surface_integral_precision = medium
}

Molecule {
$coords
H 0.86387418 -0.02789966 0.13501906
Li -0.44463806 -0.04858572 -0.04959119
$end
}
```
I start the simulation with:
```
mrchem --dryrun hli.inp
mpirun -np 4 mrchem.x hli.json
```

src/surface_forces/SurfaceForce.cpp
gitpeterwind: In an OrbitalVector, not all orbitals are directly available to all MPI processes. You should do something like:
     if (mrcpp::mpi::my_orb(phi_i)) {
           pos[0] = gridPos(i, 0);
            pos[1] = gridPos(i, 1);
            pos[2] = gridPos(i, 2);
            n1 = nablaPhi[iOrb][0].real().evalf(pos);
            n2 = nablaPhi[iOrb][1].real().evalf(pos);
            n3 = nablaPhi[iOrb][2].real().evalf(pos);
            stress[i](0, 0) -= occ * n1 * n1;
            stress[i](1, 1) -= occ * n2 * n2;
            stress[i](2, 2) -= occ * n3 * n3;
            stress[i](0, 1) -= occ * n1 * n2;
            stress[i](0, 2) -= occ * n1 * n3;
            stress[i](1, 2) -= occ * n2 * n3;
           }
         }
    }
    And then use      mrcpp::mpi::allreduce_vector to collect all the results. It is a bit difficult with the way you have defined the stress vector. But if you cannot easily define it as a DoubleVector, or DoubleMatrix , I can try to give a detailed way to do it (one way is simply to copy all the values into a DoubleVector, the do the allreduce operation, and copy back).

moritzgubler: That makes sense, thanks for your quick answer. I will use arrays for the stress tensor.

src/surface_forces/SurfaceForce.cpp
gitpeterwind: I think it is necessary (at least it won't harm) to add:
```
std::vector<mrchem::OrbitalVector> nablaPhi;
    mrchem::OrbitalVector hessRho = hess(rho);
    for (int i = 0; i < Phi.size(); i++) {
        nablaPhi.push_back(nabla(Phi[i]));
    }
    nablaPhi.distribute();
    
    ```
(    distribute will tell each element in the vector which rank it has)

File: src/surface_forces/SurfaceForce.cpp (hunk 1 above)
Comment (moritzgubler) [reply src/surface_forces/SurfaceForce.cpp:194]: @gitpeterwind I think it might be a bit more complicated, I need
```
            nablaPhi[iOrb][0]
            nablaPhi[iOrb][1]
            nablaPhi[iOrb][2]
```
all on the same rank. How can I make sure that this is the case?

Comment (gitpeterwind) [reply src/surface_forces/SurfaceForce.cpp:194]: I think it is correct: the last [0],[1],[2] picks out one OrbitalVector, and iOrb only determines the rank. Or did I misunderstood?

Comment (gitpeterwind) [reply src/surface_forces/SurfaceForce.cpp:194]: The rank is just the value of iOrb, i.e. the index in the vector.

Comment (gitpeterwind) [reply src/surface_forces/SurfaceForce.cpp:194]: You can also simply use:
`if (mrcpp::mpi::my_orb(iOrb))`

**Comment by ilfreddy** [comment]
@moritzgubler  would you mind adding also some testing to your code? According to the code coverage report, your patch is only marginally covered. Tests are essential to make sure the code does not get broken by others later on.

File: src/qmoperators/one_electron/HessianOperator.h
Comment (ilfreddy) [inline src/qmoperators/one_electron/HessianOperator.h:1]: Could you add some class documentation?

**Review by ilfreddy — COMMENTED** [review]
Just a general comment from my side: could you add some extra testing:
1. There is no unit-testing as far as I can see
2. There is only one regression test

Some classes could also benefit from some additional documentation.

As for the details of the code, I guess it is best if Stig or Peter comment on those.

File: src/surface_forces/SurfaceForce.cpp (hunk 1 above)
Comment (moritzgubler) [reply src/surface_forces/SurfaceForce.cpp:194]: nablaPhi is a vector of Orbital vectors. iOrb is not an index of an orbitalvecor, it is an index of a standard vector.

Comment (moritzgubler) [reply src/surface_forces/SurfaceForce.cpp:194]: The problem I have is that when I apply the nabla operator on an orbital I get back an orbitalvector where not all orbitals are on the same rank.

Comment (gitpeterwind) [reply src/surface_forces/SurfaceForce.cpp:194]: Your nabla operator will return a vector with 3 orbitals, right?
Then you cannot push_back this into  nablaPhi, since it expects to get an OrbitalVector. 
Shouldn't you should write into nablaPhi it with something like:
```
    std::vector<mrchem::OrbitalVector> nablaPhi;
    for (int i = 0; i < Phi.size(); i++) {
        auto grad=nabla(Phi[i]);
        nablaPhi[0].push_back(grad[0]);
        nablaPhi[1].push_back(grad[1]);
        nablaPhi[2].push_back(grad[2]);
    }
```
nablaPhi[0] is an OrbitalVector. It contain the x component of each orbital.
Or am I completely wrong?

Comment (moritzgubler) [reply src/surface_forces/SurfaceForce.cpp:194]: No, nablaPhi[0][0] contains the x component of the gradient of the zeroth orbital. nablaPhi[0] is an orbital vector that contains the x, y and z component of the gradient of orbital 0.

Comment (moritzgubler) [reply src/surface_forces/SurfaceForce.cpp:194]: Adding the following print to the loop
``` cpp
    for (int i = 0; i < Phi.size(); i++) {
        nablaPhi.push_back(nabla(Phi[i]));
        nablaPhi[i].distribute();
        if (mrcpp::wrk_rank == 0){
            std::cout << "gradient of orbital " << i << " stored on rank: x: " << nablaPhi[i][0].getRank() << " y: " << nablaPhi[i][1].getRank() << " z: " << nablaPhi[i][2].getRank() << std::endl;
        }
    }
```
results in:
```
gradient of orbital 0 stored on rank: x: 0 y: 1 z: 2
gradient of orbital 1 stored on rank: x: 0 y: 1 z: 2
gradient of orbital 2 stored on rank: x: 0 y: 1 z: 2
gradient of orbital 3 stored on rank: x: 0 y: 1 z: 2
```
but I need all components of the gradient on the same rank

gitpeterwind: The OrbitalVector class expects some properties, like that the size is the "usual" number of orbitals, and that each MPI process only access its own indices. If a vector has only have 3 orbitals, you should not define it as an OrbitalVector (and "distribute them is meaningless). Can you simply define `std::vector<std::vector<Orbital>> nablaPhi(3);` ? or do you need some OrbitalVector specific properties? Still Phi[i] will only be defined on the MPI process with the right rank.
Anyway, if you use the `if (mrcpp::mpi::my_orb(phi_i)) {` , do you still get the seg fault?

moritzgubler: This is a good idea. I thought that this wouldn't work because nabla(Phi) returns an orbitalvector. I tried your suggestion and it compiled. I will try to solve the problem like that, thanks for your help

gitpeterwind: There was a bug in the master code. A PR has been sent

src/qmoperators/one_electron/HessianOperator.h
moritzgubler: @ilfreddy I added some documentation

ilfreddy:
This looks good to me now. Just a couple of minor questions.
@moritzgubler 
1. could you update the description at the top of the PR to reflect the actual state of the PR (it still says GGA is not included). 
2. a question: how about other functionals (like meta-GGA), Would they require additional implementation
3. Is HF exchange working now?

@stigrj this branch includes a merge a few commits ago. Would you let it pass or should it be changed to a rebase?

moritzgubler:
@ilfreddy 
1. lda, gga, closed and open shell are working, the description is updated.
2. the meta-GGA contribution to the stress would have to be implemented. This should be quite easy. I thought only l and gga functionals are implemented in mrchem.
3. No. This would require the exchange stress density. This paper might help: https://pubs.acs.org/doi/10.1021/acs.jctc.0c01194?ref=pdf

ilfreddy:
> @ilfreddy
> 
> 1. lda, gga, closed and open shell are working, the description is updated.
> 2. the meta-GGA contribution to the stress would have to be implemented. This should be quite easy. I thought only l and gga functionals are implemented in mrchem.
> 3. No. This would require the exchange stress density. This paper might help: https://pubs.acs.org/doi/10.1021/acs.jctc.0c01194?ref=pdf

Thanks for the reference! Indeed we only have GGA at present, but we have some plans to make a libXC interface and thus include also metaGGA functionals. But I guess at this point it might be more relevant to focus on HF exchange.

Issue #487, opened 2024-05-30
Failing response SCF test


stigrj:
Random segfault in `Projecting occupied space` in response SCF test

stigrj:
@gitpeterwind I see you have written this comment in `MRCPP/src/utils/ComplexFunction.cpp` :
```
//For some unknown reason the h2_mag_lda test sometimes fails when schedule(dynamic) is chosen
```

I have traced a failing test `li_pol_lda` to the same OMP loop, even with `schedule(static)`, so there seems to be some issue here.

stigrj:
The issue appears when compiling with pure OMP, no MPI, and is randomly triggered when running the test case as is, but seems to always trigger if I change `world_prec` from 1e-3 to 1e-2.

gitpeterwind:
There seems to be an obvious bug: the line 1804:
        ` Sreal(orbVecBra[i], orbVecKet[j]) += S_temp(i, j);`
should be replaced in the same way as is done for the real case (20 lines higher):
```
                        // must ensure that threads are not competing
                        double &Srealij = Sreal(orbVecBra[i], orbVecKet[j]);
                        double &Stempij = S_temp(i, j);
#pragma omp atomic
                        Srealij += Stempij;
```
If you have a setup where you can test this easily, that would be nice (even push a fix?)

stigrj:
Yes, I will test it later 🙂

stigrj:
Hmm, but I'm now running in the `if (serial)` branch where the `omp atomic` is already added :thinking: But for some reason I'm not able to trigger the error anymore, even without changing the code... Need to look closer tomorrow.

I remember having some issues with the `+=` operator in Eigen before, though.

gitpeterwind:
Sorry, I read wrongly the if-else block. So my previous comment was meaningless.
It is difficult with these non-reproducible bugs. But I cannot see another unsafe part than that +=. We can put that in a safer way, but I cannot do that now.

Issue #474, opened 2023-11-17
Numerical problems with relativistic zora calculations


moritzgubler:
Hi

I recently noticed, that I am not able to converge zora calculations with tight precisions. The problem is the following: consider the input file
```
world_prec = 1.0e-6
world_unit = angstrom
WaveFunction {
  method = lda
  relativity = zora
}
Molecule {
$coords
H 0.0 0.0 0.0
H 0.9 0.0 0.0
$end
}
```
This does not converge. It is close though, it almost reaches the threshold after 10 SCF iterations. After that the MO residual stays slightly above the convergence threshold for the next 20 SCF iteration. Then the simulation goes really wrong and the residual starts to diverge. Also, the memory required by mrchem continuously grows and after 47 iterations i stopped the program because it used up the entire memory of my workstation (32 GB). 

I used the newest version of mrchem and tested the functionals lda, pbe and pbe0 where the problem is the same. It is also present in all the other molecules I tested: H_2O, NH_4, it is even more severe in the NH_4 molecule.  When the world_prec is increased to 1e-4 or the zora method is not used, the calculation converges. Below you find the output of the SCF cycle from the input file from above.

I tried to play around with some SCF keywords to get these calculations to converge, but was not able to do so. Do you have an idea what could go wrong here? Thanks a lot for looking into this.

```
===========================================================================
 Iter           MO residual             Total energy                Update
---------------------------------------------------------------------------
    0          1.000000e+00          -1.126549655620         -1.126550e+00
    1          5.588320e-02          -1.137476166839         -1.092651e-02
    2          5.683084e-03          -1.137719316527         -2.431497e-04
    3          1.426500e-03          -1.137732864726         -1.354820e-05
    4          8.197909e-04          -1.137724341320          8.523405e-06
    5          7.077225e-04          -1.137733969255         -9.627935e-06
    6          5.908577e-05          -1.137733876118          9.313768e-08
    7          6.918668e-05          -1.137733974038         -9.792065e-08
    8          3.802536e-05          -1.137733939618          3.442023e-08
    9          4.411585e-05          -1.137733973516         -3.389800e-08
   10          2.451983e-05          -1.137733968648          4.867959e-09
   11          4.849671e-05          -1.137733930095          3.855352e-08
   12          5.043205e-05          -1.137733966139         -3.604488e-08
   13          2.160990e-05          -1.137733974271         -8.131049e-09
   14          5.883780e-05          -1.137733878856          9.541477e-08
[...]

PR #491, opened 2024-08-15
Kappa minus one

ilfreddy:
The current ZORA implementation has a major flaw. The $\kappa$ function is projected "as is", which means that its precision is significantly reduced, because the function is essentially one in the whole volume and has cusps in the neighbourhood of each nucleus. As such the function is strictly speaking not an L2 function. We can still represent it because the simulation box is finite. However its norm gets basically as large as the volume, affecting the actual precision of the representation of the cusps. This will in turn affect also the representation of derivatives (zero everywhere except at the nuclei)

The function $\kappa - 1$ is therefore a much better choice.

This PR implements the projection of $\chi = \kappa - 1$. Whenever $\kappa$ is required, we use $\chi + 1$ instead.

Notice also that $\nabla \kappa = \nabla \chi$. Derivatives are therefore unaffected and (hopefully) computed at a better precision.


moritzgubler:
The changes we made would solve both the examples (NH3 and H2 with tight precisions) that did not converge in issue #474. I brought the branch up to date with master, added of documentation and cleaned up the code. @ilfreddy I think you can remove the draft status, the PR is ready for review now in my opinion.

stigrj (approved):
The failing test seems to be the same issue as #487, so I will just bypass the test requirement and merge this

PR #496, opened 2024-10-17
Atomic Zora Relativity

moritzgubler:
This branch implements the atomic ZORA relativistic approximation where the Zora potential is precomputed for each element and a superposition of atomic potentials is the zora potential.


ilfreddy (commented):
Thanks for implementing AZORA! Could you also include tests for it? It does not seem to me there are any at the moment.

moritzgubler:
@ilfreddy I added testing

ilfreddy (approved):

PR #500, opened 2025-01-28
Remove macos-12 test (temporarily)

Gabrielgerez:
Same as in MRCPP


ilfreddy (approved):
In order to get going with the PRs which are piling up we suspend testing on MAC, but we should resume it eventually

PR #508, opened 2025-09-01
Update Functional.cpp

ilfreddy:
Bugfix for GGA precision issue. Found by @gitpeterwind !

### DIFF
--- a/src/mrdft/Functional.cpp
+++ b/src/mrdft/Functional.cpp
@@ -307,6 +307,7 @@ void Functional::makepot(mrcpp::FunctionTreeVector<3> &inp, std::vector<mrcpp::F
                 node.attachCoefs(Ctrout.col(xc_outsize + 3*(i-1) + d).data());
                 node.cvTransform(mrcpp::Backward);
                 node.mwTransform(mrcpp::Compression);
+                node.calcNorms();
                 mrcpp::DerivativeCalculator<3> derivcalc(d,*this->derivOp, *rho0);//TODO: define outside loops
                 mrcpp::MWNode<3> noded(rho0->getNode(nodeIdx),true,false);
                 derivcalc.calcNode(node, noded);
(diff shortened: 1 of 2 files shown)


stigrj:
Looks like something bad is still happening in the "Li_polarizability_LDA" test. I get some random hangs/crashes when I run it with many OMP threads

gitpeterwind:
> Looks like something bad is still happening in the "Li_polarizability_LDA" test. I get some random hangs/crashes when I run it with many OMP threads

I could not reproduce (tried with up to 64 threads on my 8 cores PC). It might be system dependent.

ilfreddy:
But is that the reason why the tests are failing here? I only see a timeout issue...

stigrj:
Hehe, I actually remember that Ne discrepancy by heart :sweat_smile: It's from the functional parameters in xcfun vs libxc. Try to uncomment line 39 in `xcfun/src/config.hpp: #define XCFUN_REF_PBEX_MU` and recompile, I'll bet the agreement gets better :slightly_smiling_face:

ilfreddy:
@stigrj, could it also be that the input needs some modifications? A standard input will give you a relative precision of 1e-6, but the calculations from 2017 were using somehow stricter threshold to stay on the safe side (correct me if I'm wrong).

You can find the full description here:
[computational_details.pdf](https://github.com/user-attachments/files/22095232/computational_details.pdf)

In summary:
1. the threshold is in absolute value (Hartree)
2. the relative energy threshold is adapted accordingly (eps_rel=eps_abs/E_estimated)
3. the SCF convergence on the energy is reached when DeltaE < eps_abs/10 ( believe there is a mistake in the formula here, and it should be eps_abs, otherwise it does not make sense)
4. the orbital error norm threshold is sqrt(eps_rel/10)
5. the polynomial order is k = - 1.5log(eps_rel)

For Ne (E_est about 100) you should then use eps_rel=1e-8 and k=12 (and the XCFUN modifications from Stig).

PR #515, opened 2025-10-24
Copy OrbitalVector into complex type

gitpeterwind:
Must be used in conjunction with corresponding mrcpp PR (so the tests will fail to compile until then!)


ilfreddy (commented):
It would be good to add at least one unit test of the new functionality. It shoudl have ben aded to mrcpp as well though...

PR #521, opened 2025-11-05
Adding Libxc support

Tirtho13:
# Libxc interface in mrchem
This is a detailed list of things to do in the integration of Libxc into MRChem. The checklist should have all the necessary changes to complete the integration process. 

This is the checklist of all the necessary changes.
NOTE: Please refrain from making commits to the checklists before the changes are made public for review. For owning a particular job, note your name at the end of the specified job. If you want to make a note, add it to the dev notes.

### CMAKE files
- [x] Fix autocmake for libxc, currently hard coded paths in Factory.h and Functional.h. (Tirthonkor)
-  Works, but very slow --> needs optimizing (check compile flags?)
- Update: currently not working anymore
- Udate: working fine, but still have to make final edits

### Runtime backend
- [x] Simultaneous support for XCFun and Libxc backends: the choice should be made at runtime based on user input, which backend to use, xcfun or libxc. (Ylva)
-  Can be decided with the keyword DFT { xc_library = libxc }, default is xcfun.

### Libxc input
- [x]  Human readable Libxc input: pass in keyword strings like GGA_X_PBE and parse them with xc_functional_get_number() to get the numerical ID (you can still trivially also support integer IDs) (Ylva)
-  I think this is done? Code needs expanded list of functionals, but as it is now, the functional name is eg. "slaterx" or "beckex" from mrchem/xcfun which gets translated (manually) to corresponding libxc functionals, "XC_LDA_X" etc.
- Susi implemented this

### Density threshold
- [x] Printing out the functional's density threshold since that will be really important for high precision calculations (Ylva)

## To do during/after restructuring (to avoid doing it twice)

Making functionals class  abstract and create new classes for libxc/xcfun inheriting from functionals.
Ylva will make a rough separation of classes and then we can update a to do list of which parts that should be done (a lot of the functional class functions are using xcfun and should be made general, but that can be done function by function).

### Libxc and functional citation
- [x] Proper printout of the Libxc citation information; both for Libxc itself (xc_version(), xc_reference(), xc_reference_doi()) and the used functionals (see Libxc manual for examples) (Ylva) (Tirthonkor)
-  Prints citation for libxc, not yet for the used functional
- Should use libxc function to automatically get correct citation
- Citations for individual functionals added

### Get hyb parameters from libxc
- [x] Checking that hybrid parameters are parsed from Libxc, not hard-coded in MRChem 
-  Done

### Overwrite Libxc functional settings

- [ ] Support for overwriting Libxc functional settings; this requires defining parameter arrays for each functional used in the calculation and calling xc_func_set_ext_params for each of them before the functional is used (the parameters can change e.g. the fraction of exact exchange) (Tirthonkor)
- This can be done in a later pull; there remains too much to be done

### Testing
- [x] Make current xcfun-based unit tests work
- [x] Make current xcfun-based regression tests work
- [x] Make new libxc-based unit tests
- [x] Make new libxc-based regression tests

### Documentation
- [ ] User manual with some input examples
- [ ] Programmer's manual outlining how the interfacing of the two libraries has been designed and implemented

In the interest of time we postpone this to the next PR which will include other missing points from this PR.

### DIFF
--- a/src/mrdft/Factory.cpp
+++ b/src/mrdft/Factory.cpp
@@ -80,6 +80,25 @@ void MapFuncName(std::string name, std::vector<int> &ids, std::vector<double> &c
         ids = {XC_GGA_X_B88, XC_GGA_C_PW91};
         coefs = {1.0, 1.0};
         return;
+    } else if (name == "B3LYP") {
+        // Keep as b3lyp5 for now to be consistent with xcfun
+        // TODO: change the definition of b3lyp in mrchem to not be b3lyp5
+        ids = {XC_HYB_GGA_XC_B3LYP5};
+        // ids = {XC_HYB_GGA_XC_B3LYP};
+        coefs = {1.0};
+        return;
+    } else if (name == "B3LYP5") {
+        ids = {XC_HYB_GGA_XC_B3LYP5};
+        coefs = {1.0};
+        return;
+    } else {
+        // Check if Libxc has this functional
+        int number = xc_functional_get_number(name.c_str());
+        if (number == -1) { MSG_ABORT(name + " is not a known shorthand in MRChem nor a functional in Libxc!\n"); }
+
+        ids = {number};
+        coefs = {1.0};
+        return;
     }
 }
 


ylvao:
We still have some numerical differences with GGA.
Example input file is attached as a txt file (github does not allow .inp files, apparently...)
[h2_libxc.txt](https://github.com/user-attachments/files/23822766/h2_libxc.txt)

### For slaterx / LDA_X - H2
world_prec = 1.0e-2 
Total energy difference between xcfun and libxc: -2.2000e-11 hartree

This is okay

### For beckex / GGA_X_B88 - H2
world_prec = 1.0e-2
Total energy difference between xcfun and libxc: **-0.00663** hartree
Output files:  
[h2_libxc_out.txt](https://github.com/user-attachments/files/23822896/h2_libxc_out.txt) 
[h2_xcfun_out.txt](https://github.com/user-attachments/files/23822897/h2_xcfun_out.txt)


world_prec = 1.0e-5
Total energy difference between xcfun and libxc: **-0.00532** hartree

This is not okay

### What I have checked:

- [x] Functionals do (as expected) give exact same  output for a single "arbitrary" point --> problem is not functionals
- [x] Density cutoff is the same when running mrchem. Either -1 or 0, but always the same (i think this depends on different implementations around the program)
- [ ]  Look at specific grid points in mrchem and see if there are some trends (I have tried this before without anything notable, but I will try again as we have done some bug fixes since then)

If you have any other ideas what to check/look for, I appreciate all ideas :)

src/mrenv.cpp
Tirtho13: Does this not print the citation for both the libraries twice?

src/mrdft/Functional.cpp
```diff
@@ -74,19 +128,130 @@ Eigen::MatrixXd Functional::evaluate_transposed(Eigen::MatrixXd &inp) const {
...
+                        for (size_t k = 0; k < nPts; k++) {
+                            rho_spin(k * 2, 0) = inp(k, 0);
+                            rho_spin(k * 2 + 1, 0) = inp(k, 1);
+                        }
+                        for (size_t j = 0; j < nPts; j++) {
+                            // Susi: check that the code is correct
+                            // Libxc expects reduced gradients: up-up, up-down, down-down
+                            sigma(3 * j, 0) = inp(j, 1) * inp(j, 1) + inp(j, 2) * inp(j, 2) + inp(j, 3) * inp(j, 3);
+                            sigma(3 * j + 1, 0) = inp(j, 1) * inp(j, 4) + inp(j, 2) * inp(j, 5) + inp(j, 3) * inp(j, 5);
+                            sigma(3 * j + 2, 0) = inp(j, 4) * inp(j, 4) + inp(j, 5) * inp(j, 5) + inp(j, 6) * inp(j, 6);
```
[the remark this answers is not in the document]
susilehtola: this is now correct

ylvao:
> > So, many things work now; the hardest part remains: isolating the XCFun and Libxc implementations.
> > The code looks quite like spaghetti to me. For some reason, there is both Factory and Functional, both of which contain functional stuff. Moreover, Functional already has inheritance; it is instantiated into LDA, SpinLDA, GGA, or SpinGGA, which each have some side effects. Some code is also in driver.cpp.
> > The backend should be hidden behind an abstract interface that only provides the key functionality needed for MRChem. However, I still have not figured out a way to do this after spending most of the day to untangle the code.
> 
> Yes I agree this is going to be challenging. But at the same time are you sure you want to do this now? In my opinion at this point the goal is a code which works fine (gives the right answer) and can choose either libxc or xcfun at runtime. I believe this is now there. We can then tag the version and use it for production. Thereafter we can work on disentangling the mess.

I have an idea on the restructuring that shouldn’t take that much time and effort. I’ll set up a PR on this once a draft is done, however, I do agree with Luca that it might not be the most urgent part. 

My priority is getting some test working, but after that I don’t know what we should prioritize? As for the tests Im thinking of duplicating some xcfun test to check lda/gga both with/without spin (four tests). Anything else?

(also make the current tests pass)

src/mrdft/Functional.cpp
```diff
@@ -171,54 +171,57 @@ void Functional::evaluate_data(const Eigen::MatrixXd &inp, Eigen::MatrixXd &out)
...
-                        vxc = Eigen::MatrixXd::Zero(2, nPts);
-                        for (size_t j = 0; j < nPts; j++) {
+                        vxc = Eigen::MatrixXd::Zero(1, 2 * nPts);
+                        for (size_t k = 0; k < nPts; k++) {
                             // alpha_1, beta_1, alpha_2, beta_2, ..
-                            rho(0, j) = inp(0, j);
-                            rho(1, j) = inp(1, j);
+                            rho[2 * k]     = inp(k, 0);
+                            rho[2 * k + 1] = inp(k, 1);
                         }
```
susilehtola: I don't like this change

src/mrdft/Functional.cpp
```diff
@@ -171,54 +171,57 @@ void Functional::evaluate_data(const Eigen::MatrixXd &inp, Eigen::MatrixXd &out)
...
+                            //  xcfun computes rho * exc for energy density, so we do the same
+                            //    aka xcfun calculates actual energy density while libxc calculates
+                            //    energy density per electron density
+
+                            // rho = rho_alpha + rho_beta (energy is much closer to xcfun when using both)
                             out(0, j) += exc(0, j) * libxc_coeffs[i] * (inp(0, j) + inp(1, j));
-                            out(1, j) += vxc(0, j) * libxc_coeffs[i];
-                            out(2, j) += vxc(1, j) * libxc_coeffs[i];
+                            out(1, j) += vxc(0, 2 * j + 0) * libxc_coeffs[i];
+                            out(2, j) += vxc(0, 2 * j + 1) * libxc_coeffs[i];
```
susilehtola: same here, one should not have to manually compute the index; this is just a matter of choosing the right layout

src/mrdft/Functional.cpp
```diff
@@ -283,15 +276,6 @@ void Functional::evaluate_data(const Eigen::MatrixXd &inp, Eigen::MatrixXd &out)
...
 }
 
-/** @brief Run a collection of grid points through Libxc or XCFun
- *
- * Each column corresponds to one grid point.
- * From a performance point of view, (in pre and postprocessing) it is much more
- * efficient to have the two consecutive points in two consecutive adresses in memory
- *
- * param[in] inp_data Matrix of input values
- * param[out] out_data Matrix of output values
```
susilehtola: same here

src/mrdft/Factory.cpp
```diff
@@ -77,6 +81,10 @@ void MapFuncName(std::string name, std::vector<int> &ids, std::vector<double> &c
         ids = {XC_GGA_X_B88, XC_GGA_C_PW91};
         coefs = {1.0, 1.0};
         return;
+    } else if (name == "B3LYP") {
+        ids = {XC_HYB_GGA_XC_B3LYP5};
```
susilehtola:
```suggestion
        ids = {XC_HYB_GGA_XC_B3LYP};
```

susilehtola:
> @susilehtola could you be a bit more specific? We have merged the pseudopotentials PR, then Ylva made sure the libxc branch passes the tests (#502 introduced some naming conflicts). I don't think anything else has happened and this should now be merged into master IMO.

new functional aliases were added, of which several were incorrect.

ilfreddy:
> ping @ilfreddy @ylvao

Ylva is travelling and I'm on vacation. But she will address your comments as soon as she is back

PR #525, opened 2025-12-02
Population analysis

msnik1999:
Added the possibility to perform population analysis as a new property.

Also possible for half of the space at a time, by using only half of the root nodes.

Needs this [MRCPP PR](https://github.com/MRChemSoft/mrcpp/pull/271).


src/properties/PopulationAnalysis.h
```diff
@@ -0,0 +1,83 @@
...
+
+#include "mrchem.h"
+#include "utils/print_utils.h"
+
+#include <nlohmann/json.hpp>
+#include <string>
+
+namespace mrchem {
+
+class PopulationAnalysis final {
```
ilfreddy: Please document class and methods :-)

doc/users/user_inp.rst
```diff
@@ -274,6 +274,8 @@ properties are available (all but the dipole moment are ``false`` by default)
       geometric_derivative = false          # Compute geometric derivative
       plot_density = false                  # Plot converged density
       plot_orbitals = []                    # Plot converged orbitals
+      population_analysis = false           # Compute population analysis, also for half of the space
```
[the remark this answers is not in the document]
msnik1999: Yeah, that makes sense, should I just fully remove the pop analysis for the whole space?

src/driver.h
```diff
@@ -23,6 +23,8 @@
  * <https://mrchem.readthedocs.io/>
  */
 
+#pragma once
+
```
msnik1999, line 27: I added this here, as it is the case in the other Header files as well. Was there a specific reason this was missing here? @ilfreddy @robertodr

ilfreddy (approved):

PR #532, opened 2025-12-18
Compute two electron integrals and shared memory bug fix

gitpeterwind:
A function that returns all the two electron integrals given an OrbitalVector. 
Eigen::Tensor<std::complex<double>, 4> calc_2elintegrals(double prec, OrbitalVector &Phi)
The optimized setup for exchange is modified for this. Still requires N^2/2 applications of Poisson operator, and therefore the most time consuming part  of a run if required.
Tested on alkanes with 41 orbitals, about 15 minutes on 512 cpus (Betzy).

Also fixed a bug when shared memory for Nuclear or Coulomb potentials are used. The code would hang in that case.

### DIFF
--- a/tests/qmoperators/two_el_int.cpp
+++ b/tests/qmoperators/two_el_int.cpp
@@ -67,7 +67,6 @@ TEST_CASE("TwoElInt", "[two_el_int]") {
 
     Eigen::Tensor<std::complex<double>, 4>two_el_int(nShells, nShells, nShells, nShells);
 
-    int i = 0;
     double t0000 = 0.6249989698;
     double t1000 = 0.0893550109;
     double t1100 = 0.0219478674;


src/qmoperators/two_electron/two_electron_utils.cpp
```diff
@@ -0,0 +1,307 @@
...
+/** @brief computes Int(phi_i^dag*phi_j/|r-r'|)
+ *
+ *  \param[in] phi_i orbital to be conjugated and multiplied by phi_j
+ *  \param[in] phi_j orbital to be multiplied by phi_i^dag
+ *  \param[in] rho is normally the density. It is used to screen the final output
+ *  \param[out] V_ij is the resulting potential function
+  *
+ * Computes the product of complex conjugate of phi_i and phi_j,
+ * then applies the Poisson operator. The result is given in V_ij.
+ */
```
ilfreddy: Move documentation to header file

src/qmoperators/two_electron/two_electron_utils.h
```diff
@@ -0,0 +1,11 @@
+#pragma once
+
+#include "mrchem.h"
+#include "qmfunctions/Orbital.h"
+#include "qmfunctions/qmfunction_fwd.h"
+#include <unsupported/Eigen/CXX11/Tensor>
+
```
ilfreddy: Add file documentation

gitpeterwind:
The test results have changed, because in the latest mrcpp the default for (CompFunction) crop is to use absolute precision instead for relative precision.

tests/qmoperators/two_el_int.cpp
```diff
@@ -0,0 +1,92 @@
...
+    Phi.distribute();
+
+    for (int i = 0; i < Phi.size(); i++) {
+        HydrogenFunction f(ns[i], ls[i], ms[i]);
+        if (mrcpp::mpi::my_func(Phi[i])) mrcpp::project(Phi[i], f, prec);
+    }
+
+    Eigen::Tensor<std::complex<double>, 4>two_el_int(nShells, nShells, nShells, nShells);
+
+    int i = 0;
```
robertodr: seems unused?
```suggestion
```

tests/qmoperators/two_el_int.cpp
```diff
@@ -0,0 +1,92 @@
...
+    double t0000 = 0.6249989698;
```
robertodr: are the reference values from analytical formulas? Or from a previous run of the code itself?

tests/qmoperators/xc_hessian_pbe.cpp
```diff
@@ -103,10 +103,10 @@ TEST_CASE("XCHessianPBE", "[xc_hessian_pbe]") {
     int i = 0;
     DoubleMatrix E_P = DoubleMatrix::Zero(Phi.size(), Phi.size());
 
-    E_P(0, 0) = -0.0496021801;
-    E_P(0, 1) = -0.0234437207;
-    E_P(1, 0) = -0.0234437207;
-    E_P(1, 1) = 0.0061055193;
+    E_P(0, 0) = -5.044421437856e-02;
```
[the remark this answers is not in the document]
gitpeterwind: I had to do a lot of testing to find out: it is density computation for the XCOperator which uses the crop.
The default crop is now using absolute precision by default in mrcpp (relative precision is IMHO very dangerous, since the total norm of a potential can easily be very large or even infinite).
Note that prec is 1e-3 here and we keep 1e3 absolute precision i the results.

tests/qmoperators/two_el_int.cpp
gitpeterwind: It is from the code itself. It requires much (at least for me) work to find the exact values by other means and it is not so useful. 
My first plan was to include a MP2 setup, which then uses the two electron integrals and can be compared to literature results. 
I managed starting from an "initial guess" , and thereby checked that the results where reasonable. 
Although simple on paper, I did not have time to implement mp2 with converged orbitals. That would be a very nice feature to get in the future, if somebody has time. We could discuss this at some meeting.

tests/qmoperators/xc_hessian_pbe.cpp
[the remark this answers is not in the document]
robertodr: Thanks for the explanation Peter!

tests/qmoperators/two_el_int.cpp
robertodr: MP2 would indeed be nice, but you need the virtual orbitals, no? Or some way of representing the two-body correlation space. Anyway, for H2 I might find a reference with the analytical results for hydrogenic orbitals. I can look it up next week.

PR #535, opened 2025-12-21
Update to SAD routines and data

susilehtola:
Currently, the SAD guesses in MRChem have been generated using single-determinant HF on atoms. This easily leads to symmetry breaking, and therefore unpredictable guess performance. However, PySCF contains an implementation of fractionally-occupied spin-restricted DFT, which can be used to determine good symmetric atomic densities.

I have also refactored the Python routine in tools/sad.py to remove the needless dependency on Mendeleev (similar functionality is already present in the BSE), switched to full decontraction of the basis, and updated the basis set from the 3-21G basis to the modular AHGBS-7 basis set, which is larger and better converged. If you disagree with that change, I can also revert the basis to 3-21G, but still think it can be (i) decontracted and (ii) updated to a symmetric density.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -48,10 +48,16 @@ variables:
       shell: /bin/bash
       command: |
         cd build
-        ctest --output-on-failure --verbose
+        ctest -N \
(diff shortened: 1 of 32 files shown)


robertodr (approved):
Thanks @susilehtola

PR #540, opened 2026-01-20
KAIN for complex orbitals

edinvay:
Changed the complex L2 inner product to the real L2 inner product.
It makes it consistent with optimization problems appearing in quantum chemistry.

We need to check:
H2_magnetic_properties_LDA (Failed)

We should operate with different inner products in energy expressions (`mrcpp::dot`) and in optimization tools (real part of `mrcpp::dot`). They are conceptually different. For example, the gradient of Lagrangian is an optimization tool, and so it uses the latter one (with the real part).


gitpeterwind:
Can you explain why the imaginary part must be truncated in optimization tools?

PR #545, opened 2026-02-10
Replace duplicated file by a symlink

### LINKED ISSUE #544: Duplicated file
`user_ref.rst` is duplicated in two places, and the files only have a single difference
```
$ wc -l users/user_ref.rst ../python/mrchem/input_parser/docs/user_ref.rst 
 12290 users/user_ref.rst
 12290 ../python/mrchem/input_parser/docs/user_ref.rst
 24580 total
$ diff -up users/user_ref.rst ../python/mrchem/input_parser/docs/user_ref.rst 
--- users/user_ref.rst	2026-02-02 19:26:11.349473306 +0200
+++ ../python/mrchem/input_parser/docs/user_ref.rst	2026-02-02 19:26:11.351473279 +0200
@@ -496,7 +496,7 @@ User input reference
 
     **Type** ``float``
 
-    **Default** ``0.5``
+    **Default** ``0.6``
 
  :ExternalFields: Define external electromagnetic fields.
 
@@ -1104,7 +1104,7 @@ User input reference
     **Type** ``str``
   
     **Default** ````
-
+  
  :Constants: Physical and mathematical constants used by MRChem
 
   :red:`Keywords`
```

The difference was introduced in commit b822448d41017ffa17ef59f07dece4823eff32d9

susilehtola:
Closes #544


ilfreddy:
Yes the files ARE autogenerated, but in a way that requires packages that are normally not required. This is therefore the lesser of two evils (burdening MRChem with a dependency that is seldom needed OR keeping two autogenerated files in the repository). The latter is maybe less aesthetically pleasing but it is the one requiring less work from our side.

PR #550, opened 2026-02-24
Fixed some issues with the compile scripts

msnik1999:
- Fixed an issue with the Olivia environment.
- Removed version numbers from build and install directories in all compilation scripts (Saga, Betzy and Olivia).
- Updated module versions in saga.env.

### DIFF
--- a/tools/saga.env
+++ b/tools/saga.env
@@ -1,5 +1,5 @@
 # Module environment for Saga
 module --quiet purge
-module load foss/2021b
-module load CMake/3.21.1-GCCcore-11.2.0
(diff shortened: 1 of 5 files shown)


tools/saga.env
Gabrielgerez: I assume you double checked these work in Saga, right?

Gabrielgerez (commented):
I think these are good changes. Though, completely removing version tags from the install and build dirs feels a bit extreme. Is there any way we can keep some form of "versioning" in the installs of MRChem through the setup script?  
Maybe something like the latest git hash or similar?

Gabrielgerez (approved):
Looks good to me

PR #552, opened 2026-03-09
Add support for f orbitals (and contracted d orbitals) in guess type gto

MarcusTL12:
New features:
- Support d and f orbitals in gto initial guess
- call build_grid on screened MO if projected orbital is far from normalized
- add utility python script in tools that can convert orca json files into mrchem.bas and mrchem.mop/moa/mob files


src/initial_guess/gto.cpp
```diff
@@ -148,14 +148,26 @@ void initial_guess::gto::project_mo(OrbitalVector &Phi, double prec, const std::
     Timer t3;
     for (int i = 0; i < Phi.size(); i++) {
         Timer t_i;
+        bool built_grid = false;
```
msnik1999: do you mean `build_grid`, or that a grid has been built? Then I think `grid_built` sounds a bit better.

MarcusTL12: It is tracking wether the build_grid was called, hence the tense. It is only used for adding the * to the print of the norm of the orbital, which could be removed, but it has been useful when debugging this code.

msnik1999 (approved):

ilfreddy (approved):

PR #554, opened 2026-03-17
Bugfix for MPI on solvent

Gabrielgerez:
Rank was being set multiple times every iteration.


Gabrielgerez:
Maybe have an if statement to only run this on rank 0?

msnik1999 (approved):

PR #555, opened 2026-03-18
Fetch libxc release tarball instead of full git repository in autocmake

MarcusTL12:
This reduces the needed download size considaribly, about halfing the time to run the setup script on my machine on work wifi.


ilfreddy:
I believe it was like that originally, but we had some CMake issues and I switched to the git tag instead. If it works, and it is also faster, it's just win-win.
@susilehtola what do you think?
UPDATE: OK we are not using the git tag any longer any way. Well then I guess this is just fine

ilfreddy (approved):

PR #557, opened 2026-04-21
Set cmake setting for xcfun to give same pbe numbers as libxc

MarcusTL12:
This uses the un-rounded version of the pbex mu parameter in xcfun, which makes it agree with libxc, and with the old elephant in the room numbers.


msnik1999 (approved):

ilfreddy (approved):

msnik1999:
The test passed now with the loosened threshold, there are, however, still debug print statements.

MarcusTL12:
> The test passed now with the loosened threshold, there are, however, still debug print statements.

The test is temporarily loosened anyways until we fix the compilation of xcfun in the pipeline, so we might as well leave the debug prints there until we fix the test.

PR #558, opened 2026-04-21
Updating tests

ylvao:
Updating the tests in mrchem, this includes:

- [x] Increase tolerance of tests
- [x] Reconstruct into three sets of tests: 1) run every time something is committed, 2) run once for a pull request etc, 3) Extensive set that can be ran sometimes (one folder for each)
- [x] Add xc library name in label + input file
- [x] Increase tolerance and update reference for libxc tests
- [x] Update documentation
- [x] Update the `config.yml` for CircleCI to allow for staging of pipelines and the test splitting
- [x] Reformatting input files that use json format
- [x] All tests should have one of the three labels: short (default), medium (can be ran locally), long (should not be in ctest))
- [x] Folder should have the same name as the test in lower case
- Labels should at least include: mrchem, method, molecule, short/medium/long, tests should not have its name as a label
- [x] imom tests does not have to run every time
- [x] dft tests using xcfun should be similar to libxc (restricted/unrestricted, hyb-gga/lda)

Feel free to add more/edit


msnik1999 (dismissed):

ylvao:
Tests that fail with increased tolerance (1e-10):

- [x] h2_scf_hf
- [x] h2o_hirshfeld_charges_lda
- [x] h2_static_polarizability_lda
- [x] h2_polarizability_solvent_effect
- [x] li_static_polarizability_lda

ylvao:
@MarcusTL12 and @msnik1999, I have made a draft for the documentation, please look over it. Especially the "Division of Tests" is a section that I am not 100 % sure of. Also maybe ad some text between the two first sections?

msnik1999:
The long tests are not directly part of our testing infrastructure, so I'm not sure if we should mention them here?

tests/li_scf_libxc_lda/reference/li.out
```diff
@@ -253,33 +259,45 @@ using xcfun
...
+ Kinetic energy          :            (au)                  7.219654939792
+ E-N energy              :            (au)                -16.902251628177
+ Coulomb energy          :            (au)                  4.002230311753
+ Exchange energy         :            (au)                  0.000000000000
+ X-C energy              :            (au)                 -1.662617392431
  N-N energy              :            (au)                  0.000000000000
+ Non-local pp energy     :            (au)                  0.000000000000
 ---------------------------------------------------------------------------
- Electronic energy       :            (au)                 -7.481513824146
+ Electronic energy       :            (au)                 -7.342983769063
```
ilfreddy, line 270: This is changing quite a lot.... Is there a reason?

tests/li_scf_xcfun_lda/reference/li.out
ilfreddy, line 1: Some small changes in the result here as well, although this is not PBE...

tests/li_scf_libxc_lda/reference/li.out
msnik1999: I think the reason for that is a git-diff thing. If I see it correctly, it is comparing the new LibXC LDA file to the old LibXC B3LYP file, due to all the renaming and moving of files.

tests/li_scf_xcfun_lda/reference/li.out
msnik1999: @ylvao might that difference be due to the slightly different cutoff (`1e-11` to `0.0`)?

ilfreddy (approved):

tests/li_scf_xcfun_lda/reference/li.out
ylvao: yes. The difference is due to the cutoff, I just double checked to be sure :)

PR #560, opened 2026-04-23
Modernize MRChem

### LINKED ISSUE #391: Static code analysis (MRChem)
Copy of https://github.com/MRChemSoft/mrcpp/issues/191 for MRChem.

[cppcheck.log](https://github.com/MRChemSoft/mrchem/files/7791269/cppcheck.log)

### LINKED ISSUE #551: Need to Modernize MRChem
Mrchem does not compile with CMake version > 4 and other external packages have very non-strict versioning. This is not documented ReadTheDocs says:
- Python-3.9 (or later)
- CMake-3.14 (or later)
- GNU-5.4 or Intel-17 (or later) compilers (C++14 standard)

These statements are no longer true.
We can do one of two things:
- properly set requirements and versioning with upper limits (can lead to short lifespan of MRChem due to lack of support on older dependencies)
- Make MRChem compatible with the latest version of all dependencies

The first option needs to be done. The second option is essential for long term lifetime of MRChem

msnik1999:
Modernizing MRChem

- [x] clean-up cmake
  - [x] more user-accessible cmake options
  - [x] "unify" cmake files (make different cmake files more similar)
  - [x] remove warnings with cmake 4.x.x
  - [x] more cmake output (progress / status) -> probably not important
- [x] libxc folders (should be fixable in fetch_libxc.cmake)
- [x] locking current upper versioning in libraries 
  - [x] XCFun
  - [x] LibXC
  - [x] Python
  - [x] cmake
  - [x] nLohmann JSON
  - [x] compilers (Intel a bit unclear, as "classic" ICC compilers are no longer updated --> removed support for old intel compilers)
  - [x] Eigen3
  - [x] Catch
- [x] Ubuntu image for testing contains XCFun from conda, should be built in the pipeline instead
- [x] Moved from "Updating tests" PR: Adding more tests (hf, lda, pbe/b3lyp, another standard xc functional) at high precision up to medium size molecules) -> will be done later
- [x] Clean up dft functional type classes (LDA/GGA). Some functions are never used (eg. preprocess, postprocess) -> will be done in #573 
- [x] Maybe review install documentation
- [x] Maybe add more pipelines for different compilers (clang) / configurations (macOS) here? -> later
- [x] Fix Singularity containers, or remove them? -> Discussion -> later
- [x] Switch to CPM


msnik1999:
Ready for review, but https://github.com/MRChemSoft/mrcpp/pull/281 has to be merged first

src/driver.cpp
ylvao:
```suggestion
            success_y = initial_guess::cube::setup(Y, prec, cube_yp, cube_ya, cube_yb);
```

ylvao (dismissed):
Should 'success_x' -> 'success_y'?

external/upstream/fetch_eigen3.cmake
```diff
@@ -1,20 +1,25 @@
-find_package(Eigen3 3.4 CONFIG QUIET
-  NO_CMAKE_PATH
-  NO_CMAKE_PACKAGE_REGISTRY
-  )
-
+if(EIGEN3_FIND_BEHAVIOUR STREQUAL "default" OR EIGEN3_FIND_BEHAVIOUR STREQUAL "onlylocal")
```
robertodr: I've recently found out about [CPM](https://github.com/cpm-cmake/CPM.cmake) which basically lets you remove all of the boilerplate in these `fetch_*.cmake` scripts. I'd recommend switching to it.

src/properties/OrbitalEnergies.h
```diff
@@ -88,7 +88,7 @@ class OrbitalEnergies final {
         DoubleVector occ = getOccupation();
         const DoubleVector &eps = getEpsilon();
         std::vector<std::string> spn;
-        for (auto i = 0; i < spinVec.size(); i++) {
+        for (int i = 0; i < spinVec.size(); i++) {
```
robertodr, line 91: why not `auto` here? and is it `spinVec.size()` a `size_t` ?

robertodr (approved):
thanks @msnik1999! My only additional suggested change is to use `CPM` instead of the hand-written `fetch_*.cmake` scripts.

robertodr (approved):

msnik1999:
Wait before merging, we will discuss when, how and in what order to merge

external/upstream/fetch_xcfun.cmake
```diff
@@ -7,8 +7,9 @@ else()
   add_compile_definitions(XCFUN_REF_PBEX_MU)
 endif()
 CPMAddPackage(
-  NAME xcfun
-  VERSION 2.1.1
+  NAME XCFun
+  VERSION 2.1
+  GIT_TAG v2.1.0
```
msnik1999: why 2.1.0 instead of 2.1.1?

ilfreddy (approved):

Issue #516, opened 2025-10-27
Eigen3 v3.3 not compatible with MRChem


msnik1999:
In this [PR](https://github.com/MRChemSoft/mrchem/pull/514), @susilehtola removed the version requirements for Eigen3, which seems to be the only way to get Eigen3 v5.0.0 to work, according to their [GitLab](https://gitlab.com/libeigen/eigen/-/issues/2972). MRChem, however, needs a minimum version of v3.4.0, as I was unable to compile it using v3.3.7.
I don't think there currently is a solution to have a minimum requirement of v3.4.0 and still be able to use v5.0.0, so I don't really know a solution to this dilemma. @ilfreddy

msnik1999:
Has been temporarily solved by not supporting Eigen3 v5.x.x and will be adressed in #551 and #560.

PR #565, opened 2026-05-29
Fix Eigen3 detection for version 5.x and newer

susilehtola:
## Summary
- `find_package(Eigen3 3.4 CONFIG)` is rejected by modern `Eigen3ConfigVersion.cmake`, where a single requested version `3.4` means "3.4.x only" (`3.4 <= v < 3.5`).
- This caused an installed Eigen **5.x** to be treated as incompatible, falling back to fetching/building Eigen 3.4.0.
- Switch to a CMake version range `3.4...6` so any `3.4 <= v < 6.0` (including 5.x) is accepted.

## Test plan
- [x] Verified `find_package(Eigen3 3.4...6 CONFIG REQUIRED ...)` detects system Eigen `5.0.1` (previously fell through to FetchContent).
- [ ] Confirm a clean configure still works on systems with Eigen 3.4.x.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/external/upstream/fetch_eigen3.cmake
+++ b/external/upstream/fetch_eigen3.cmake
@@ -1,4 +1,7 @@
-find_package(Eigen3 3.4 CONFIG QUIET
+# Use a version range so that newer Eigen releases are accepted: with modern
+# Eigen3ConfigVersion.cmake a bare "3.4" means "3.4.x only" (3.4 <= v < 3.5),
+# which rejects e.g. Eigen 5.x. The range matches 3.4.0 up to (but not incl.) 6.0.
+find_package(Eigen3 3.4...6 CONFIG QUIET
   NO_CMAKE_PATH
   NO_CMAKE_PACKAGE_REGISTRY
   )


robertodr (approved):

msnik1999:
I wrote an issue about that problem here: #516 . The problem is that, at least at the time,  there was no way to require a minimum version of 3.4, while also allowing for 5.0, due to some minor incompatibilities on the side of Eigen. We will try to address this in #560 .

msnik1999:
> I wrote an issue about that problem here: #516 . The problem is that, at least at the time, there was no way to require a minimum version of 3.4, while also allowing for 5.0, due to some minor incompatibilities on the side of Eigen. We will try to address this in #560 .

This problem seems to be resolved now, I tested the `3.4...6` version range and it produces the expected behaviour.
As compilation is currently failing when compiling with Eigen3 5.0.x, we decided to strictly require Eigen3 3.4 for now. In the future, we will look into testing Eigen3 5.x.x properly and enabling it once we fix the compilation errors.

Therefore, I will close this PR for now.

PR #569, opened 2026-06-22
Fix MPI update issue for solvent

### LINKED ISSUE #563: mpi issue with solvent kain
I have now tried to debug the accelerator in GPESolver.cpp for around a week debugging time, a month in real time. 
the code freezes when a certain amount of mpi processes and kain history are used for solvent. This was originally a rank_mismatch that I "fixed" in a previous pull request.

steps to replicate:
mpi tasks: 3 or more
solvent kain: 2 or more

Code will freeze when calling `allreduce_vector()` later in the calculation when computing the fock matrix. in this [line](https://github.com/MRChemSoft/mrchem/blob/dd1cbfa780fc089a85c4ab6fb2bb6b9e779066ed/src/qmoperators/two_electron/FockBuilder.cpp#L252)

 This makes it really hard to debug, as what triggers the error is nowhere close to this line of code.

I suspect the bug is in this [function](https://github.com/MRChemSoft/mrchem/blob/dd1cbfa780fc089a85c4ab6fb2bb6b9e779066ed/src/environment/GPESolver.cpp#L142-L159) in GPESolver.cpp
```
void GPESolver::accelerateConvergence(mrcpp::CompFunction<3> &dfunc, mrcpp::CompFunction<3> &func, KAIN &kain) {
    OrbitalVector phi_n(1);
    OrbitalVector dPhi_n(1);

    phi_n[0] = func;
    dPhi_n[0] = dfunc;

    phi_n[0].setRank(0);
    dPhi_n[0].setRank(0);

    kain.accelerate(this->apply_prec, phi_n, dPhi_n);

    func = phi_n[0];
    dfunc = dPhi_n[0]; // see if you can remove these two lines

    phi_n.clear();
    dPhi_n.clear();
}
```
Here is an input file that replicates the freeze:

```
{
  "world_prec": 1e-05,
  "world_unit": "angstrom",
    "MPI": {
    "omp_threads" : 12
  },
  "Printer":{
    "print_level" : 3
  },
  "Molecule": {
    "charge": 0,
    "multiplicity": 1,
    "coords": "  1    0.8510330000    0.6759020000    0.0000000000\n  1    0.3535560000   -1.0275050000    0.0000000000"
  },
  "WaveFunction": {
    "method": "pbe0",
    "environment": "pcm"
  },
  "PCM": {
    "SCRF": {
      "kain": 2,
      "max_iter": 100,
      "dynamic_thrs": true
    },
    "Solvent": {
      "Permittivity": {
        "epsilon_in": 1.0,
        "epsilon_out": {
          "static": 80.0
        },
        "formulation": "exponential"
      }
    }
  },
  "SCF": {
    "run": true,
    "kain": 6,
    "guess_type": "sad_gto"
  }
}
```
2 MPI tasks do not have this issue, I am guessing since rank 0 is the master rank and only rank 1 is doing any work. Try with MPI tasks = 3 for the freeze to happen

I have tried to use deep_copy to fix this, also chage ranks, set barriers before and after. 

I apologize if this issue seems passive aggressive, this bug is literally stopping me from finishing my phd.

Gabrielgerez:
I had a sessions with Claude to fix this error. 

Code was freezing in a later part of the run.
Solves bug in [#563](https://github.com/MRChemSoft/mrchem/issues/563)

### DIFF
--- a/src/environment/GPESolver.cpp
+++ b/src/environment/GPESolver.cpp
@@ -186,13 +186,15 @@ void GPESolver::runMicroIterations(const mrcpp::CompFunction<3> &V_vac, const De
 
         if (iter > 1 and this->history > 0) {
             accelerateConvergence(dVr_n, Vr_n, kain);
-            // KAIN's expandSolution only runs on the rank that owns the function
-            // (rank 0, set via setRank(0) in accelerateConvergence). Broadcast the
-            // updated step to all worker ranks before using it to build Vr_np1,
-            // otherwise ranks 1+ carry a stale dVr_n and diverge from rank 0.
-            mrcpp::mpi::broadcast_function(dVr_n, mrcpp::mpi::comm_wrk);
+            // After KAIN, dVr_n is valid only on rank 0: expandSolution is guarded by
+            // my_func (rank 0 only), and param_copy leaves Ncomp=1 but CompD[0]=nullptr
+            // on other ranks, making recv_function unable to allocate correctly if we
+            // tried to broadcast dVr_n directly. Compute Vr_np1 on rank 0 only, then
+            // broadcast the result (Vr_np1 is freed first so recv_function sees Ncomp=0
+            // and allocates fresh trees before receiving).
             Vr_np1.free();
-            mrcpp::add(Vr_np1, 1.0, Vr_n, 1.0, dVr_n, -1.0);
+            if (mrcpp::mpi::wrk_rank == 0) mrcpp::add(Vr_np1, 1.0, Vr_n, 1.0, dVr_n, -1.0);
+            mrcpp::mpi::broadcast_function(Vr_np1, mrcpp::mpi::comm_wrk);
         }
 
         // set up for next iteration
--- a/src/environment/GPESolver.cpp
+++ b/src/environment/GPESolver.cpp
@@ -27,6 +27,7 @@
 
 #include <MRCPP/MWFunctions>
 #include <MRCPP/MWOperators>
+#include <MRCPP/Parallel>
 #include <MRCPP/Printer>
 #include <MRCPP/Timer>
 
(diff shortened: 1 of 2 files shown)


src/environment/GPESolver.cpp
```diff
@@ -185,8 +186,15 @@ void GPESolver::runMicroIterations(const mrcpp::CompFunction<3> &V_vac, const De
...
+            // After KAIN, dVr_n is valid only on rank 0: expandSolution is guarded by
+            // my_func (rank 0 only), and param_copy leaves Ncomp=1 but CompD[0]=nullptr
+            // on other ranks, making recv_function unable to allocate correctly if we
+            // tried to broadcast dVr_n directly. Compute Vr_np1 on rank 0 only, then
+            // broadcast the result (Vr_np1 is freed first so recv_function sees Ncomp=0
+            // and allocates fresh trees before receiving).
             Vr_np1.free();
-            mrcpp::add(Vr_np1, 1.0, Vr_n, 1.0, dVr_n, -1.0);
+            if (mrcpp::mpi::wrk_rank == 0) mrcpp::add(Vr_np1, 1.0, Vr_n, 1.0, dVr_n, -1.0);
+            mrcpp::mpi::broadcast_function(Vr_np1, mrcpp::mpi::comm_wrk);
```
msnik1999, line 197: @gitpeterwind I really don't know:
Does this have to be done for every rank or just for the first one? I.e. does this function actively send its content to all ranks or do all ranks have to actively request it by calling this function?

msnik1999 (approved):

src/environment/GPESolver.cpp
gitpeterwind: If dVr_n is only defined on rank 0, then we must broadcast, yes. Otherwise better to let each rank does the job. (I do not remember the context)

msnik1999: I meant the broadcast function. Does it have to be called on each rank?

gitpeterwind: every rank has to call the broadcast function, yes!

msnik1999: good to know, thank you

ilfreddy (approved):

PR #570, opened 2026-06-24
Fix for failing singularity docker master pipeline

msnik1999:
Fixes the failing singularity docker master pipelines.
This is a temporary fix, we probably need a new singularity (or apptainer) image longterm. We will loook at this in #560 .

### DIFF
--- a/.github/workflows/build-containers.yml
+++ b/.github/workflows/build-containers.yml
@@ -3,9 +3,6 @@ on:
   push:
     branches:
       - master
-  pull_request:
-    branches:
-      - master
   release:
     types:
       - published
@@ -13,8 +10,8 @@ on:
 jobs:
   build-test-containers:
     runs-on: ubuntu-latest
-    # env:
-    #   ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
+    env:
+      ACTIONS_ALLOW_USE_UNSECURE_NODE_VERSION: true
     strategy:
       # Keep going on other deployments if anything bloops
       fail-fast: false
@@ -34,7 +31,7 @@ jobs:
     name: Build ${{ matrix.program }}-${{ matrix.variant }}
     steps:
       - name: Check out code for the container builds
-        uses: actions/checkout@v7
+        uses: actions/checkout@v2
 
       - name: Build container
         env:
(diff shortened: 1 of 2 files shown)


.github/workflows/build-containers.yml
robertodr: Niklas, the proper fix is to upgrade the action version here. Now I can't recall if latest is V6 or V7, but the error has nothing to do with singularity 

```suggestion
        uses: actions/checkout@v6
```

msnik1999: I'm trying this fix instead. I enabled this pipeline in this PR again, so this can't be merged yet. This pipeline will fail, because it can't store its results, but we will see if it gets past the previously failing step

msnik1999: And it immediately fails again. 
I asked Claude code and that is its output:

This error is different from a simple action version issue. The problem is the job runs inside the quay.io/singularity/singularity:v3.9.2 container (line 27), which is based on old Alpine Linux. Node 24's binary requires pthread_getname_np which that old Alpine doesn't have.

The root cause: the Singularity container image is too old to support Node 24.



https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/

robertodr: Ok. Thanks for teaching me a new thing! And upgrading the version of the image breaks the build I'm guessing?

msnik1999: I have not tried that, I just wanted a quick fix here. Marcus and I can look at this in #560, we're doing a bunch of stuff like this in that PR anyways.

robertodr: I see. FWIW, I've lately given up on making container images directly in singularity/apptainer format, because the ecosystem for it on GitHub and the GitHub container registry is really hit-or-miss. What I've ended up with is making them as docker images with GitHub actions and storing them on ghcr.

* You can pull them in singularity format anyway
* you skip the headache of figuring out how to produce them correctly as OCI images using the singularity tools
* The ecosystem of actions for docker is excellent

msnik1999: So we just do the temporary fix here and then change the infrastructure in the other branch?

robertodr: Yes. Seems the best solution. Also to consider: is anyone actually using these images? 😅

msnik1999: I have no idea. I have never used them.

robertodr (approved):

PR #572, opened 2026-06-28
Option to remove build/install dirs in sigma2 compile scripts

ylvao:
Added a terminal prompt to delete existing build/install folders before compiling using sigma2 compile scripts. Defaults to aborting if no confirmation is given.


msnik1999 (approved):
I assume you have tested the scripts on all clusters?

ylvao:
> I assume you have tested the scripts on all clusters?

Yessir🫡

msnik1999 (approved):

PR #573, opened 2026-07-02
Restructure MRDFT

ylvao:
Clean up and restructure MRDFT. 

Summary:

All code that previously relied on an if-statement for deciding what xc-library to use or call functions from the libraries has been moved to one of the two subclasses xclib_xcfun and xclib_libxc, which inherit from the abstract class xclib. The xc functional family classes (LDA, GGA, ...) has been clean and merged with their header files due to the small amount of code remaining. The goal is to extract xc library dependent code and isolate it so mrchem can be compile with either/none of the xc libraries.

What has been done (might have missed some details):

- Removed unused functions from functional family classes (LDA, GGA, ...): preprocess(), postprocess(), getCtrInputLength(), setupXCInput(), setupCtrInput(). Function makepot() in Functional.cpp handles these functionalities
- Result of point above: Functional family .cpp files are merged with .h files due to size
- All xc library files collected in xc_libraries subdir inside mrdft
- Created new class XCLib wth child classes Libxc and XCfun, handling all "if (libxc)" instances
- Move spin, exx and functional names to xclib class
- Functional::evaluate() is removed (not used) and Functional::evaluate_transposed() is renamed evaluate()

To dos:

- [x] Document/remove unused functions
- [x] Make functional class abstract and make xcfun and libxc inherit from this
- [x] Remove functional classes or at least functions that are used in makepot in functional.cpp
- [x] Cleanup #include headers
- [x] Documentaion (docfiles and code)
- [ ] Fix derivop variable in functional as done in mgga branch
- [ ] Make sure set derivative is called when constructing factory in driver so this can be set in input file (Derivative {dft})
- [ ] Remove (restructure) contract vs contract_transposed

Notes for reviewers (aka help needed):

- Numerical issues causing the test h2_scf_libxc_b3lyp to fail with tol 1.0e-10 (sum_occ, e_kin and e_ee sat to 1.0e-9 to make tests pass). Could be due to Libxc not accepting cutoff = -1
- Factory.cpp currently #include both xclib_XCFun.h and xclib_Libxc.h to be able to construct the correct xclib object (either xcfun or libxc). Ideally, it would only need xclib.h so we can compile without one of the libraries. However, we still need to initialize which library to use; is there a way to do this (maybe with some cmake flags to only include the files if the lib is compiled)? (@MarcusTL12 Has fixed this)
- The density cutoff is used differently for calculations with libxc and xcfun. For xcfun the cutoff is used in mrchem after xcfun is called to decide if the electron density is too low to be included. In libxc the cutoff is given to the libxc object and the cutoff is then dealt with internally in libxc. This results in the variable being "thrown around" in factory.cpp, functional.cpp and the xclib backends, could this be handled more smoothly? (@MarcusTL12 has fixed this)
- Any other thoughts on what should be changed/done


ylvao:
I have now made a draft ready for revision by some more "experienced" eyes. I know there's a lot of stuff going on, but I would appreciate if any of you, @ilfreddy, @msnik1999 and @MarcusTL12 could have a look and share some of your thoughts :blush: 

(Program compiles and tests pass on my computer at least)

src/mrdft/Functional.cpp
```diff
@@ -306,15 +89,13 @@ Eigen::MatrixXd Functional::contract(Eigen::MatrixXd &xc_data, Eigen::MatrixXd &
     for (int i = 0; i < this->xc_mask.rows(); i++) {
         Eigen::VectorXd cont_i = Eigen::VectorXd::Zero(nPts);
         for (int j = 0; j < this->xc_mask.cols(); j++) {
-            Eigen::VectorXd cont_ij = Eigen::VectorXd::Zero(nPts);
             int xc_idx = this->xc_mask(i, j);
             int d_idx = this->d_mask(j);
             if (d_idx >= 0) {
-                cont_ij = xc_data.row(xc_idx).array() * d_data.row(d_idx).array();
+                cont_i.array() += xc_data.row(xc_idx).array() * d_data.row(d_idx).array();
```
msnik1999: You could make it even shorter, by removing cont_i completely:
```suggestion
                out_data.row(i + 1).array() += xc_data.row(xc_idx).array() * d_data.row(d_idx).array();
```

src/mrdft/Functional.cpp
```diff
@@ -23,278 +23,61 @@
...
-            } else {
-              if (inp(0, i) < cutoff) continue;
-            }
-            xcfun_eval(xcfun.get(), inp.col(i).data(), out.col(i).data());
-        }
-    }
+    xclib->callLibEval(inp, out, nPts, nInp, nOut, cutoff);
 }
 
+// NOT USED: evaluate_transposed is used instead
```
msnik1999: can it be removed then?

src/mrdft/Functional.cpp
```diff
@@ -330,11 +111,10 @@ Eigen::MatrixXd Functional::contract_transposed(Eigen::MatrixXd &xc_data, Eigen:
     for (int i = 0; i < this->xc_mask.rows(); i++) {
         Eigen::VectorXd cont_i = Eigen::VectorXd::Zero(nPts);
```
msnik1999: not used
```suggestion
```

src/mrdft/Functional.cpp
msnik1999:
```suggestion
    node.attachCoefs(coef); // reestablish the original link (for proper destructor behaviour)
```

src/mrdft/xc_libraries/XCLib.cpp
msnik1999, line 1: Should probably begin with a capital letter

src/mrdft/Functional.cpp
ylvao: yes, i have mentioned mentioned it in one of the notes in the top comment

msnik1999:
> Currently the initFunctionalLibrary() (in xclib.h), which is overridden by xclib_libxc and xclib_xcfun, takes in the parameters "order" and "gamma", however they are only needed for xcfun. Is there a way to avoid common class functions having variables only needed by one of the subclasses? (This is also a problem with the density cutoff)

You are calling in general `xclib->initFunctionalLibrary(lda, gga, mgga, order, gamma);` in Factory.cpp, therefore you need to give those variables here, because you don't know which library you are calling. You should add those two lines to xclib_Libxc.cpp: l. 92:
`(void)order;`
`(void)gamma;`
to get rid of potential warnings.

src/mrdft/xc_libraries/XCLib.cpp
ylvao: XCLib, Xclib or XClib? (class name will be updated to match)

msnik1999: XCLib sounds good to me

PR #580, opened 2026-09-08
Update MRCPP repository and tag in CMake file

Valentyn113:
I linked new mrcpp tag


external/upstream/fetch_mrcpp.cmake
msnik1999, line 1: You should not link to your own repository here, it should always be the mrcpp repo. So the mrcpp pr has tonbe merged first, and then it can be updated here

msnik1999 (approved):

